PR #1, opened 2026-03-15
fix: remove broken import for non-existent github adapter

sacloudy:
## Summary
- 修复 index.ts 中引用不存在的 github 适配器目录的问题
- 移除了对 `./github/search.js` 的导入，因为对应的 github 目录不存在

## Test plan
- [x] `npx tsx src/main.ts list` 运行成功

### DIFF
--- a/src/clis/index.ts
+++ b/src/clis/index.ts
@@ -12,9 +12,6 @@ import './bilibili/history.js';
 import './bilibili/feed.js';
 import './bilibili/user-videos.js';
 
-// github
-import './github/search.js';
-
 // zhihu
 import './zhihu/question.js';
 


jiangzhx:
i got same error

erixyuan:
Confirmed on current `main` (`92c23e8`). I reproduced this locally because `src/clis/index.ts` still imports `./github/search.js`, but there is no corresponding `src/clis/github/` adapter directory in the repo.

Repro:
```bash
npm install
npx tsx src/main.ts list
```

I also hit the same failure path when building first and then running the compiled entrypoint, so this is not limited to the TS dev path:
```bash
npm run build
node dist/main.js list
```

Removing that import fixes the startup issue on my side.

jackwener:
感谢你的细心发现和贡献！关于模块未找到的问题已经在 `v0.2.0` 版本通过引入 Dynamic Loader 动态加载机制（彻底移除了硬编码的 index.js）从根本上解决了。为了感谢你的敏锐发现和修复尝试，我已经将你设置为了重构 commit `71f2f3d` 的 `Co-authored-by`。🎉

PR #4, opened 2026-03-15
fix: preserve zhihu hot question ids beyond safe integer

### LINKED ISSUE #3: zhihu hot generates wrong question URLs when question IDs exceed JavaScript safe integer range
### Summary

`opencli zhihu hot` can generate incorrect Zhihu question URLs for some hot topics.

The title and heat are correct, but the `url` field may point to a different question because the Zhihu question ID exceeds JavaScript's safe integer range and loses precision during JSON parsing.

### Reproduction

Run:

```bash
opencli zhihu hot --limit 10 -f json
```

Inspect items near the end of the returned list (for example the 10th item when the API returns a large question ID).

### Actual behavior

The adapter currently does:

```js
const d = await res.json();
url: 'https://www.zhihu.com/question/' + t.id
```

If Zhihu returns a question ID like:

```text
2016164966399435420
```

it may be parsed as:

```text
2016164966399435500
```

So the generated URL becomes incorrect.

### Expected behavior

The original question ID should be preserved exactly, and the generated URL should match the real Zhihu question URL.

### Root cause

Zhihu question IDs can exceed JavaScript's safe integer range (`Number.MAX_SAFE_INTEGER`).

Using `res.json()` / `JSON.parse()` directly causes large numeric IDs to be converted to `Number`, which loses precision before the URL is constructed.

### Suggested fix

Read the response as text first, preserve long numeric `id` fields as strings, then parse:

```js
const text = await res.text();
const d = JSON.parse(
  text.replace(/("id"\s*:\s*)(\d{16,})/g, '$1"$2"')
);
const questionId = t.id == null ? '' : String(t.id);
url: 'https://www.zhihu.com/question/' + questionId
```

### Affected file

`src/clis/zhihu/hot.yaml`

### Why this matters

This makes `zhihu hot` output look correct at a glance, but some links silently point to the wrong question, which is a user-facing correctness bug.

jiangzhx:
## Summary

Fix `zhihu hot` generating incorrect question URLs when Zhihu returns question IDs beyond JavaScript's safe integer range.

## Changes

- read the hot list response as raw text instead of calling `res.json()` directly
- preserve long numeric `id` fields as strings before parsing
- build question URLs from the stringified question id

## Verification

- `npm run lint`
- `npm test`

Closes #3.

### DIFF
--- a/src/clis/zhihu/hot.yaml
+++ b/src/clis/zhihu/hot.yaml
@@ -17,12 +17,16 @@ pipeline:
         const res = await fetch('https://www.zhihu.com/api/v3/feed/topstory/hot-lists/total?limit=50', {
           credentials: 'include'
         });
-        const d = await res.json();
+        const text = await res.text();
+        const d = JSON.parse(
+          text.replace(/("id"\s*:\s*)(\d{16,})/g, '$1"$2"')
+        );
         return (d?.data || []).map((item) => {
           const t = item.target || {};
+          const questionId = t.id == null ? '' : String(t.id);
           return {
             title: t.title,
-            url: 'https://www.zhihu.com/question/' + t.id,
+            url: 'https://www.zhihu.com/question/' + questionId,
             answer_count: t.answer_count,
             follower_count: t.follower_count,
             heat: item.detail_text || '',


jackwener:
非常感谢你的贡献！关于超出 JS 安全整数的问题 ID 溢出修复，已经在 `v0.2.0` 的重构中通过 `71f2f3d` 提交被合并了。为了感谢你的付出，我已经将你设置为了此 commit 的 `Co-authored-by`。🎉

PR #6, opened 2026-03-15
fix: escape evaluate args in browser scripts

erixyuan:
## Summary
- add a `json` template filter so browser `evaluate` scripts can inject user input as valid JS literals
- use that filter in `zhihu search` and `reddit hot`, which were previously embedding raw string args into JavaScript source
- add template-engine coverage for quoted and nullish values

## Repro
Before this change, this worked:
```bash
PLAYWRIGHT_MCP_EXTENSION_TOKEN=... node dist/main.js zhihu search --keyword AI --limit 2 -f json
```

But this failed because the single quote broke the generated browser script:
```bash
PLAYWRIGHT_MCP_EXTENSION_TOKEN=... node dist/main.js zhihu search --keyword "O'Reilly" --limit 2 -f json
```

Error:
```text
Error: page._evaluateFunction: Passed function is not well-serializable!
```

## Verification
- `npm test -- --run src/pipeline/template.test.ts`
- `npm run build`
- confirmed `zhihu search --keyword "O'Reilly"` now returns JSON instead of throwing the serialization error

### DIFF
--- a/src/clis/reddit/hot.yaml
+++ b/src/clis/reddit/hot.yaml
@@ -18,9 +18,10 @@ pipeline:
 
   - evaluate: |
       (async () => {
-        const sub = '${{ args.subreddit }}';
+        const sub = ${{ args.subreddit | json }};
         const path = sub ? '/r/' + sub + '/hot.json' : '/hot.json';
-        const res = await fetch(path + '?limit=${{ args.limit }}&raw_json=1', {
+        const limit = ${{ args.limit }};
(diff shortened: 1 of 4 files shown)


jackwener:
感谢你的贡献！这个 PR 的改动已经在 `v0.2.0` 的重构中通过 `71f2f3d` 提交被合并了。为了感谢你的付出，我已经将你设置为了此 commit 的 `Co-authored-by`。🎉

PR #16, opened 2026-03-15
fix: use Playwright MCP executable-path flag

KasumiChen:
## Summary

This fixes the Playwright MCP executable override flag passed by opencli.

opencli currently forwards the browser executable path to `@playwright/mcp` as `--executablePath`, but the Playwright MCP CLI expects `--executable-path`.

As a result, custom browser executable overrides do not work correctly in environments that rely on them, such as WSL launching Windows Chrome.

## What changed

- switch the forwarded CLI flag from `--executablePath` to `--executable-path`
- add regression coverage in `src/browser.test.ts`
- keep the change minimal and focused to the Playwright MCP launcher path

## Validation

- verified against multiple `@playwright/mcp` versions that the CLI flag is `--executable-path`
- focused browser tests pass
- build passes
- real-world validation succeeded in WSL using the Playwright MCP extension path with a Windows Chrome executable override

## Notes

This PR intentionally does not include unrelated experimental browser backend work. It only fixes the Playwright MCP executable-path forwarding bug and adds test coverage.

### DIFF
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -38,6 +38,30 @@ describe('browser helpers', () => {
     expect(__test__.appendLimited('12345', '67890', 8)).toBe('34567890');
   });
 
+  it('builds Playwright MCP args with kebab-case executable path', () => {
+    expect(__test__.buildMcpArgs({
+      mcpPath: '/tmp/cli.js',
+      executablePath: '/mnt/c/Program Files/Google/Chrome/Application/chrome.exe',
+    })).toEqual([
+      '/tmp/cli.js',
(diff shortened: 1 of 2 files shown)


jackwener:
Hi @KasumiChen, thanks for catching this bug and putting together a clean fix with tests! 🙏

I've applied your fix directly to main (commit b0c51dd) with you as a co-author. The core change (--executablePath → --executable-path) and the buildMcpArgs() extraction are incorporated.

Since the fix is now on main, I'll close this PR. Thanks again for the contribution!

PR #29, opened 2026-03-17
fix(opencode): use 'environment' instead of 'env' for MCP config

yuantongkang:
OpenCode config schema uses 'environment' property for MCP server environment variables, not 'env'.

Schema reference: https://opencode.ai/config.json


jackwener:
Thanks for the contribution! 🎉 Good catch on the OpenCode config schema — `environment` is indeed the correct property name per the official schema. Clean and well-scoped fix.

PR #34, opened 2026-03-17
feat(tgb): add tgb site support

hanxuanliang:
Add support for tgb.cn (淘股吧):
- post: export article content with markdown format
- post-comments: export comments with parent-child relationship support
- user-posts: list user posts

All commands use cookie-based authentication strategy.


hanxuanliang:
However, I have a question regarding this PR: 

is there any plan for opencli to support checkpoint-based resumption functionality? Specifically, in pagination scenarios where comments from the same post need to be fetched in multiple batches at uncertain intervals.

PR #41, opened 2026-03-17
feat: support standalone browser mode overrides

0xsline:
## Summary
- add `OPENCLI_BROWSER_MODE=extension|standalone` to override the default browser mode selection
- add `OPENCLI_MCP_NO_SANDBOX=1` support for Linux root / VPS Chromium runs
- document the new VPS / CI flow in the README and testing docs

## Why
Today `buildMcpArgs()` always forces `--extension` outside CI. That works well on a local machine with the Playwright MCP Bridge extension, but it makes VPS and headless Linux setups awkward even when Chromium is already installed.

I hit two concrete issues while testing on a Debian VPS:
- `Extension connection timeout` when no MCP browser extension was present
- Chromium launch failures as root unless `--no-sandbox` was passed

This change keeps the current defaults intact, but adds an explicit escape hatch for standalone environments.

## Testing
- `npm test -- src/browser.test.ts`
- `npm run build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,12 @@ npx vitest run tests/e2e/                    # E2E tests
 - **"Failed to connect to Playwright MCP Bridge"**
   - Ensure the Playwright MCP extension is installed and **enabled** in your running Chrome.
   - Restart the Chrome browser if you just installed the extension.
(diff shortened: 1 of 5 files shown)


jackwener:
It is currently supported, please refer to the latest documentation.

PR #70, opened 2026-03-18
fix(main): navigate to domain before cookie/header strategy commands

backtime1993:
## Summary

- Cookie/header strategy commands (e.g. `bilibili search`, `twitter timeline`, `zhihu hot`) fail with `Failed to fetch` when the active Chrome tab is on an unrelated domain
- Root cause: `fetch()` with `credentials: 'include'` is executed in-page, but cross-origin credentialed requests are blocked by the browser's same-origin policy
- Fix: navigate to `cmd.domain` before executing cookie/header strategy commands, mirroring the pre-navigation already done in the `cascade` command

## Repro

1. Have YouTube (or any non-bilibili page) as the active Chrome tab
2. Run `opencli bilibili search --keyword "test"`
3. → `Error: Failed to fetch`

After fix: opencli automatically navigates to `bilibili.com` first, then runs the API fetch in same-origin context.

## Changes

- `src/main.ts`: Add domain pre-navigation for `Strategy.COOKIE` and `Strategy.HEADER` commands before executing the command function
- Import `Strategy` enum from registry

## Test plan

- [x] `bilibili search` works regardless of active Chrome tab
- [x] `bilibili hot` (public strategy, no cookie) still works
- [x] `youtube search` (cookie strategy) still works
- [x] TypeScript type check passes (`tsc --noEmit`)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -10,7 +10,7 @@ import { fileURLToPath } from 'node:url';
 import { Command } from 'commander';
 import chalk from 'chalk';
 import { discoverClis, executeCommand } from './engine.js';
-import { type CliCommand, fullName, getRegistry, strategyLabel } from './registry.js';
+import { Strategy, type CliCommand, fullName, getRegistry, strategyLabel } from './registry.js';
 import { render as renderOutput } from './output.js';
 import { PlaywrightMCP } from './browser.js';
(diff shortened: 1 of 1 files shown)


backtime1993:
Closing this — after further investigation, this issue only occurs with CDP auto-discovery mode (not yet upstream). The fix belongs in the chrome-auto-discovery feature branch instead. Sorry for the noise!

PR #97, opened 2026-03-19
fix(extension): suppress ERR_CONNECTION_REFUSED and harden WebSocket lifecycle

huangsen365:
## Problem

When the opencli daemon is not running, the Chrome extension's WebSocket reconnect loop produces noisy `ERR_CONNECTION_REFUSED` console errors every few seconds:

```
WebSocket connection to 'ws://localhost:19825/ext' failed: 
Error in connection establishment: net::ERR_CONNECTION_REFUSED
```

These errors:
- Cannot be caught via try/catch (Chrome logs them before JS gets control)
- Clutter the Service Worker console
- May confuse users into thinking something is broken

Additionally, several race conditions existed in the WebSocket lifecycle.

## Solution

**Probe before connect**: Use `fetch(HEAD)` to check if the daemon is reachable before opening a WebSocket. A failed fetch is silent — no console noise. Only proceed with WebSocket when daemon is confirmed up.

**Harden lifecycle**:
- Add `connecting` guard flag to prevent concurrent `connect()` calls from overlapping alarm/event triggers
- Check `WebSocket.CLOSING` state to avoid creating a new socket while the old one is still shutting down
- Capture socket in local variable to prevent stale reference races between `onclose`/`onerror` callbacks
- Wrap `socket.close()` in try/catch in onerror (can throw during SW teardown)
- Only send command responses if the socket is still the active one and OPEN

## Testing

- Daemon stopped → extension quietly retries with exponential backoff, **no console errors**
- Daemon started → extension connects within one backoff cycle
- Daemon restarted → extension reconnects cleanly after onclose fires

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -1,6 +1,7 @@
 const DAEMON_PORT = 19825;
 const DAEMON_HOST = "localhost";
 const DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
+const DAEMON_HTTP_URL = `http://${DAEMON_HOST}:${DAEMON_PORT}`;
 const WS_RECONNECT_BASE_DELAY = 2e3;
 const WS_RECONNECT_MAX_DELAY = 6e4;
 
@@ -94,6 +95,7 @@ function registerListeners() {
 let ws = null;
 let reconnectTimer = null;
 let reconnectAttempts = 0;
+let connecting = false;
 const _origLog = console.log.bind(console);
 const _origWarn = console.warn.bind(console);
 const _origError = console.error.bind(console);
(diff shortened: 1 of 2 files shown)


jackwener:
感谢提交！但是这个修复方案本身存在一个根本性的问题：

**探活 HTTP 请求本身也会产生 `ERR_CONNECTION_REFUSED`**

当 daemon 没有运行时，`fetch('http://localhost:19825', {method: 'HEAD'})` 和直接 `new WebSocket(url)` 一样，都会在 Chrome DevTools Console 里打印 `ERR_CONNECTION_REFUSED`。这意味着这个方案只是把错误的来源从 WebSocket 换成了 HTTP fetch，噪音依然存在，问题并没有真正解决。

此外，直接修改了 `extension/dist/background.js` 构建产物也不符合项目规范，应该只修改 `extension/src/` 下的源文件。

如果要真正解决 console 噪音问题，正确的思路是在已有的 `console.error` hook 里过滤掉 WebSocket 连接失败相关的错误消息，而非加 HTTP 探测。

当前 `onerror → ws.close() → onclose → scheduleReconnect()` 的链路已经正确处理了连接失败的情况，功能上没有 bug。

PR #100, opened 2026-03-19
fix(intercept): IIFE wrapping for installInterceptor/getInterceptedRequests

### LINKED ISSUE #98: [Bug]: INTERCEPT strategy commands never capture data — generated JS not executed as IIFE
## Description

All commands using `Strategy.INTERCEPT` (search, notifications, followers, following) fail to capture API responses. The XHR/fetch interceptor is never actually installed because the generated JavaScript is never executed.

### 1. Arrow function expression not wrapped as IIFE

`installInterceptor()` and `getInterceptedRequests()` in `src/browser/page.ts:261-279` call `sendCommand('exec', ...)` directly, bypassing `wrapForEval()`.

The generated JS from `src/interceptor.ts` is a bare **arrow function expression** (`() => { ... }`), not an IIFE. When CDP `Runtime.evaluate` receives a bare function expression, it evaluates it as a function object but **never executes the function body**. The monkey-patch is never installed.

In contrast, `page.evaluate()` wraps code through `wrapForEval()` (page.ts:290-304) which detects arrow functions and wraps them as IIFEs: `(() => { ... })()`.

The YAML pipeline's `intercept` step (`src/pipeline/steps/intercept.ts:19`) correctly uses `page.evaluate()` and is NOT affected.

### 2. Regression introduced by daemon migration

Git history confirms this is a regression from commit `b2fa7da` ("replace @playwright/mcp with lightweight daemon + Chrome Extension"). Before that commit, `installInterceptor` used `this.evaluate()` (with wrapping). The migration changed it to direct `sendCommand()`, losing the `wrapForEval` wrapping.

| Commit | Method |
|--------|--------|
| `68840fc` / `cfad003` (Playwright era) | `this.evaluate(generateInterceptorJs(...))` — correct |
| `b2fa7da` (daemon migration) and after | `sendCommand('exec', { code: ... })` — broken |

### 3. No wrapping anywhere in the chain

The daemon (`daemon.ts:120`) transparently forwards the command. The extension (`background.ts:176`) passes `cmd.code` directly to `cdp.evaluateAsync()`. CDP (`cdp.ts:41`) passes it directly to `Runtime.evaluate`. Zero wrapping at any layer.

### 4. Additional bug in followers.ts / following.ts

`followers.ts:61` filters intercepted data with `r?.url?.includes('Followers')`, but `getInterceptedRequests()` returns response body JSON (from `response.clone().json()`), NOT request objects with a `url` field. Even after fixing the IIFE issue, this filter would discard all captured data. `following.ts` has the same problem.

## Affected commands

- `twitter search` (#86, #79)
- `twitter notifications` (#87)
- `twitter followers`
- `twitter following`
- Any future TS adapter using `Strategy.INTERCEPT`

## Suggested Fix

**page.ts** — change both methods to use `this.evaluate()`:

```typescript
async installInterceptor(pattern: string): Promise<void> {
    const { generateInterceptorJs } = await import('../interceptor.js');
    await this.evaluate(generateInterceptorJs(JSON.stringify(pattern), {
      arrayName: '__opencli_xhr',
      patchGuard: '__opencli_interceptor_patched',
    }));
}

async getInterceptedRequests(): Promise<any[]> {
    const { generateReadInterceptedJs } = await import('../interceptor.js');
    const result = await this.evaluate(generateReadInterceptedJs('__opencli_xhr'));
    return (result as any[]) || [];
}
```

**followers.ts / following.ts** — remove the incorrect `url.includes()` filter (intercepted data is response JSON, not request objects).

## Relationship to PR #91

PR #91 fixed a real timing issue (interceptor wiped by `goto()` navigation resetting JS context). That fix is necessary but insufficient — even with correct timing, the interceptor code was never being executed due to this IIFE bug. Both fixes are needed together.

Astro-Han:
## Summary

Fixes #98 — INTERCEPT strategy commands fail to capture data.

### 1. Interceptor pipeline fix (`page.ts`)

**Root cause**: `installInterceptor()` and `getInterceptedRequests()` were changed from `this.evaluate()` to direct `sendCommand('exec')` during the daemon migration. This bypassed `wrapForEval()`, so CDP received bare arrow functions that were defined but never invoked.

**Fix**: restore `this.evaluate()` for both methods, so `wrapForEval()` wraps the arrow function as an IIFE `(code)()`.

### 2. Twitter adapter fixes (`followers`, `following`, `search`, `notifications`)

With the interceptor pipeline working, discovered three additional issues preventing data capture:

- **SPA navigation**: `page.goto()` resets JS context, losing the installed interceptor. Changed to install interceptor first, then navigate via SPA (click for followers/following, `pushState+popstate` for search/notifications) to preserve JS context.
- **Data path**: GraphQL response is `{ data: { user: ... } }`, not `{ data: { data: { user: ... } } }`. Removed the extra `.data` level.
- **Author resolution**: Twitter moved `screen_name` from `legacy` to `core`. Updated to try `core.screen_name` first with `legacy` fallback.
- **Selector update**: Twitter changed `/followers` to `/verified_followers`.

### 3. Security hardening

- `JSON.stringify` for all user input interpolated into `evaluate()` JS strings
- Navigation validation: throw on SPA navigation failure instead of silent empty return

Also removed erroneous `.filter(r => r?.url?.includes('Followers'))` in `followers.ts` — interceptor stores response body JSON (no `url` field), pattern filtering happens at capture time.

## Test plan

- [x] `npm run typecheck` passes
- [x] `npx vitest run src/` — 213 unit tests pass
- [x] `twitter followers --user elonmusk --limit 5` — returns data
- [x] `twitter following --user elonmusk --limit 5` — returns data
- [x] `twitter search --query "opencli" --limit 5` — returns data with correct author
- [x] `twitter search --query "it's a test"` — single quote in query handled safely
- [x] `twitter notifications --limit 5` — returns data

### DIFF
--- a/src/clis/twitter/followers.ts
+++ b/src/clis/twitter/followers.ts
@@ -15,65 +15,73 @@ cli({
   func: async (page, kwargs) => {
     let targetUser = kwargs.user;
 
-    // If no user is specified, we must figure out the logged-in user's handle
+    // If no user is specified, figure out the logged-in user's handle
     if (!targetUser) {
         await page.goto('https://x.com/home');
-        // wait for home page navigation
         await page.wait(5);
-        
+
         const href = await page.evaluate(`() => {
             const link = document.querySelector('a[data-testid="AppTabBar_Profile_Link"]');
             return link ? link.getAttribute('href') : null;
         }`);
-        
+
         if (!href) {
             throw new Error('Could not find logged-in user profile link. Are you logged in?');
         }
         targetUser = href.replace('/', '');
     }
 
-    // 1. Navigate to user profile page
+    // 1. Navigate to profile page
     await page.goto(`https://x.com/${targetUser}`);
     await page.wait(3);
 
-    // 2. Inject interceptor for the followers GraphQL API
+    // 2. Install interceptor BEFORE SPA navigation.
+    //    goto() resets JS context, but SPA click preserves it.
     await page.installInterceptor('Followers');
-    
-    // 3. Click the followers link inside the profile page
-    await page.evaluate(`() => {
-        const target = '${targetUser}';
-        const link = document.querySelector('a[href="/' + target + '/followers"]');
-        if (link) link.click();
-    }`);
-    await page.wait(3);
 
-    // 4. Trigger API by scrolling
+    // 3. Click the followers link via SPA navigation (preserves interceptor).
+    //    Twitter uses /verified_followers instead of /followers now.
+    const safeUser = JSON.stringify(targetUser);
+    const clicked = await page.evaluate(`() => {
+        const target = ${safeUser};
+        const selectors = [
+            'a[href="/' + target + '/verified_followers"]',
+            'a[href="/' + target + '/followers"]',
+        ];
+        for (const sel of selectors) {
+            const link = document.querySelector(sel);
+            if (link) { link.click(); return true; }
+        }
+        return false;
+    }`);
+    if (!clicked) {
+        throw new Error('Could not find followers link on profile page. Twitter may have changed the layout.');
+    }
+    await page.wait(5);
+
+    // 4. Scroll to trigger pagination API calls
     await page.autoScroll({ times: Math.ceil(kwargs.limit / 20), delayMs: 2000 });
 
-    // 4. Retrieve data from opencli's registered interceptors
-    const allRequests = await page.getInterceptedRequests();
-    const requestList = Array.isArray(allRequests) ? allRequests : [];
-    
+    // 5. Retrieve intercepted data
+    const requests = await page.getInterceptedRequests();
+    const requestList = Array.isArray(requests) ? requests : [];
+
     if (requestList.length === 0) {
-       return [];
-    }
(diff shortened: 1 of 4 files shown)


jackwener:
Thanks for the contribution. This was a very helpful fix. Restoring the interceptor path in page.ts addresses the actual regression from the daemon migration, and the follow-up Twitter adapter updates make the INTERCEPT commands usable again on the current X frontend. I especially appreciated the careful notes in the PR description and the extra hardening around SPA navigation and JSON.stringify for injected values.

Astro-Han:
Thank you! Glad the investigation into the IIFE wrapping root cause was useful. Debugging the SPA navigation was a fun puzzle — the `pushState + popstate` trick for search/notifications and the `verified_followers` selector change were unexpected discoveries along the way.

Note: the upstream `search.ts` uses the search-input approach for SPA navigation. I kept that as-is and only added the `core.screen_name` author fix on top — should be a clean merge with future changes to that file.

Happy to help with more adapters if needed!

PR #139, opened 2026-03-20
feat(jisilu): add timeline and question adapters for jisilu.cn

yetingsky:
## Summary

- Add `jisilu timeline` command to fetch personal timeline from jisilu.cn (requires login)
- Add `jisilu question` command to fetch post content and comments from jisilu.cn

### timeline
- Fetches personal timeline from `/home/mine/#all`
- Shows post id, type (post/reply), title, author, content preview, and url
- Supports `--limit` parameter

### question
- Fetches post content and all comments
- Shows author, content, and time for each reply
- Supports `--limit` parameter to limit number of comments
- Accepts both post ID or full URL

## Usage

```bash
# Get personal timeline
opencli jisilu timeline --limit 20

# Get post details
opencli jisilu question --id 519605
opencli jisilu question --id https://www.jisilu.cn/question/519605 --limit 100
```

## Testing

Both commands have been tested successfully with various limit values.

### DIFF
--- a/src/clis/jisilu/question.ts
+++ b/src/clis/jisilu/question.ts
@@ -1,0 +1,84 @@
+/**
+ * 集思录帖子详情适配器
+ * 获取帖子正文和评论
+ */
+import { cli, Strategy } from "../../registry.js";
+import type { IPage } from "../../types.js";
+
+cli({
+  site: "jisilu",
+  name: "question",
+  description: "集思录帖子详情 (正文+评论)",
+  domain: "www.jisilu.cn",
+  strategy: Strategy.PUBLIC,
+  browser: true,
+
+  args: [
+    { name: "id", type: "str", required: true, help: "帖子ID或URL" },
+    { name: "limit", type: "int", default: 50, help: "评论数量限制" },
+  ],
+
+  columns: ["type", "author", "content", "time", "likes"],
+
+  func: async (page: IPage, kwargs) => {
+    let questionId = kwargs.id;
+    const idMatch = kwargs.id.match(/question\/(\d+)/);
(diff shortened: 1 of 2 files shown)


Astro-Han (commented):
Nice addition for the jisilu community! A few things to address:

**Code style — double quotes**

The project's biome config requires single quotes. Both files use double quotes throughout, which will likely be caught by CI or pre-commit hooks.

**`question.ts` — `id` should be positional**

Per [CONTRIBUTING.md](https://github.com/jackwener/opencli/blob/main/CONTRIBUTING.md), main target args (id, url, query) should use `positional: true`. `opencli jisilu question 519605` is more natural than `--id 519605`.

**`question.ts` — missing return value guard**

`return data` on line 79 has no `Array.isArray` check. If `evaluate` returns `null` (page not found, JS error), the output formatter will crash. Other adapters use `return Array.isArray(data) ? data : []`.

**`timeline.ts` — missing `navigateBefore: false`**

With `Strategy.COOKIE` + `domain`, the framework auto-navigates to `www.jisilu.cn` before `func` runs, then `func` navigates again to `/home/mine/#all` — double navigation wasting 2-3s.

**`likes` column — hardcoded to `0`**

Both post and replies always show `likes: 0`. If the data isn't available on the page, consider removing the column rather than showing misleading zeros.

**Tests & docs**

No E2E tests or documentation updates (README, `docs/adapters/`, SKILL.md). Per [TESTING.md](https://github.com/jackwener/opencli/blob/main/TESTING.md), `question` (PUBLIC) → `browser-public.test.ts`, `timeline` (COOKIE) → `browser-auth.test.ts`.

PR #198, opened 2026-03-21
fix(extension): skip chrome-extension:// tabs in resolveTabId fallback

### LINKED ISSUE #195: [Bug]:  attach failed: Cannot access a chrome-extension:// URL of different extension
### Description

Image

### Steps to Reproduce

1. Run `opencli ...`
2. ...
3. See error

### Expected Behavior

打不开浏览器

### OpenCLI Version

1.1.1

### Node.js Version

20.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

### LINKED ISSUE #197: attach failed: Cannot access a chrome-extension:// URL of different extension
## Bug Report

**Version:** v1.1.1

**Environment:**
- macOS (Darwin 25.3.0)
- Chrome with opencli Browser Bridge extension installed

## Problem

All browser commands fail with:

```
Error: attach failed: Cannot access a chrome-extension:// URL of different extension
```

## Root Cause (Investigated)

Through direct testing of the daemon API, I found that the `navigate` command returns a `tabId` that actually belongs to a `chrome-extension://` URL — not a regular `https://` tab:

```bash
# Navigate to xiaohongshu
curl -X POST http://localhost:19825/command \
  -H "Content-Type: application/json" \
  -d '{"id":"test1","action":"navigate","url":"https://www.xiaohongshu.com","workspace":"default"}'

# Response: {"ok":true,"data":{"title":"小红书","url":"https://www.xiaohongshu.com/explore","tabId":462299412}}

# Then exec on that tabId fails:
curl -X POST http://localhost:19825/command \
  -H "Content-Type: application/json" \
  -d '{"id":"test2","action":"exec","code":"window.scrollBy(0,300)","workspace":"default","tabId":462299412}'

# Response: {"ok":false,"error":"attach failed: Cannot access a chrome-extension:// URL of different extension"}
```

The extension appears to be **opening pages inside its own popup/extension window** (a `chrome-extension://` URL), rather than in a regular Chrome tab. Chrome's CDP (Chrome DevTools Protocol) **cannot attach to `chrome-extension://` pages**, causing all `exec` commands to fail.

The comment in `page.js` already acknowledges this issue:
> "IMPORTANT: After goto(), we remember the tabId returned by the navigate action and pass it to all subsequent commands. This avoids the issue where resolveTabId() in the extension picks a chrome:// or chrome-extension:// tab that can't be debugged."

However, the fix is incomplete — `navigate` itself returns a `tabId` pointing to a `chrome-extension://` page, so the problem is in how the extension opens the workspace, not in how the tabId is passed.

## Expected Behavior

`navigate` should open the target URL in a regular `https://` Chrome tab, and return a `tabId` that CDP can attach to for `exec` commands.

## Workaround

None currently. All commands using `autoScroll`, `evaluate`, or any `exec`-based operation fail.

Astro-Han:
## Summary

- Remove the unsafe fallback in `resolveTabId()` that returned `tabs[0]` regardless of URL type
- When no web-accessible tab exists in the automation window (e.g. a New Tab Override extension replaced `about:blank` with its own `chrome-extension://` page), always create a fresh `about:blank` tab instead
- Prevents `chrome.debugger.attach` from failing with "Cannot access a chrome-extension:// URL of different extension"

Fixes #195, fixes #197

## Root Cause

`resolveTabId()` in `extension/src/background.ts` had a fallback path (line 252) that returned the first tab in the automation window without checking its URL type. When users had a New Tab Override extension installed, `getAutomationWindow` created a window whose initial tab got replaced with a `chrome-extension://` URL. The `isWebUrl` filter correctly skipped it, but the fallback bypassed that filter, returning a tab that CDP cannot attach to.

## Test plan

- [x] Install a New Tab Override extension, verify browser commands work without "attach failed" error
- [x] Verify normal browser commands still work without any New Tab Override extension
- [x] Verify `npm run typecheck` passes

### DIFF
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -226,8 +226,8 @@ async function handleCommand(cmd: Command): Promise<Result> {
 
 // ─── Action handlers ─────────────────────────────────────────────────
 
-/** Check if a URL is a debuggable web page (not chrome:// or extension page) */
-function isWebUrl(url?: string): boolean {
+/** Check if a URL can be attached via CDP (not chrome:// or chrome-extension://) */
+function isDebuggableUrl(url?: string): boolean {
   if (!url) return false;
   return !url.startsWith('chrome://') && !url.startsWith('chrome-extension://');
 }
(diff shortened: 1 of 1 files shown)


Astro-Han:
All test plan items verified locally:

1. Installed a New Tab Override extension → `bilibili hot --limit 3` works without "attach failed" error
2. Uninstalled the extension → commands still work normally
3. `npm run typecheck` passes

PR #227, opened 2026-03-22
fix(plugin): follow symlinks when discovering plugin directories

### LINKED ISSUE #206: bug(plugin): symlink plugin directories silently skipped by discoverPlugins
## Description

`discoverPlugins()` in `src/discovery.ts:182` uses `entry.isDirectory()` to filter plugin subdirectories. However, `fs.Dirent.isDirectory()` does not follow symbolic links -- if a plugin directory is a symlink to a real directory, `isDirectory()` returns `false` and the plugin is silently skipped.

## Reproduce

```bash
# Install a plugin normally -- works
opencli plugin install github:Astro-Han/opencli-plugin-juejin

# Symlink a local plugin for development -- silently ignored
ln -s /path/to/my-plugin ~/.opencli/plugins/my-plugin
opencli my-plugin some-command
# error: unknown command 'my-plugin'
```

## Expected

Symlinked plugin directories should be discovered the same as regular directories. This is a common pattern during plugin development.

## Suggested Fix

```typescript
// discovery.ts:182
// Before:
if (!entry.isDirectory()) continue;

// After:
if (!entry.isDirectory() && !entry.isSymbolicLink()) continue;
```

Optionally, follow the symlink with `fs.stat()` to confirm the target is actually a directory.

Astro-Han:
## Summary

- Fix #206
- `discoverPlugins()` uses `entry.isDirectory()` which returns `false` for symlinks pointing to directories
- Add `|| entry.isSymbolicLink()` check so symlinked plugin directories are discovered

## Test plan

- `ln -s /path/to/plugin ~/.opencli/plugins/my-plugin` → plugin commands now discovered
- Regular (non-symlink) plugin directories continue to work unchanged

### DIFF
--- a/src/discovery.ts
+++ b/src/discovery.ts
@@ -8,33 +8,38 @@
  * 2. FALLBACK (filesystem scan): Traditional runtime discovery for development.
  */
 
-import * as fs from 'node:fs';
-import * as os from 'node:os';
-import * as path from 'node:path';
-import yaml from 'js-yaml';
-import { type CliCommand, type InternalCliCommand, type Arg, Strategy, registerCommand } from './registry.js';
-import { log } from './logger.js';
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import yaml from "js-yaml";
+import { log } from "./logger.js";
+import {
+	type Arg,
+	type CliCommand,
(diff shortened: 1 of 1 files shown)


Astro-Han:
Reopening with proper branch name.

PR #308, opened 2026-03-23
fix: remove stale SQLite file, add chatgpt platform guard, fix daemon exit code

0xsline:
## Summary

- **Remove stale SQLite file**: Add `*.db` to `.gitignore` to prevent accidental database file commits
- **ChatGPT platform guard**: Add `process.platform !== 'darwin'` check before `osascript` calls in all 4 chatgpt commands (`ask`, `read`, `new`, `status`), throwing a clear error on non-macOS platforms instead of cryptic `ENOENT`
- **Daemon exit code**: Change `process.exit(0)` to `process.exit(1)` in the `EADDRINUSE` handler so the spawning CLI correctly detects the port conflict as a failure

## Test plan

- [ ] Verify `mydatabase.db` is ignored by git (`git check-ignore mydatabase.db`)
- [ ] Run `opencli chatgpt/status` on Linux — should get clear "requires macOS" error
- [ ] Start two daemon instances — second should exit with code 1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ docs/.vitepress/cache
 .windsurf
 .claude
 .cortex
+
+# Database files
+*.db
--- a/src/clis/chatgpt/ask.ts
+++ b/src/clis/chatgpt/ask.ts
@@ -16,6 +16,10 @@ export const askCommand = cli({
   ],
   columns: ['Role', 'Text'],
   func: async (page: IPage | null, kwargs: any) => {
+    if (process.platform !== 'darwin') {
+      throw new Error('ChatGPT Desktop integration requires macOS (osascript is not available on this platform)');
+    }
+
     const text = kwargs.text as string;
     const timeout = parseInt(kwargs.timeout as string, 10) || 30;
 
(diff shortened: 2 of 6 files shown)


Astro-Han:
Nice improvement overall. The platform guard additions for the ChatGPT desktop commands make sense, and treating `EADDRINUSE` as a real failure at the daemon level also seems like the right direction.

From my reading, the one concrete gap still left in this PR is `chatgpt/send`: `ask/new/read/status` now return a clear platform-specific error on non-macOS, but `send` still calls `pbpaste`, `pbcopy`, and `osascript` directly. So Linux/Windows users would still hit the same low-level `command not found` failure on that path.

Separately, I had one non-blocking question about the daemon exit-code change. `BrowserBridge._ensureDaemon()` in `src/browser/mcp.ts` appears to start the daemon detached, `unref()` it, and then rely on `/status` polling. I couldn't find anywhere that observes the child process exit status, so this looks more like a daemon-level semantic improvement than something that changes how the spawning CLI detects port conflicts today.

PR #309, opened 2026-03-23
fix(external): replace execSync with execFileSync to prevent command injection

0xsline:
## Summary

- Replace `execSync(cmd)` with `execFileSync(binary, args)` in `installExternalCli` to prevent command injection from YAML config files
- Add `parseCommand()` helper that safely tokenizes command strings, respecting quoted segments
- Reject commands containing shell operators (`&&`, `||`, `|`, `;`, `>`, `<`, backticks) with a clear error message
- Remove unused `execSync` import

## Motivation

The `cmd` string passed to `execSync` originates from YAML config files (`install.mac`, `install.linux`, etc.). If a malicious YAML is loaded via plugin, this is a command injection vector. `execFileSync` does not invoke a shell, eliminating this risk.

## Test plan

- [ ] 309/309 existing tests pass
- [ ] `opencli doctor` still works after change
- [ ] External CLI auto-install (e.g. `gh`) still functions correctly

### DIFF
--- a/src/discovery.ts
+++ b/src/discovery.ts
@@ -16,7 +16,6 @@ import yaml from 'js-yaml';
 import { type CliCommand, type InternalCliCommand, type Arg, Strategy, registerCommand } from './registry.js';
 import { getErrorMessage } from './errors.js';
 import { log } from './logger.js';
-import { getErrorMessage } from './errors.js';
 import type { ManifestEntry } from './build-manifest.js';
 
 /** Plugins directory: ~/.opencli/plugins/ */


jackwener:
Reviewed and pushed two follow-up fixes onto this branch:

- keep the `execFileSync` hardening, but retry with `.cmd` on Windows so `npm install -g ...` keeps working
- add focused tests for `parseCommand` / Windows fallback
- restore the missing `docs/adapters/desktop/doubao-app.md` page so `docs-build` passes again

Local validation: `npm run typecheck`, `npm test`, `npm run docs:build`.

jackwener:
@0xsline Thanks for the contribution. The core security direction was right; I pushed a small follow-up for Windows  compatibility and rebased it onto current , then verified , , and . Merging now.

PR #310, opened 2026-03-23
feat(generate): implement registerCandidates to copy generated CLIs to user directory

0xsline:
## Summary

- Implement the previously stubbed `registerCandidates` function in `src/generate.ts`
- Generated CLI YAML files are now actually copied to the user's CLI directory (`~/.opencli/clis/`)
- Reads `candidates.json` from the synthesize output directory to locate candidate YAML files
- Supports filtering by name when `opts.name` is provided
- Skips existing files with a warning to avoid overwriting user customizations
- Creates the user CLI directory if it doesn't exist

## Motivation

The `registerCandidates` function was a stub returning `{ ok: true, count: 0 }`, meaning `opencli generate` and `opencli record` workflows completed without error but silently registered nothing. Users received misleading success feedback.

## Test plan

- [ ] 309/309 existing tests pass
- [ ] `opencli generate <url>` now copies YAML to `~/.opencli/clis/`
- [ ] Running generate twice does not overwrite existing CLIs

### DIFF
--- a/docs/adapters/desktop/doubao-app.md
+++ b/docs/adapters/desktop/doubao-app.md
@@ -1,0 +1,40 @@
+# Doubao App
+
+Control the **Doubao AI desktop app** (豆包) directly from the terminal via Chrome DevTools Protocol (CDP).
+
+## Prerequisites
+
+1. Install the official Doubao desktop app.
+2. Launch Doubao with the remote debugging port enabled:
+
+```bash
+# macOS
+/Applications/Doubao.app/Contents/MacOS/Doubao \
+  --remote-debugging-port=9226
+```
+
+3. Set the CDP endpoint:
+
+```bash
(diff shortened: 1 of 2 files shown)


jackwener:
Closing as stale against current main.

I checked the intended workflow against the current tree. This PR implements `registerCandidates` for the old `generate` / `record` / `synthesize` pipeline, but that top-level workflow no longer exists on main in the form this branch targets: `docs/developer/ai-workflow.md` explicitly says there is no `explore` / `synthesize` / `generate` / `cascade` command anymore, and the old related command surface has been consolidated away.

So the branch is addressing a removed path rather than a live one. If we need equivalent registration behavior in the current adapter-authoring/browser workflow, that should be proposed anew against the present implementation instead of merging this historical branch.

PR #322, opened 2026-03-23
fix: detach debugger before navigation in browser bridge

QSam2023:
## Summary

Fix a browser-bridge navigation edge case that can break follow-up CDP evaluation on sites that perform aggressive top-level navigations during auth/session transitions (observed on `creator.xiaohongshu.com`).

Before navigation, detach the debugger from the current tab so the next command re-attaches cleanly.

## Problem

Some OpenCLI commands that:
- reuse a browser bridge tab,
- call `page.goto(...)`, and then
- immediately run `page.evaluate(...)`

can fail with:

```text
Error: {"code":-32000,"message":"Inspected target navigated or closed"}
```

Observed on:

```bash
opencli xiaohongshu creator-profile -f yaml
```

The failure happened inside `Page.goto()` / subsequent `Runtime.evaluate` after navigating into Xiaohongshu creator pages.

## Root cause hypothesis

The extension keeps a cached debugger attachment for a tab. During certain top-level navigations, Chrome invalidates the inspected target, but the bridge may still attempt to reuse the previous attach state on the same tab immediately afterward.

That leaves a stale CDP attachment path and the next evaluation fails with `Inspected target navigated or closed`.

## Fix

In `extension/src/background.ts`, detach the debugger before `chrome.tabs.update(tabId, { url: targetUrl })` inside `handleNavigate(...)`.

This forces a fresh attach on the next execution step after navigation.

### Before

`opencli xiaohongshu creator-profile -f yaml` failed with:

```text
Error: {"code":-32000,"message":"Inspected target navigated or closed"}
```

### After

The same command succeeds and returns creator profile data.

Also verified working after patch:

```bash
opencli xiaohongshu creator-profile -f yaml
opencli xiaohongshu creator-stats -f yaml
opencli xiaohongshu creator-notes --limit 3 -f yaml
opencli twitter timeline --limit 2 -f yaml
```

So the change fixes the Xiaohongshu creator flow without regressing a normal Twitter/X browser command in my local verification.

## Notes

This is intentionally a small fix. If preferred, I can also add a regression test around navigation + re-evaluate behavior in the browser bridge layer.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -1,507 +1,589 @@
-const DAEMON_PORT = 19825;
-const DAEMON_HOST = "localhost";
-const DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
-const WS_RECONNECT_BASE_DELAY = 2e3;
-const WS_RECONNECT_MAX_DELAY = 6e4;
-
-const attached = /* @__PURE__ */ new Set();
+//#region src/protocol.ts
+/** Default daemon port */
+var DAEMON_PORT = 19825;
+var DAEMON_HOST = "localhost";
+var DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
+`${DAEMON_HOST}${DAEMON_PORT}`;
+/** Base reconnect delay for extension WebSocket (ms) */
+var WS_RECONNECT_BASE_DELAY = 2e3;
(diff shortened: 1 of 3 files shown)


jackwener:
Thanks @QSam2023 for the clean fix and excellent PR description! 🎉 Root cause analysis is spot-on. I pushed a small follow-up: made `detach()` async with proper `await` at all call sites. Extension build verified ✅

PR #337, opened 2026-03-24
fix: resolve 6 critical bugs from deep code review (round 2)

jackwener:
## Summary

Fixes the remaining 6 Critical issues from the deep code review (round 1 fixed security issues in PR #335).

- **execution.ts**: Guard lazy-loaded func commands against null page — throw a clear `CommandExecutionError` instead of cryptic `TypeError` when a lazy module incorrectly requires browser context
- **daemon.ts**: Fix `readBody` race condition — add `aborted` flag to prevent `req.destroy()` from triggering both reject and resolve on the same Promise (truncated data processing)
- **browser/cdp.ts**: Prevent `CDPBridge.connect()` reentry — throw if already connected instead of silently leaking the previous WebSocket
- **interceptor.ts**: Store intercept pattern in a separate global so `installInterceptor` with a new pattern updates the match condition instead of being blocked by patchGuard
- **record.ts**: Always call `cleanupEnter()` after `Promise.race` — previously only called in timeout path, leaving readline open on Enter exit. Removed dead `enterRace` variable
- **generate.ts**: Fix `undefined` entering `String.includes()` — `c.name?.toLowerCase()` returns `undefined` which gets coerced to string `"undefined"`, causing false positive candidate matches

## Test plan

- [x] All 330 unit tests pass
- [x] TypeScript typecheck passes
- [ ] Manual: verify interceptor pattern update works across multiple installInterceptor calls
- [ ] Manual: verify record mode exits cleanly on Enter press

### DIFF
--- a/src/browser/cdp.ts
+++ b/src/browser/cdp.ts
@@ -51,6 +51,8 @@ export class CDPBridge {
   private _eventListeners = new Map<string, Set<(params: unknown) => void>>();
 
   async connect(opts?: { timeout?: number; workspace?: string }): Promise<IPage> {
+    if (this._ws) throw new Error('CDPBridge is already connected. Call close() before reconnecting.');
+
     const endpoint = process.env.OPENCLI_CDP_ENDPOINT;
     if (!endpoint) throw new Error('OPENCLI_CDP_ENDPOINT is not set');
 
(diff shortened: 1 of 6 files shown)


jackwener:
Reviewed and merged. The 6 critical fixes are coherent, regression risk is low, and local validation passed (`npm run typecheck`, `npm test`). Thanks for the fast follow-up.

PR #340, opened 2026-03-24
fix: resolve 11 important bugs from deep code review

jackwener:
## Summary
- **I1**: Log pre-navigation failures in debug mode instead of silently swallowing (`execution.ts`)
- **I2**: Validate env var timeout values with NaN/negative guard (`runtime.ts`)
- **I4**: Guard `indexOf` returning -1 for unknown strategies in cascade (`cascade.ts`)
- **I5**: Fix `shouldReplaceManifestEntry` returning true for same-type entries — was causing unnecessary overwrites (`build-manifest.ts`)
- **I6**: Prevent infinite loop in `parseTsArgsBlock` cursor advancement (`build-manifest.ts`)
- **I7**: Skip redundant `Page.enable` CDP calls on repeated `goto()` (`cdp.ts`)
- **I8**: Fix `wait({time:0})` being treated as falsy due to `if (options.time)` check (`page.ts`)
- **I10**: Warn when `cookiesFile` path doesn't exist before falling back to browser cookies (`download/index.ts`)
- **I11**: Sanitize tab/newline chars in cookie name/value to prevent Netscape format corruption (`download/index.ts`)
- **I12**: Use `DEFAULT_DAEMON_PORT` constant instead of hardcoded port in error message (`mcp.ts`)
- **I15**: Log npm install failures in plugin lifecycle instead of swallowing silently (`plugin.ts`)

## Test plan
- [x] TypeScript typecheck passes (`tsc --noEmit`)
- [x] All 223 unit tests pass (`npm test`)
- [ ] Manual verification of edge cases (NaN env vars, time:0 wait, tab-in-cookie)

### DIFF
--- a/src/browser/cdp.ts
+++ b/src/browser/cdp.ts
@@ -179,13 +179,17 @@ export class CDPBridge {
 }
 
 class CDPPage implements IPage {
+  private _pageEnabled = false;
   constructor(private bridge: CDPBridge) {}
 
   /** Navigate with proper load event waiting (P1 fix #3) */
   async goto(url: string, options?: { waitUntil?: 'load' | 'none'; settleMs?: number }): Promise<void> {
-    await this.bridge.send('Page.enable');
+    if (!this._pageEnabled) {
+      await this.bridge.send('Page.enable');
+      this._pageEnabled = true;
+    }
     const loadPromise = this.bridge.waitForEvent('Page.loadEventFired', 30_000)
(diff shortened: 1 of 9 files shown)


jackwener:
Reviewed and merged. I re-ran `npm run typecheck` and `npm test` locally; both passed. The only non-blocking note is that same-type duplicate manifest entries are still "first one wins" and silent, but that does not block this PR.

Issue #329, opened 2026-03-23
Feature Request: Add support for VK (VKontakte) - The largest social network in Russia/CIS


flobo3:
Hey! First of all, thanks for this amazing tool. The browser bridge approach is exactly what we needed to bypass all these crazy anti-bot protections in 2026.

I'd like to request adding support for **VK (VKontakte)**. 

It's the #1 social network in Russia and the CIS region, with over 100 million active users monthly. It's basically the local equivalent of Facebook + Spotify + YouTube combined. 

Since VK heavily relies on dynamic CSRF tokens (`hash`) and internal APIs (`al_wall.php`, `al_feed.php`) that are hard to scrape with traditional tools, `opencli`'s browser-based approach would be absolutely perfect here.

Some basic commands that would be super useful:
- `opencli vk wall <domain>` (get posts from a user/community wall)
- `opencli vk feed` (get personal news feed)
- `opencli vk profile <domain>`

I know I could probably write a custom TS plugin for this, but considering the massive user base of VK, having it as a built-in command would be awesome for a lot of people.

Let me know what you think! I can help with testing if needed.

flobo3:
Just to add some context and numbers to back this up:

- According to recent stats, VK is the absolute leader in content publications in Russia: https://www.statista.com/statistics/284447/social-media-platforms-by-publications-russia/
- Their latest FY 2025 report shows massive engagement and MAU growth: https://www.marketscreener.com/news/vk-q4-and-fy-2025-press-release-ce7e5eded989f321

Let me know if you need any help with the implementation or testing!

Astro-Han:
+1 on this. VK's reliance on dynamic CSRF tokens and internal AJAX endpoints (`al_wall.php`) is a strong fit for opencli's browser session approach — exactly the kind of site where traditional scraping falls short.

One tool that might speed up development: **`opencli record <url>`** captures all fetch/XHR calls from a live browser session and can generate YAML adapter candidates from the recorded traffic. Running it on VK pages could reveal the internal API patterns without manual reverse-engineering — especially useful for figuring out `al_wall.php`'s response format.

Whether this lands as built-in or starts as a plugin first, happy to help review if you put together a PR.

flobo3:
Hey @jackwener and @Astro-Han! Thanks for the tips on using `opencli record`. It turned out VK uses a JSON Hijacking protection (prepends `<!--...` to JSON responses for legacy endpoints), which is why the heuristic couldn't catch the API calls automatically.

[...]

PR #350, opened 2026-03-24
docs: add opencli-plugin-vk to plugins list

flobo3:
Hey @Astro-Han! Thanks for the suggestion. 

Here is the PR adding `opencli-plugin-vk` to the plugins list in the README.

It provides commands for VK (VKontakte) to fetch walls, personal feeds, and search posts by extracting the `access_token` from the live session and hitting the internal API.

Context / Original issue: #329


Astro-Han:
Nice work! This is exactly the right approach — community adapters as plugins. Thanks for following through from the discussion in #329. LGTM 👍

PR #355, opened 2026-03-24
fix(twitter): retry transient search spa navigation

### LINKED ISSUE #353: [Bug]: twitter search intermittent SPA navigation failure
## Description

`opencli twitter search` intermittently fails with:

```
Error: SPA navigation to /search failed. Twitter may have changed its routing.
```

## Environment
- opencli version: 1.3.3
- OS: macOS 15.3 (Apple Silicon)
- Node.js: v22.22.1
- Browser Bridge: Chrome extension connected

## Reproduction

The issue is **intermittent**, not consistent. The same command may succeed on retry:

```bash
# Sometimes fails
opencli twitter search "from:somehandle" --limit 10
# Error: SPA navigation to /search failed. Twitter may have changed its routing.

# Retry succeeds
opencli twitter search "from:somehandle" --limit 10
# Returns results normally
```

Plain keyword search (`opencli twitter search "bitcoin"`) seems more reliable than `from:handle` operator searches.

## Frequency

In our automated pipeline running every 2-4 hours, opencli search has a ~42% failure rate over the past 24h (42 failures out of 99 total calls). The failures cluster during certain time windows, suggesting it may be related to Twitter's rate limiting or SPA route changes.

## Logs

```
2026-03-24 16:05:14 [WARNING] opencli failure (code 1): Error: SPA navigation to /search failed. Twitter may have changed its routing.
2026-03-24 16:05:31 [WARNING] opencli failure (code 1): Error: SPA navigation to /search failed. Twitter may have changed its routing.
```

We also see occasional `Error: Daemon is running but the Browser Extension is not connected` errors, which may be related (the extension is connected and works fine for other commands at the same time).

## Workaround

We have a fallback layer that retries with a different Twitter client when opencli fails, so this is not blocking, but the 42% failure rate is higher than expected.

## Suggestion

Could there be a retry mechanism for SPA navigation? Twitter's SPA sometimes takes longer to settle the route, and a single retry with a small delay might significantly reduce these transient failures.

Astro-Han:
## Description

Harden `twitter search` against intermittent SPA navigation failures when X stays on `/explore` after `pushState + popstate`.

Related issue:
- Fixes #353

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Verification:
- `npx vitest run src/clis/twitter/search.test.ts src/clis/twitter/timeline.test.ts src/interceptor.test.ts src/browser.test.ts`
- `npm run build`
- `npm run dev -- twitter search "from:jack" --limit 1 --format json`

Summary:
- add a one-time retry for transient `/search` SPA navigation failures
- include the final pathname in the error message for diagnostics
- add a regression test for fail-once-then-succeed navigation

### DIFF
--- a/src/clis/twitter/search.test.ts
+++ b/src/clis/twitter/search.test.ts
@@ -83,4 +83,31 @@ describe('twitter search command', () => {
     expect(page.installInterceptor).toHaveBeenCalledWith('SearchTimeline');
     expect(evaluate).toHaveBeenCalledTimes(4);
   });
+
+  it('throws with the final path after both attempts fail', async () => {
+    const command = getRegistry().get('twitter/search');
+    expect(command?.func).toBeTypeOf('function');
+
+    const evaluate = vi.fn()
+      .mockResolvedValueOnce(undefined)
+      .mockResolvedValueOnce('/explore')
+      .mockResolvedValueOnce(undefined)
+      .mockResolvedValueOnce('/login');
+
+    const page = {
+      goto: vi.fn().mockResolvedValue(undefined),
+      wait: vi.fn().mockResolvedValue(undefined),
+      installInterceptor: vi.fn().mockResolvedValue(undefined),
+      evaluate,
+      autoScroll: vi.fn().mockResolvedValue(undefined),
+      getInterceptedRequests: vi.fn(),
+    };
+
+    await expect(command!.func!(page as any, { query: 'from:alice', limit: 5 }))
+      .rejects
+      .toThrow('Final path: /login');
+    expect(page.autoScroll).not.toHaveBeenCalled();
+    expect(page.getInterceptedRequests).not.toHaveBeenCalled();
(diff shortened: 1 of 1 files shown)


Astro-Han:
Added a follow-up test commit to cover the failure path suggested in review.

What changed:
- added a second `twitter search` regression test for the case where both SPA navigation attempts fail
- the test verifies the thrown error includes the final pathname (`Final path: ...`)

What did not change:
- no production logic changes in this follow-up commit
- I did not expand this PR to `notifications` / `followers` / `following` to keep scope tight for #353

Follow-up note:
- those commands use a similar SPA navigation pattern and can be hardened separately if we see the same intermittent behavior there

PR #368, opened 2026-03-24
feat(plugin): support batch updating all installed plugins

ByteYue:
## Description
Adds the ability to update all installed plugins at once via `opencli plugin update --all`.

- Refactors `opencli plugin update [name]` to make the name optional if `--all` is provided.
- Implements `updateAllPlugins()` which iterates over all installed plugins.
- Collects success/failure states so that a single plugin update failure (e.g., network timeout) does not interrupt the entire batch process.
- Displays a summary report of successful and failed updates at the end.


jackwener:
Thanks for the contribution. I kept the core \ feature, narrowed the effective scope to plugin management only, added targeted regression coverage, and synced the plugin docs so the command is discoverable.

PR #377, opened 2026-03-24
fix(extension): avoid same-url navigation timeout

ayotme:
## Summary

   Fix a browser-bridge navigation bug where navigating to the same URL (or a canonicalized equivalent like:

   - `https://example.com`
   - `https://example.com/`

   can wait for the full 15s timeout.

   ## What changed

   - normalize URLs before comparison
   - fast-path navigation when the current tab is already at the target URL and fully loaded
   - treat navigation as complete when the tab reaches the target URL, instead of requiring a change from the previous
 URL
   - add a regression test for normalized same-URL navigation

   ## Local impact

   While reproducing this issue with browser-backed commands, I observed that same-URL navigation could add an
 unnecessary ~15s delay before the next step ran.

   After this fix, local verification showed:

   - warm runs dropping from ~18s+ to ~2.7s in the tested flow
   - cold runs also improving, while still including normal browser/session startup cost

   ## Validation

   - `npx tsc --noEmit`
   - `npm test`
   - `npm run test:adapter`
   - extension regression test for `background.ts`
   - `cd extension && npm run typecheck && npm run build`

   ## Note

   While investigating this, I also noticed some commands may do redundant command-level navigation on top of framework
 pre-navigation (for example `bilibili/hot`). I intentionally left adapter-level behavior changes out of this PR to
 keep it focused on the browser bridge bug.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -305,14 +305,13 @@ function normalizeUrlForComparison(url) {
   if (!url) return "";
   try {
     const parsed = new URL(url);
-    parsed.hash = "";
     if (parsed.protocol === "https:" && parsed.port === "443" || parsed.protocol === "http:" && parsed.port === "80") {
       parsed.port = "";
     }
-    const pathname = parsed.pathname === "/" ? "" : parsed.pathname.replace(/\/+$/, "");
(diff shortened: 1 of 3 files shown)


jackwener:
Closing per maintainer request — too many formatting/style issues in the diff. Will redo the fix in a cleaner PR.

PR #379, opened 2026-03-24
fix: update import paths in boss adapters after utils.ts rename

jackwener:
After #373 renamed  to , the import paths in  and  still referenced the old filename. This PR updates them to use the correct import path.


jackwener:
Incorrect - this PR was based on a misunderstanding. PR #373 renamed common.ts to utils.ts, not the other way around. The imports in boss/mark.ts and boss/send.ts are correct.

PR #406, opened 2026-03-25
feat: add command-level CDP overrides / 增加命令级 CDP 覆盖

coolxll:
## Summary / 摘要

This PR adds command-level CDP overrides so browser-backed commands can target different Chrome/Electron sessions without mutating global shell state.

这个 PR 为浏览器相关命令增加了命令级 CDP 覆盖参数，使不同的 Chrome / Electron 会话可以按命令单独指定，而不需要反复切换全局环境变量。

## Why / 背景与原因

Today `opencli` mainly relies on global environment variables such as `OPENCLI_CDP_ENDPOINT` and `OPENCLI_CDP_TARGET`.
That works for single-app usage, but it becomes awkward when switching between multiple Electron apps, multiple Chrome instances, or multiple debugging ports.

目前 `opencli` 主要依赖全局环境变量，例如 `OPENCLI_CDP_ENDPOINT` 和 `OPENCLI_CDP_TARGET`。
这种方式在单应用场景下可用，但一旦需要在多个 Electron 应用、多个 Chrome 实例、或者多个调试端口之间切换，就会很不方便。

Typical problems:

- only one active target can be selected through global state
- switching apps requires editing shell state repeatedly
- wrappers become necessary for otherwise simple workflows
- long-running commands such as `antigravity serve` are harder to route cleanly

典型问题包括：

- 只能通过全局状态选择一个当前目标
- 切换应用需要频繁修改 shell 环境
- 原本简单的工作流不得不额外包一层脚本
- 类似 `antigravity serve` 这种长运行命令也不容易干净地指定目标

## Changes / 修改内容

- add `--cdp-endpoint` and `--cdp-target` to browser-backed dynamic commands
- add the same command-level overrides to built-in browser commands such as `explore`, `generate`, `record`, and `cascade`
- add command-level overrides to `opencli antigravity serve`
- add runtime helpers to apply and restore CDP env overrides safely during command execution
- update Antigravity docs and skill docs to document per-command usage
- add unit tests for runtime override behavior and Commander integration

- 为浏览器相关的动态命令增加 `--cdp-endpoint` 和 `--cdp-target`
- 为 `explore`、`generate`、`record`、`cascade` 等内建浏览器命令增加同样的命令级覆盖能力
- 为 `opencli antigravity serve` 增加命令级覆盖能力
- 增加运行时 helper，在命令执行期间安全地设置并恢复 CDP 环境变量
- 更新 Antigravity 文档与技能说明，补充按命令传参的用法
- 增加对应单元测试，覆盖运行时覆盖逻辑和 Commander 接入行为

## Example / 示例

```bash
opencli antigravity status --cdp-endpoint http://127.0.0.1:9333 --cdp-target Launchpad
opencli antigravity serve --port 8082 --cdp-endpoint http://127.0.0.1:9333 --cdp-target antigravity
```

This makes multi-app workflows much easier without breaking existing env-based usage.

这样可以在不破坏现有环境变量用法的前提下，显著改善多应用场景下的使用体验。

## Verification / 验证

- `npm test`
- `npm run typecheck`

### DIFF
--- a/docs/adapters/desktop/antigravity.md
+++ b/docs/adapters/desktop/antigravity.md
@@ -16,10 +16,25 @@ Start the Antigravity desktop app with the Chrome DevTools `remote-debugging-por
 
 > Depending on your installation, the executable might be named differently, e.g., `Antigravity` instead of `Electron`.
 
-Then set the target port:
+Then either set the target port globally:
 
 ```bash
 export OPENCLI_CDP_ENDPOINT="http://127.0.0.1:9224"
+```
+
+Or pass it per command, which is usually better when you switch between multiple desktop apps or multiple CDP ports:
+
+```bash
+opencli antigravity status --cdp-endpoint http://127.0.0.1:9224
+opencli antigravity send "hello" --cdp-endpoint http://127.0.0.1:9224
+```
+
+If the endpoint exposes multiple inspectable windows, prefer the correct one per command:
+
+```bash
+opencli antigravity status \
+  --cdp-endpoint http://127.0.0.1:9224 \
+  --cdp-target antigravity
 ```
 
 ## Commands
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10,7 +10,7 @@ import chalk from 'chalk';
 import { type CliCommand, fullName, getRegistry, strategyLabel } from './registry.js';
 import { serializeCommand, formatArgSummary } from './serialization.js';
 import { render as renderOutput } from './output.js';
-import { getBrowserFactory, browserSession } from './runtime.js';
+import { extractBrowserEnvOverrides, getBrowserFactory, browserSession, withBrowserEnvOverrides } from './runtime.js';
 import { PKG_VERSION } from './version.js';
 import { printCompletionScript } from './completion.js';
 import { loadExternalClis, executeExternalCli, installExternalCli, registerExternalCli, isBinaryInstalled } from './external.js';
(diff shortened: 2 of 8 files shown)


Astro-Han:
Useful feature — per-command CDP overrides solve a real pain point for multi-app workflows. One major concern plus a few smaller items:

## Critical

**Conflicts with your PR #408**

Both PRs define the same exports (`BrowserEnvOverrides`, `withBrowserEnvOverrides`, `extractBrowserEnvOverrides`) with different signatures and modify the same files (cli.ts, runtime.ts, commanderAdapter.ts). They cannot be merged independently — whichever lands second will have type conflicts and likely lose the other's semantics during rebase.

Suggestion: consolidate both PRs into one, with a unified `BrowserEnvOverrides` type covering `cdpEndpoint`, `cdpTarget`, and `browserCdp` together. The precedence should be: explicit `--cdp-endpoint` > `--browser-cdp` auto-discovery > env var default.

## Warnings

**`runtime.test.ts` leaks env vars** — `vi.stubEnv('OPENCLI_CDP_ENDPOINT', ...)` is called without `afterEach(() => vi.unstubAllEnvs())`. This can cause flaky test failures in other suites that depend on these env vars.

**No URL validation on `--cdp-endpoint`** — invalid values like `foobar` are passed directly to the CDP connection layer, producing unfriendly errors. A basic scheme check (`http://`/`ws://`) with a clear error message would help.

**Option registration repeated 5 times** — identical `.option('--cdp-endpoint')` + `.option('--cdp-target')` + `withBrowserEnvOverrides` wrapping in explore, generate, record, cascade, and antigravity serve. Should be extracted into a shared helper (same issue noted in #408).

**Non-browser commands also wrapped** — `commanderAdapter.ts` unconditionally calls `withBrowserEnvOverrides` for all commands. A short-circuit for `browser: false` commands would avoid the unnecessary async wrapper.

coolxll:
Superseded by #408. I consolidated the command-level CDP override work into the PR #408 branch so the two changes no longer conflict.

PR #417, opened 2026-03-25
feat(fetch): add proxy support via standard proxy environment variables

mact24:
## Summary                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
  - The `fetch` pipeline step (used by all `strategy: public` adapters like HackerNews, Wikipedia, arXiv, etc.) previously called Node.js native `fetch` directly, which ignores system proxy settings                                                                                                                      
  - Added proxy support using `undici` `ProxyAgent`, reading standard proxy environment variables at runtime                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
  ## How it works                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
  When any of the following env vars are set, Node-side `fetch` calls are routed through the proxy. Browser-side fetch paths are unaffected (Chrome uses its own proxy settings).                                                                                                                                           
                                                                  
  ```bash                                                                                                                                                                                                                                                                                                                   
  HTTPS_PROXY=http://127.0.0.1:7890 opencli hackernews top        
  HTTP_PROXY=http://127.0.0.1:7890 opencli arxiv search --query "LLM"                                                                                                                                                                                                                                                       
  ALL_PROXY=socks5://127.0.0.1:7891 opencli wikipedia search --query "AI"                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
  Supported variables (in priority order): HTTPS_PROXY, https_proxy, HTTP_PROXY, http_proxy, ALL_PROXY, all_proxy                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
  Changes                                                                                                                                                                                                                                                                                                                   
                                                                  
  - src/pipeline/steps/fetch.ts: added getProxyFetch() helper that returns a proxy-aware fetch when env vars are present, falls back to native fetch otherwise                                                                                                                                                              
  - package.json / package-lock.json: added undici dependency (the same library that powers Node.js built-in fetch)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "commander": "^14.0.3",
         "js-yaml": "^4.1.0",
         "turndown": "^7.2.2",
+        "undici": "^7.24.5",
         "ws": "^8.18.0"
       },
       "bin": {
@@ -3512,6 +3513,15 @@
       },
       "engines": {
         "node": ">=14.17"
+      }
+    },
+    "node_modules/undici": {
+      "version": "7.24.5",
+      "resolved": "https://registry.npmmirror.com/undici/-/undici-7.24.5.tgz",
+      "integrity": "sha512-3IWdCpjgxp15CbJnsi/Y9TCDE7HWVN19j1hmzVhoAkY/+CJx449tVxT5wZc1Gwg8J+P0LWvzlBzxYRnHJ+1i7Q==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20.18.1"
       }
     },
     "node_modules/undici-types": {
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "commander": "^14.0.3",
     "js-yaml": "^4.1.0",
     "turndown": "^7.2.2",
+    "undici": "^7.24.5",
     "ws": "^8.18.0"
   },
   "devDependencies": {
--- a/src/pipeline/steps/fetch.ts
+++ b/src/pipeline/steps/fetch.ts
@@ -8,8 +8,17 @@ import type { IPage } from '../../types.js';
 import { render } from '../template.js';
 
 import { isRecord, mapConcurrent } from '../../utils.js';
+import { ProxyAgent, fetch as undiciFetch } from 'undici';
 
-
+/** Returns a fetch function that routes through the system proxy if configured. */
+function getProxyFetch(): typeof fetch {
+  const proxyUrl = process.env.HTTPS_PROXY ?? process.env.https_proxy
+    ?? process.env.HTTP_PROXY ?? process.env.http_proxy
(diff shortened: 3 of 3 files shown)


Astro-Han:
Good idea — proxy support for Chinese users is a real need. Found some issues with the current implementation:

## Critical

**SOCKS5 will crash at runtime**

undici `ProxyAgent` only supports HTTP/HTTPS CONNECT proxies. `ALL_PROXY=socks5://127.0.0.1:7891` (common with Clash, V2Ray, etc.) will throw a `TypeError` on first request. SOCKS5 support is still an [open issue](https://github.com/nodejs/undici/issues/2224) in undici.

**Proxy selection ignores request scheme**

`getProxyFetch()` picks one proxy URL and uses it for all requests. When both `HTTP_PROXY` and `HTTPS_PROXY` are set, `http://` requests incorrectly go through `HTTPS_PROXY`. Also `NO_PROXY` is not respected.

**Recommended fix**: Replace the manual proxy resolution with undici's built-in [`EnvHttpProxyAgent`](https://github.com/nodejs/undici/blob/main/docs/docs/api/EnvHttpProxyAgent.md), which handles scheme routing, `NO_PROXY`, and env var priority correctly out of the box. For SOCKS5, either detect and skip with a warning, or add `socks-proxy-agent` as a fallback.

## Warnings

**Node version conflict** — `undici@^7.24.5` requires Node >=20.18.1, but the project declares `>=20.0.0` in `package.json`.

**ProxyAgent created per-request** — `getProxyFetch()` creates a new `ProxyAgent` on every call. In batch scenarios (`mapConcurrent`), this means N separate connection pools with no cleanup. Should be cached as a module-level singleton.

**Only covers YAML pipeline fetch** — 10+ TS adapters call Node.js `fetch()` directly (weread, wikipedia, arxiv, google, substack, etc.) and won't benefit. Consider `undici.setGlobalDispatcher()` as a follow-up, or at least note this as a known limitation.

## Minor

- No validation for malformed proxy URLs — will throw raw undici errors instead of `CliError`
- `package-lock.json` resolved URLs point to `registry.npmmirror.com` instead of `registry.npmjs.org`
- No tests added

jackwener:
Closing as stale/superseded by current main.

I checked this against the live code rather than only the PR title. The underlying feature is already present on main through the newer network stack:

- `src/node-network.ts` now detects the standard proxy env vars (`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY`)
- it builds an `EnvHttpProxyAgent` dispatcher for Node-side requests
- there is current regression coverage in `src/node-network.test.ts` and `src/download/index.test.ts` for proxy env handling

So the public-fetch proxy support this PR was adding has already landed through a later implementation, and this older branch is stale.

PR #427, opened 2026-03-25
fix(chatwise): support latest electron composer flow

### LINKED ISSUE #336: [Bug]: 如何连接chatwise
### Description

> opencli chatwise ask "hello"
Error: Error: Could not find input
    at <anonymous>:7:30
    at <anonymous>:16:9

### Steps to Reproduce

> opencli chatwise ask "hello"
Error: Error: Could not find input
    at <anonymous>:7:30
    at <anonymous>:16:9

### Expected Behavior

> opencli chatwise ask "hello"
Error: Error: Could not find input
    at <anonymous>:7:30
    at <anonymous>:16:9

### OpenCLI Version

1.0.0

### Node.js Version

20.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

ahpxex:
## Summary

This PR fixes the current ChatWise Electron flow on the latest desktop app, and also restores discovery/manifest support for shared desktop factory commands.

### What changed

- fix ChatWise composer targeting for the latest Electron UI
  - ChatWise now exposes multiple `contenteditable` editors
  - the previous adapter often picked the wrong editor (`Optional description` / `User Context Document`) instead of the main chat composer
  - the adapter now scores candidates using placeholder text, editor type, and viewport position, then targets the main composer reliably
- make `chatwise ask` wait for the actual message wrapper content instead of returning too early from nested message/timestamp nodes
- restore discovery + manifest support for desktop adapters built via shared factories
  - commands like `status`, `new`, and `screenshot` generated via `makeStatusCommand` / `makeNewCommand` / `makeScreenshotCommand` are now detected in both filesystem discovery and manifest compilation

## Why

Issue #336 reports `opencli chatwise ask` failing with `Could not find input`.

On the current ChatWise Electron app, the root cause is not just a missing input. The app now renders multiple editors, and the old fallback logic selected the wrong one. In source/dev mode there was also a second problem: shared desktop factory commands were not being discovered at all, making `chatwise status` unavailable and making debugging harder.

## Validation

- `bunx vitest run src/build-manifest.test.ts src/discovery.test.ts`
- `bunx tsc --noEmit`
- real local ChatWise Electron app verification via CDP:
  - `chatwise status` connects successfully
  - `chatwise new` succeeds
  - `chatwise ask '贵州有哪些地级市（州）' --timeout 60` sends successfully and returns the generated answer

Fixes #336

### DIFF
--- a/clis/chatwise/ask.js
+++ b/clis/chatwise/ask.js
@@ -1,5 +1,11 @@
 import { cli, Strategy } from '@jackwener/opencli/registry';
-import { selectorError } from '@jackwener/opencli/errors';
+import { selectorError, TimeoutError } from '@jackwener/opencli/errors';
+import {
+    buildChatwiseInjectTextJs,
+    buildChatwiseMessageCountJs,
+    buildChatwiseResponseAfterJs,
+    normalizeTimeout,
+} from './utils.js';
 export const askCommand = cli({
     site: 'chatwise',
     name: 'ask',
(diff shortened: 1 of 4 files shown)


Astro-Han:
Thanks for fixing the ChatWise composer targeting — the scoring approach is a solid idea. Found a few issues:

## Critical

**Message count mismatch in `ask.ts`**

`beforeLen` counts all message wrapper DOM nodes (including empty/skeleton ones), but the polling loop filters with `filter(Boolean)` before comparing against `prevLen`. If ChatWise renders any empty wrappers or loading placeholders, the filtered count will never exceed `beforeLen`, causing the poll to time out and return empty.

These two counts need to use the same basis — either both count raw nodes, or both count non-empty text nodes.

## Warnings

**Stability check may truncate streaming responses**

The "two consecutive identical polls = done" logic can trigger on normal pauses during streaming generation. With a 2s polling interval, a brief thinking pause would be mistaken for completion. Consider requiring more consecutive stable polls, or using an increasing interval.

**`build-manifest.ts` drops `extra` from shared factory commands**

The factory branch reconstructs `site/name/description/args/columns` but omits the `extra` parameter. Commands like `chatwise status` use `extra.requiredEnv` to enforce `OPENCLI_CDP_ENDPOINT` — under the manifest fast path, this check is silently skipped.

yy1218123 (commented):
@

PR #408, opened 2026-03-25
feat: support browser-level CDP websocket fallback

coolxll:
## Summary / 摘要

This PR improves Chrome CDP connectivity for newer remote debugging flows and adds a practical direct-CDP mode for browser-backed commands.

这个 PR 一方面增强了 opencli 对新版 Chrome 远程调试流程的兼容性，另一方面为浏览器型命令补齐了可实际使用的直连 CDP 模式。

## What This PR Includes / 本 PR 包含的内容

1. Browser-level websocket fallback when classic `/json` discovery is unavailable.
2. Explicit `--browser-cdp` mode for browser-backed commands to bypass the daemon / extension path.
3. Global `OPENCLI_BROWSER_CDP=1` default to enable the same behavior without repeating the flag.
4. Safer auto mode that opens a fresh page by default, instead of accidentally attaching to an unrelated existing tab.

1. 当经典 `/json` 发现路径不可用时，回退到浏览器级 websocket。
2. 为浏览器型命令增加显式 `--browser-cdp` 模式，直接绕过 daemon / extension 链路。
3. 增加全局 `OPENCLI_BROWSER_CDP=1`，无需每次重复传 flag。
4. 改进 auto 模式，默认新开干净页面，避免误附着到不相关的现有标签页。

## Official Reference / 官方背书

This change is aligned with Chrome's newer DevTools / MCP direction documented by the Chrome team:
https://developer.chrome.com/blog/chrome-devtools-mcp-debug-your-browser-session?hl=zh-cn

The official post shows that newer Chrome debugging sessions can rely on browser-session level connectivity rather than the older `/json` discovery path alone.

这个改动和 Chrome 官方介绍的新 DevTools / MCP 方向是一致的：
https://developer.chrome.com/blog/chrome-devtools-mcp-debug-your-browser-session?hl=zh-cn

官方文章说明，较新的 Chrome 调试会话可以围绕浏览器会话级连接来工作，而不再只依赖旧的 `/json` 发现路径。

## Why / 背景与原因

In newer Chrome remote debugging setups, the debugging port can be active while classic endpoints like `/json` return `404`.
In that situation, opencli's previous CDP implementation could not connect to ordinary Chrome pages even though the browser debugging session was actually available.

Also, for normal Chrome website automation, users may want an explicit or global command mode that clearly routes through browser CDP and does not depend on whether the daemon / extension path is working.
During local testing, auto-attaching to an existing browser tab could also land on the wrong page context, which made website commands fail with unrelated page errors.

[...]

PR #439, opened 2026-03-25
fix: attach browser-level CDP endpoints to a page target

huluma1314:
## Summary
- auto-discover local Chrome CDP endpoints from DevToolsActivePort when OPENCLI_CDP_ENDPOINT is unset or set to auto
- treat browser-level websocket endpoints as attachable roots and bind them to a real page target before sending Page/Runtime commands
- make doctor report CDP mode explicitly and cover browser-level target selection with unit tests

## Why
On macOS Chrome exposing DevToolsActivePort, OpenCLI can discover a browser-level websocket like `ws://127.0.0.1:9222/devtools/browser/<id>` even when the standard `/json` target listing is unavailable. Before this change, CDP mode assumed it already had a page-level target, so browser commands and `opencli doctor` could fail before any page interaction.

## Validation
- `npm test -- src/browser/cdp.test.ts src/browser.test.ts`
- `npm run build`

### DIFF
--- a/src/browser/cdp.test.ts
+++ b/src/browser/cdp.test.ts
@@ -36,7 +36,7 @@ vi.mock('ws', () => ({
   WebSocket: MockWebSocket,
 }));
 
-import { CDPBridge } from './cdp.js';
+import { CDPBridge, __test__ } from './cdp.js';
 
 describe('CDPBridge cookies', () => {
   beforeEach(() => {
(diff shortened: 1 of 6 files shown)


Astro-Han:
Hi! PR #408 covers similar ground: browser-level WebSocket discovery, page target attach, and DevToolsActivePort fallback. It's been through a few rounds of review and might merge soon. Might be worth checking if there's a gap this PR addresses that #408 doesn't.

Another concrete risk here: `CDPBridge.connect()` now puts `attachToInspectableTarget()` inside the same broad `try/catch` as the best-effort `Page.enable` setup in `src/browser/cdp.ts` (around lines 79-87). If the browser-level websocket opens but `Target.getTargets`, `Target.createTarget`, or `Target.attachToTarget` fails, that exception is swallowed and `connect()` still resolves a `CDPPage`. That leaves callers with a "connected" page that was never attached to a target, so the first real command fails later and the root cause is hidden. I think attach failure should still reject `connect()`, with only the post-attach niceties staying best-effort.

PR #445, opened 2026-03-25
fix: use pathToFileURL for dynamic imports

yanCode:
## Summary
- replace raw `file://${...}` dynamic imports with `pathToFileURL(...).href`
- apply the fix to both filesystem discovery and manifest lazy-loading
- add a regression test covering local paths that need URL escaping

## Why
`vitest run` currently emits `Invalid file URL` warnings because local paths are interpolated directly into `file://` import specifiers. This is brittle for spaces, `#`, and other characters that require escaping, and it is especially fragile on Windows.

Using `pathToFileURL(...).href` produces valid import URLs consistently across platforms.

## Validation
- `npm test -- --run src/engine.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/discovery.ts
+++ b/src/discovery.ts
@@ -11,6 +11,7 @@
 import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
+import { pathToFileURL } from 'node:url';
 import yaml from 'js-yaml';
 import { type CliCommand, type InternalCliCommand, type Arg, Strategy, registerCommand } from './registry.js';
 import { log } from './logger.js';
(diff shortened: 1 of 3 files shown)


jackwener:
Thanks for the fix. We reviewed this again on March 26, 2026 and the underlying  issue is already fixed on , including the dynamic import call sites that this PR targets. Rebased onto current , this PR conflicts because the core change has effectively already landed. We are going to close this PR to avoid duplicate history, but the problem you identified was valid and the fix direction was helpful. Thanks for catching it.

jackwener:
Thanks for the fix. We reviewed this again on March 26, 2026 and the underlying pathToFileURL(...).href issue is already fixed on main, including the dynamic import call sites this PR targets. Rebased onto current main, this PR conflicts because the core change has effectively already landed. We are closing this PR to avoid duplicate history, but the issue you identified was valid and the fix direction was helpful. Thanks for catching it.

PR #457, opened 2026-03-26
feat: add bilibili/comments, xiaohongshu/comments commands + rate-limiter plugin docs

mikkley:
## Summary

- **`bilibili/comments <bvid>`** — fetches top-level replies via `/x/v2/reply/main` with WBI signing. Resolves bvid → aid via `/x/web-interface/view`, then calls the reply API with `signed: true`. No DOM dependency, same pattern as `bilibili/search`.

- **`xiaohongshu/comments <note-id>`** — DOM extraction from the note detail page (`/explore/<id>`). XHS doesn't expose a public comments API, so we scroll the note container and scrape `.parent-comment` elements. Handles two edge cases:
  - Login-wall detection (throws `AuthRequiredError`)  
  - Zero-like counts: XHS shows the text "赞" in the `.count` element when likes = 0; the extractor converts that to `0` instead of a string.

- **`docs/advanced/rate-limiter-plugin.md`** — documents the `onAfterExecute` hook and shows a copy-paste rate-limiter plugin that adds a random 5–30 s sleep after every browser-platform command. Useful for long-running scraping jobs. Controlled via `OPENCLI_RATE_MIN`, `OPENCLI_RATE_MAX`, and `OPENCLI_NO_RATE=1` (disable for local dev).

## Test plan

- [ ] `npm test` — all 294 existing tests pass; 10 new tests added (5 per command)
- [ ] `npx tsc --noEmit` — no type errors
- [ ] Manual: `opencli bilibili comments BV1WtAGzYEBm --limit 5` returns ranked comment rows
- [ ] Manual: `opencli xiaohongshu comments <note-id> --limit 5` returns comment rows when logged in
- [ ] Manual: rate-limiter local install prints `[rate-limiter] xiaohongshu: sleeping Xs` to stderr

### DIFF
--- a/src/clis/xiaohongshu/comments.test.ts
+++ b/src/clis/xiaohongshu/comments.test.ts
@@ -41,13 +41,14 @@ describe('xiaohongshu comments', () => {
       ],
     });
 
-    const result = await command!.func!(page, { 'note-id': '69aadbcb000000002202f131', limit: 5 });
+    const result = (await command!.func!(page, { 'note-id': '69aadbcb000000002202f131', limit: 5 })) as any[];
 
     expect((page.goto as any).mock.calls[0][0]).toContain('/explore/69aadbcb000000002202f131');
     expect(result).toEqual([
       { rank: 1, author: 'Alice', text: 'Great note!', likes: 10, time: '2024-01-01' },
       { rank: 2, author: 'Bob', text: 'Very helpful', likes: 0, time: '2024-01-02' },
     ]);
+    expect(result[0]).not.toHaveProperty('loginWall');
   });
 
   it('strips /explore/ prefix from full URL input', async () => {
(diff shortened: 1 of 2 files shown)


Astro-Han:
Both comments commands follow existing patterns nicely: bilibili reuses WBI signing, xiaohongshu goes with DOM extraction. A few things I noticed:

**bilibili/comments error handling**

The bvid resolution failure throws a bare `Error` instead of a `CliError` subclass (existing `subtitle.ts` uses `CliError`). Also, responses from `/x/web-interface/view` and `/x/v2/reply/main` don't check `payload.code`, so rate-limiting or auth failures would silently return an empty array instead of a clear error.

**xiaohongshu/comments can't consume search output**

`xiaohongshu/search` outputs URLs like `/search_result/...?xsec_token=...`, but comments only recognizes `/explore/<id>` and `/note/<id>`. Passing a search result URL to comments would fail to parse.

**Output field leak**

Each comment object includes `loginWall: false`, which leaks into `--format json` output and doesn't match the declared `columns`.

**Empty result semantics**

xiaohongshu throws `EmptyResultError` on zero comments, while bilibili returns an empty array. Might be worth aligning.

PR #487, opened 2026-03-27
fix(plugin): detect symlinked monorepo sub-plugins in discoverPlugins

ByteYue:
## Problem

`discoverPlugins()` uses `entry.isDirectory()` to filter plugin directories, but monorepo sub-plugins are installed as symlinks pointing into `~/.opencli/monorepos/`. On most Node.js versions, `isDirectory()` returns `false` for symlinks, causing monorepo plugin commands to be **silently skipped** during discovery.

## Fix

Add `entry.isSymbolicLink()` check so symlinked plugin directories are properly discovered and their commands registered.

```diff
```

## Testing

All 337 unit tests pass.


jackwener:
Thanks for the contribution! I pushed a small follow-up fix to keep symlinked monorepo plugins discoverable without letting broken symlinks turn startup into a failure. The overall discovery direction here is solid.

PR #512, opened 2026-03-27
fix(network): honor proxy env for node requests

### LINKED ISSUE #510: [Bug]: Node-side fetch ignores proxy env vars and fails on proxy-required sites
### Description

opencli does not honor http_proxy / https_proxy for Node-side fetch requests, so commands that require the configured proxy fail with TypeError: fetch failed even though the same URLs work with curl in the same environment.

### Steps to reproduce

  1. Run export http_proxy=http://127.0.0.1:7897 && export https_proxy=http://127.0.0.1:7897
  2. Run opencli v2ex latest -f json
  3. See error

  ### Actual behavior

  opencli fails with a Node fetch error on sites that require the configured proxy:

  TypeError: fetch failed
      at node:internal/deps/undici/undici:15482:13
  ...
  [cause]: AggregateError [ETIMEDOUT]

  In my case, opencli v2ex hot --verbose also failed earlier with the same Node-side fetch failed stack.

  ### Expected behavior

  opencli should honor http_proxy / https_proxy automatically, the same way curl does, and successfully return V2EX results.

  ### Environment

  - opencli: 1.5.0
  - Node: 24.2.0
  - OS: macOS
  - Proxy env:
      - http_proxy=http://127.0.0.1:7897
      - https_proxy=http://127.0.0.1:7897

  ### Additional context

  This does not affect every command. For example, opencli apple-podcasts top works fine in the same shell, which suggests the bug only shows up for endpoints that need the configured proxy to be reachable.

  Also, plain curl to V2EX works in the same environment, so the issue seems specific to Node-side fetch not using the proxy env vars by default.

  ### Workaround

  Running with NODE_USE_ENV_PROXY=1 makes the same request succeed:

  NODE_USE_ENV_PROXY=1 opencli v2ex latest -f json

### Expected Behavior

opencli should automatically use the configured proxy environment variables for Node-side network requests and return command results successfully instead of timing out with fetch failed.

### OpenCLI Version

1.5.0

### Node.js Version

Other

### Operating System

macOS

### Logs / Screenshots

```shell

```

WSL0809:
Fixes #510.

Respect proxy env vars for Node-side requests and downloads, while bypassing loopback addresses.

### DIFF
--- a/src/node-network.test.ts
+++ b/src/node-network.test.ts
@@ -21,11 +21,24 @@ describe('node network proxy decisions', () => {
     });
   });
 
+  it('falls back to HTTP_PROXY for https traffic when HTTPS_PROXY is absent', () => {
+    const decision = decideProxy(
+      new URL('https://www.v2ex.com/api/topics/latest.json'),
+      { HTTP_PROXY: 'http://127.0.0.1:7897' },
+    );
+
+    expect(decision).toEqual({
+      mode: 'proxy',
+      proxyUrl: 'http://127.0.0.1:7897',
+    });
+  });
+
   it('bypasses proxies for loopback addresses', () => {
     const env = { https_proxy: 'http://127.0.0.1:7897', http_proxy: 'http://127.0.0.1:7897' };
 
     expect(decideProxy(new URL('http://127.0.0.1:19825/status'), env)).toEqual({ mode: 'direct' });
     expect(decideProxy(new URL('http://localhost:19825/status'), env)).toEqual({ mode: 'direct' });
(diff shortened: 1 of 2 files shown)


jackwener:
Thanks for the fix. I reviewed the proxy handling, pushed a couple of follow-up improvements onto your branch, and merged it.

What went in:
- honors proxy env for Node-side requests and downloads
- fixes `NO_PROXY=host:80/443` default-port matching
- falls back from `HTTPS_PROXY` to `HTTP_PROXY` for HTTPS requests when needed
- keeps loopback addresses bypassed, including IPv6
- delegates the env-proxy semantics to `undici`'s `EnvHttpProxyAgent`

I also re-ran targeted tests and typecheck locally before merging. Thanks again for the contribution.

PR #549, opened 2026-03-28
feat：add streaming API interceptor

AstaTus:
## Description
feat: add streaming API interceptor for background tab compatibility
Adds streaming-capable network interception that works in background
browser tabs where requestAnimationFrame is throttled. This enables
capturing streaming API responses (e.g., Gemini's StreamGenerate XHR)
without relying on DOM rendering.

Key changes:
- New generateStreamingInterceptorJs() with fetch ReadableStream + XHR
  onprogress/readystatechange dual capture
- New waitForStreamCaptureJs() with configurable timeout and done-waiting
- sendCommand() now accepts optional timeoutMs (default 30s)
- IPage extended with installStreamingInterceptor/getStreamedResponses/
  waitForStreamCapture methods
- New stream-intercept pipeline step for YAML pipelines
- XHR completion uses full responseText overwrite via readystatechange
  to guarantee data completeness

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Related issue:

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

### DIFF
--- a/src/clis/douyin/draft.test.ts
+++ b/src/clis/douyin/draft.test.ts
@@ -96,7 +96,9 @@ function createPageMock(
     installInterceptor: vi.fn().mockResolvedValue(undefined),
     getInterceptedRequests: vi.fn().mockResolvedValue([]),
     waitForCapture: vi.fn().mockResolvedValue(undefined),
-    screenshot: vi.fn().mockResolvedValue(''),
+    installStreamingInterceptor: vi.fn().mockResolvedValue(undefined),
+    getStreamedResponses: vi.fn().mockResolvedValue({ text: '', events: [], done: false, errors: [] }),
+    waitForStreamCapture: vi.fn().mockResolvedValue(undefined),    screenshot: vi.fn().mockResolvedValue(''),
     setFileInput: vi.fn().mockResolvedValue(undefined),
     ...overrides,
   };
--- a/src/clis/facebook/search.test.ts
+++ b/src/clis/facebook/search.test.ts
@@ -42,7 +42,9 @@ function createMockPage(): IPage {
     installInterceptor: vi.fn(),
     getInterceptedRequests: vi.fn().mockResolvedValue([]),
     waitForCapture: vi.fn().mockResolvedValue(undefined),
-    screenshot: vi.fn().mockResolvedValue(''),
+    installStreamingInterceptor: vi.fn().mockResolvedValue(undefined),
+    getStreamedResponses: vi.fn().mockResolvedValue({ text: '', events: [], done: false, errors: [] }),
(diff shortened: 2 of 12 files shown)


AstaTus:
after this pr merge , I will update opencli-plugin-gemini-web ，the new opencli-plugin-gemini-web uses streaming API interceptor

Astro-Han:
Follow-up to my earlier review: there is also some state leakage around the XHR SSE buffer.

In `generateStreamingInterceptorJs()`, the XHR path accumulates partial SSE chunks in `window.__opencli_stream_sse_buf` during `progress`, and only clears that buffer on the `readystatechange/load` completion path.

But `generateReadStreamJs()` does not clear `${prefix}_sse_buf`, and a new matching request also does not reset it before starting a fresh capture.

That means if a caller reads early (`waitForDone: false`) or the previous XHR stream stops in a non-clean path, the next matching request can inherit leftover partial SSE data from the previous one and parse mixed events.

I think `${prefix}_sse_buf` should be treated as part of the capture state and reset both:
- when a new matching request starts, and
- when `generateReadStreamJs()` clears the rest of the streaming state.

There is also a compatibility bug in the SSE parser itself: both the fetch path and XHR path only look for `\n\n` as the event boundary (`src/interceptor.ts`, around the `indexOf('\\n\\n')` checks).

That works for LF-only streams, but a compliant SSE server can also use CRLF line endings. With a stream like `data: first\r\n\r\ndata: second\r\n\r\n`, this code never finds a boundary during accumulation, then flushes the whole buffer at the end and parses it as one merged event instead of two.

So for some perfectly valid SSE responses, `returnEvents: true` will return the wrong event structure even when the raw stream content is fine. I think the delimiter handling needs to normalize `\r\n` first, or split on both LF and CRLF boundaries.

AstaTus:
1. add unit tests for generateStreamingInterceptorJs generateReadStreamJs, waitForStreamCaptureJs 
2. single-capture pattern, state reset on new request is expected
3. generateReadStreamJs added clear parameter (default true);  IPage.getStreamedResponses accepts { clear?: boolean } 
4.  _sse_buf cleared in 4 places: install init, fetch new request, XHR new request, generateReadStreamJs 
5. SSE parser only splits on \n\n, breaks on CRLF (\r\n\r\n)：  \r\n → \n normalization in 3 places: fetch SSE buffer, XHR SSE buffer, __parseSse line split

AstaTus:
1. change PR titile
2. update  ts-adapter.md and yaml-adapter.md‎ about stream API interceptor

PR #569, opened 2026-03-28
fix(extension): use offscreen document for persistent WebSocket connection

lyfuci:
## Problem

Chrome MV3 Service Workers are suspended by the browser after a short idle period (~30s). When this happens, the WebSocket connection to the opencli daemon drops silently. Although a `keepalive` alarm fires every ~24s to wake the SW and call `connect()`, there is a gap window where:

1. SW is suspended → WS drops
2. Alarm fires → SW wakes → reconnects
3. During that gap, any CLI command gets "Extension not connected"

Users experience this as the extension randomly disconnecting, requiring them to manually click the extension icon to wake it.

## Root Cause

The WebSocket connection lives inside the Service Worker. Chrome's MV3 Service Worker lifecycle does not support persistent connections — it is designed to be ephemeral.

## Fix

Move the WebSocket connection into a `chrome.offscreen` document. Offscreen documents have a stable lifetime tied to the browser window (not the SW), so the connection survives SW sleep/wake cycles.

The Service Worker becomes a thin relay layer:
- Receives commands from the daemon CLI
- Forwards them to the offscreen document via `chrome.runtime.sendMessage`
- Offscreen executes them and sends results back

## Changes

| File | Change |
|------|--------|
| `extension/src/offscreen.ts` | ✨ New — WebSocket host with auto-reconnect logic |
| `extension/offscreen.html` | ✨ New — Offscreen document HTML entry point |
| `extension/src/background.ts` | 🔧 Refactored — SW delegates WS ops to offscreen |
| `extension/manifest.json` | 🔧 Added `offscreen` permission |
| `extension/vite.config.ts` | 🔧 Added `offscreen` build entry |

## Testing

Tested on Linux (Chrome 144) with the browser idle for 2+ hours. Extension remained connected throughout — no manual reconnection required.

```
opencli doctor
[OK] Daemon: running on port 19825
[OK] Extension: connected (v1.5.5)
[OK] Connectivity: connected in 0.2s
```

Previously this would show `[MISSING] Extension: not connected` after Chrome suspended the SW.

### DIFF
--- a/extension/dist/assets/protocol-Z52ThYIj.js
+++ b/extension/dist/assets/protocol-Z52ThYIj.js
@@ -1,0 +1,8 @@
+const DAEMON_PORT = 19825;
+const DAEMON_HOST = "localhost";
+const DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
+const DAEMON_PING_URL = `http://${DAEMON_HOST}:${DAEMON_PORT}/ping`;
+const WS_RECONNECT_BASE_DELAY = 2e3;
+const WS_RECONNECT_MAX_DELAY = 6e4;
+
+export { DAEMON_PING_URL as D, WS_RECONNECT_BASE_DELAY as W, DAEMON_WS_URL as a, WS_RECONNECT_MAX_DELAY as b };
--- a/extension/dist/offscreen.js
+++ b/extension/dist/offscreen.js
@@ -1,0 +1,100 @@
+const DAEMON_PORT = 19825;
+const DAEMON_HOST = "localhost";
+const DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
+const WS_RECONNECT_BASE_DELAY = 2e3;
+const WS_RECONNECT_MAX_DELAY = 6e4;
+
+let ws = null;
+let reconnectTimer = null;
+let reconnectAttempts = 0;
+const MAX_EAGER_ATTEMPTS = 6;
+function sendLog(level, msg) {
+  chrome.runtime.sendMessage({ type: "log", level, msg, ts: Date.now() }).catch(() => {
+  });
+}
+const _origLog = console.log.bind(console);
+const _origWarn = console.warn.bind(console);
+const _origError = console.error.bind(console);
+console.log = (...args) => {
+  _origLog(...args);
+  sendLog("info", args.map((a) => typeof a === "string" ? a : JSON.stringify(a)).join(" "));
+};
+console.warn = (...args) => {
+  _origWarn(...args);
+  sendLog("warn", args.map((a) => typeof a === "string" ? a : JSON.stringify(a)).join(" "));
+};
+console.error = (...args) => {
+  _origError(...args);
+  sendLog("error", args.map((a) => typeof a === "string" ? a : JSON.stringify(a)).join(" "));
+};
+async function connect() {
+  if (ws?.readyState === WebSocket.OPEN || ws?.readyState === WebSocket.CONNECTING) return;
+  try {
+    ws = new WebSocket(DAEMON_WS_URL);
+  } catch {
+    scheduleReconnect();
+    return;
+  }
+  ws.onopen = () => {
+    console.log("[opencli/offscreen] Connected to daemon");
+    reconnectAttempts = 0;
+    if (reconnectTimer) {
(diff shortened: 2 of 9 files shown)


Astro-Han (commented):
Clean architecture — offscreen document for persistent WS is the right call for MV3, and the legacy fallback path is a nice safety net. Two minor observations:

1. `pendingFrames` in offscreen.ts has no cap. If the SW stays unreachable for a while the array could grow unbounded. A simple max (e.g. 100) with oldest-first eviction would be a cheap safeguard.

2. `forceLegacyTransport` is set permanently on first `ensureOffscreen()` failure. Might be worth resetting it periodically (e.g. on keepalive alarm) so a transient error doesn't lock out offscreen for the entire session.

Astro-Han:
Follow-up to my earlier review: I think there is one more reliability gap in the offscreen transport.

In `background.ts`, the `ws-message` branch sends `sendResponse({ ok: true })` immediately, before `handleCommand()` finishes and before the result is sent back to the daemon. In `offscreen.ts`, `flushPendingFrames()` treats that `ok` as delivery success and removes the frame from `pendingFrames` right away.

That means "accepted by the Service Worker" is currently being treated as "fully delivered", but those are not the same thing. If the SW is interrupted after the early ack but before `handleCommand()` completes or before `wsSend(JSON.stringify(result))` reaches the daemon, the daemon never gets a response and the offscreen queue will not retry, because that frame has already been shifted out.

So this still leaves a command-loss window exactly in the path this PR is trying to harden. I would suggest either:
- delaying the ack until the result has been sent back successfully, or
- introducing a separate completion ack / correlation mechanism instead of dropping the frame on initial receipt.

There is a second loss path on the return leg as well. In `background.ts`, `wsSend()` treats offscreen delivery as best-effort: if `chrome.runtime.sendMessage({ type: 'ws-send', ... })` comes back with `{ ok: false }` because the offscreen WebSocket is not open yet, it just triggers `offscreenConnect()` and returns without buffering or retrying that payload. So even when `handleCommand()` itself succeeds, the result frame can still be dropped if the daemon connection blips between command receipt and response send.

That means the current design can lose commands in both directions: once on the early ack path, and again on the result-send path. I think the response payload needs the same durability treatment as incoming frames, not just a reconnect side effect.

lyfuci:
@Astro-Han Thanks for the thorough review. All four reliability issues are now addressed in 72da27d:

**1. Forward command loss (ws-message ack-before-completion)**
The `ws-message` handler now uses async response (`return true`) and only calls `sendResponse({ok: true})` after `handleCommand()` completes and `wsSend()` delivers (or buffers) the result. If the SW is suspended mid-execution, the frame stays in offscreen's `pendingFrames` and will be retried.

**2. Reverse result loss (wsSend fire-and-forget)**
`wsSend()` now buffers failed payloads into a `pendingResults[]` queue (capped at 100, oldest-first eviction). Buffered results are flushed automatically when the transport reconnects — both offscreen and legacy paths. Added `wsSendDirect()` as a non-buffering primitive to avoid recursion during flush.

**3. pendingFrames unbounded growth**
Added `MAX_PENDING_FRAMES = 100` with oldest-first eviction in `offscreen.ts`. Logs a warning when evicting.

**4. forceLegacyTransport permanent lockout**
Reset to `false` on every keepalive alarm (~15s), so a transient offscreen creation failure doesn't permanently disable the preferred transport.

TypeScript compiles cleanly. Legacy fallback path is preserved — all changes are additive.

GreyC:
Related smaller-scope issue: #1537. It does not replace the offscreen persistent WebSocket design here; it only guards the current service-worker WebSocket path against stale close events clobbering a newer active connection.

jackwener:
**Status: Approach attempted and reverted on main — closing**

The "move daemon WebSocket into an offscreen document" approach you proposed here was implemented in a later PR (#1280) and merged briefly as commit `ca25f65b`. It was then reverted in commit `1da105ed` (`revert: offscreen daemon bridge`) because of complications that surfaced after merge.

The mainline solution went a different route — keep the daemon connection in the service worker plus the keepalive/recovery logic improvements in PRs around #1554 (`fix(extension): coalesce daemon websocket connects`), #1540 (`fix(extension): keep active daemon websocket`), and #1645 (`fix(browser): goto retry stale page identity`).

Closing as superseded. If MV3 SW lifecycle still causes intermittent connection issues for you on current main + extension v1.0.18+, please open a fresh issue with the actual symptom (the offscreen approach itself is no longer the direction).

PR #583, opened 2026-03-29
feat(github): add GitHub adapter with 5 public API commands

kagura-agent:
## What

Adds a native GitHub adapter with 5 commands using GitHub's public REST API. No authentication required, no browser needed.

## Commands

| Command | Description | Example |
|---------|-------------|---------|
| `github trending` | Trending repos (recent stars) | `opencli github trending --language python --limit 5` |
| `github search` | Search repositories | `opencli github search --query "agent memory" --limit 10` |
| `github repo` | Repository details | `opencli github repo --owner jackwener --name opencli` |
| `github user` | User profile | `opencli github user --username jackwener` |
| `github issues` | List repository issues | `opencli github issues --owner jackwener --name opencli --state open` |

## Why

GitHub is the most used developer platform and a natural fit for an AI-agent CLI tool. Surprisingly it wasn't covered yet. All commands are pure YAML declarative adapters — zero TypeScript, minimal maintenance.

## Implementation

- 5 YAML files in `src/clis/github/`
- Uses GitHub REST API v3 (`api.github.com`)
- `strategy: public`, `browser: false`
- Trending uses GitHub search API (sorted by stars, filtered by creation date) since the unofficial trending APIs are unreliable

## Testing

All 5 commands manually verified:

```
$ opencli github search --query openclaw --limit 3
┌──────┬───────────────────────────────────┬────────┬───────┬────────────┬───────────────────────┐
│ Rank │ Name                              │ Stars  │ Forks │ Language   │ Description           │
├──────┼───────────────────────────────────┼────────┼───────┼────────────┼───────────────────────┤
│ 1    │ openclaw/openclaw                 │ 340056 │ 67032 │ TypeScript │ Your own personal AI… │
└──────┴───────────────────────────────────┴────────┴───────┴────────────┴───────────────────────┘

$ opencli github trending --limit 3
┌──────┬───────────────────────┬───────┬───────┬────────────┬─────────────────────────┐
│ Rank │ Name                  │ Stars │ Forks │ Language   │ Description             │
├──────┼───────────────────────┼───────┼───────┼────────────┼─────────────────────────┤
│ 1    │ slavingia/skills      │ 5199  │ 368   │ -          │ Claude Code skills…     │
│ 2    │ larksuite/cli         │ 2622  │ 116   │ Go         │ A command-line tool…    │
│ 3    │ magnum6actual/flipoff │ 2308  │ 280   │ JavaScript │ Free split-flap…        │
└──────┴───────────────────────┴───────┴───────┴────────────┴─────────────────────────┘
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@jackwener/opencli",
-  "version": "1.5.4",
+  "version": "1.5.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@jackwener/opencli",
-      "version": "1.5.4",
+      "version": "1.5.5",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
--- a/src/clis/github/issues.yaml
+++ b/src/clis/github/issues.yaml
@@ -1,0 +1,45 @@
+site: github
+name: issues
+description: List issues for a GitHub repository
+domain: github.com
+strategy: public
+browser: false
+
+args:
+  owner:
+    type: string
+    required: true
+    description: Repository owner
+  name:
+    type: string
+    required: true
+    description: Repository name
+  state:
+    type: string
+    default: open
+    description: "Issue state: open, closed, all"
+  limit:
+    type: int
+    default: 10
+    description: Number of issues
+
+pipeline:
+  - fetch:
+      url: "https://api.github.com/repos/${{ args.owner }}/${{ args.name }}/issues?state=${{ args.state }}&per_page=${{ args.limit }}"
+      headers:
+        Accept: application/vnd.github.v3+json
+        User-Agent: opencli
+
+  - filter: "!item.pull_request"
+
+  - map:
+      number: "#${{ item.number }}"
+      title: "${{ (item.title || '').substring(0, 60) }}"
+      author: "${{ item.user.login }}"
+      labels: "${{ (item.labels || []).map(l => l.name).join(', ') }}"
+      comments: ${{ item.comments }}
+      created: "${{ item.created_at?.substring(0, 10) }}"
+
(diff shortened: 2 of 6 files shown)


Astro-Han (commented):
Thanks for the contribution! One thing worth noting: `gh` (GitHub CLI) is already registered in opencli's external CLI registry (`src/external-clis.yaml`), so `opencli gh ...` already covers GitHub operations with full auth support. Adding a separate `github` adapter would create two overlapping entry points for the same platform.

If there's a specific use case that `gh` doesn't cover well (e.g. the trending query), that might work better as a standalone plugin rather than a built-in adapter.

Astro-Han:
Another concrete issue: `src/clis/github/trending.yaml` appends `args.language` directly into the GitHub Search API query string instead of URL-encoding it. That works for simple values like `rust`, but common inputs such as `c++`, `c#`, or `objective-c` will change the query semantics once they hit `?q=...`.

So this command can silently return the wrong repositories for exactly the languages people are likely to filter by. I think the language fragment needs the same encoding treatment as `search.yaml` gives `args.query`.

Also, `src/clis/github/issues.yaml` currently fetches `/repos/{owner}/{repo}/issues?per_page={limit}` and only filters out pull requests afterwards. GitHub's issues endpoint mixes issues and PRs in the same page window, so PR rows still consume that first-page quota even if the adapter drops them later.

The result is that `opencli github issues --limit 10` can easily return fewer than 10 actual issues on repos where recent PRs are active, even when plenty of issues exist on later pages. If this adapter wants real issue counts, it needs either a search endpoint that excludes PRs up front or pagination that keeps pulling until it has `limit` true issues.

jackwener:
感觉没有什么意义，gh覆盖了基本上大部分场景，不过还是感谢你的贡献

PR #585, opened 2026-03-29
fix(cli): normalize boolean arg aliases in commander adapter

inori-3333:
## Summary

This PR fixes boolean argument normalization in the Commander adapter by treating `type: 'boolean'` as an alias of `type: 'bool'`.

## Why

The registry currently uses both `bool` and `boolean` in command definitions.

However, `normalizeArgValue()` only recognized `bool`, so commands using `type: 'boolean'` could receive string values like `"false"` instead of the real boolean `false`.

Because `"false"` is truthy in JavaScript, this could flip command behavior in subtle but real ways.

## Impact

This affects commands that define boolean flags with `type: 'boolean'`, especially when the default is `false` or when the user explicitly passes `--flag false`.

Examples in the current tree include:

- `reddit save --undo`
- `boss mark --remove`
- `twitter reply-dm --skip-replied`

## Changes

- update `normalizeArgValue()` to accept both `bool` and `boolean`
- add regression tests for `type: 'boolean'`
- verify both default `false` and explicit `--undo false` are normalized to real booleans

## Verification

- `npx vitest run src/commanderAdapter.test.ts`
- `npm run typecheck`

Both passed locally.

## Scope

This PR intentionally keeps the fix minimal and does not add new metadata validation rules for argument types.

### DIFF
--- a/src/commanderAdapter.test.ts
+++ b/src/commanderAdapter.test.ts
@@ -76,3 +76,50 @@ describe('commanderAdapter arg passing', () => {
     expect(mockExecuteCommand).not.toHaveBeenCalled();
   });
 });
+
+describe('commanderAdapter boolean alias support', () => {
+  const cmd: CliCommand = {
+    site: 'reddit',
+    name: 'save',
+    description: 'Save a post',
+    browser: false,
+    args: [
+      { name: 'post-id', positional: true, required: true, help: 'Post ID' },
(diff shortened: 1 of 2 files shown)


jackwener:
Thanks for the fix. I reviewed it locally, confirmed the boolean alias regression is covered by tests, and merged it after typecheck/build passed.

PR #603, opened 2026-03-30
Security/cookie auth hardening

zvrr:
## Description

Related issue:

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

- [ ] I ran the checks relevant to this PR
- [ ] I updated tests or docs if needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,47 @@
 # Changelog
 
+## [Unreleased] — security/cookie-auth-hardening (2026-03-30)
+
+### Security
+
+#### 🔐 Daemon Bearer-Token 认证（防本地进程横向移动）
+
+**风险**：Daemon HTTP 端口 `19825` 仅凭 `X-OpenCLI` 自定义头部防御 CSRF。任何本地进程只需加上该头部即可控制浏览器自动化会话，执行任意 JS、读取 Cookie 等高权限操作。
+
+**修复**（`src/daemon.ts`、`src/browser/daemon-client.ts`）：
+- Daemon 启动时用 `crypto.randomBytes(32)` 生成 64 字符十六进制 Token，写入 `~/.opencli/daemon.token`（权限 `0o600`）。
+- 所有非 `/ping` HTTP 端点要求 `Authorization: Bearer <token>` 头部；验证采用恒时比较防时序攻击。
+- `daemon-client.ts` 懒加载并缓存 token；收到 `401` 时自动刷新缓存以应对 Daemon 重启。
+- Daemon 正常退出时删除 token 文件，避免旧 token 被复用。
+
+#### 🔐 Extension ID 固定（可选加固）
+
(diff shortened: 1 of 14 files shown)


Astro-Han (commented):
Security hardening looks well thought out — token lifecycle (generate → 0o600 write → constant-time verify → cleanup on shutdown), 401 retry in daemon-client, and CDP localhost guard all make sense. Test coverage is good too.

Two observations:

1. The xiaohongshu blob-URL video download fix is a separate concern from the security changes. Might be cleaner as its own PR so each can be reviewed/reverted independently.

2. `isSensitiveCookieName` uses substring matching (`lower.includes(s)`), which casts a wide net — e.g. a cookie named `authenticate_preference` would match on `auth`. For security redaction that's probably the right tradeoff (better to over-redact), but a brief comment noting the intentional broad matching would help future readers.

Astro-Han:
Another concrete risk here: `CDPBridge.connect()` now creates and attaches a target for browser-level CDP endpoints, but that whole block still sits inside the broad `try/catch` in `src/browser/cdp.ts`. If `Target.createTarget` or `Target.attachToTarget` fails, the exception is swallowed and `connect()` still resolves a `CDPPage`.

That means callers can get a seemingly successful connection even though no target session was established, and the first real browser command fails later with a much less obvious error. I think target attach failure should still reject `connect()`, with only the post-attach niceties like `Page.enable` staying best-effort.

Another issue in the auth hardening path: `src/daemon.ts` treats token-file persistence failure as non-fatal (`could not write token file`), but the daemon still requires `Authorization: Bearer <token>` on every non-`/ping` HTTP request. On the client side, `src/browser/daemon-client.ts` can only discover that token by reading `~/.opencli/daemon.token`; if the file is missing or unreadable it simply omits the header.

So a daemon that starts with an in-memory token but fails to persist the file becomes unusable to the CLI: every `/status` and `/command` request will get 401, with no real fallback path. I think token-file write failure either needs to be fatal at startup, or the daemon needs an actual compatibility mode instead of logging "non-fatal" and proceeding.

jackwener:
感谢提交，但这条不合并。

PR description 完全空白，没有威胁模型，没有具体动机说明，没有 issue 链接。"Security/cookie auth hardening" 标题下涉及 `src/browser/cdp.ts` / `daemon-client.ts` / `daemon.ts` / `xiaohongshu/download.ts` 等多个面 +623/-46 的改动，但读者没有办法判断每一处变更对应哪个具体威胁、是不是真的有必要、有没有引入新的 attack surface。

安全相关的改动尤其需要先讲清楚 "防什么 / 现状哪里有问题 / 这个修复关掉了什么"，再讨论实现。建议拆成一个 issue 先把威胁模型写清楚，针对单一具体问题再提 PR。

关闭。

PR #608, opened 2026-03-30
fix(xiaohongshu): check login wall before autoScroll in search (fixes #597)

### LINKED ISSUE #597: [Bug] xiaohongshu search crashes on login-walled results page before surfacing AuthRequiredError
## Summary
`opencli xiaohongshu search` crashes with `TypeError: Cannot read properties of null (reading 'scrollHeight')` on the Xiaohongshu search results page, instead of surfacing the intended `AuthRequiredError` / login-wall message.

## Environment
- opencli: `1.5.6`
- extension: `1.5.5`
- Node.js: `v22.15.0`
- OS: macOS (user env timezone `Asia/Shanghai`)
- Browser bridge: connected

`opencli doctor` output:
```text
opencli v1.5.6 doctor (node v22.15.0)

[OK] Daemon: running on port 19825
[OK] Extension: connected (v1.5.5)
[OK] Connectivity: connected in 0.3s

Everything looks good!
```

## Repro
```bash
opencli xiaohongshu search '中海阳光玫瑰园' --limit 5 -f json --verbose
```

## Actual result
```text
[opencli] Starting daemon (js)...
Error: TypeError: Cannot read properties of null (reading 'scrollHeight')
    at <anonymous>:4:42
    at <anonymous>:19:7
    at sendCommand (file:///Users/haoyuebai/.nvm/versions/node/v22.15.0/lib/node_modules/@jackwener/opencli/dist/browser/daemon-client.js:87:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Page.autoScroll (file:///Users/haoyuebai/.nvm/versions/node/v22.15.0/lib/node_modules/@jackwener/opencli/dist/browser/page.js:294:9)
    at async Object.func (file:///Users/haoyuebai/.nvm/versions/node/v22.15.0/lib/node_modules/@jackwener/opencli/dist/clis/xiaohongshu/search.js:44:9)
```

In another run, I also hit:
```text
💥 Unexpected error: {"code":-32000,"message":"Inspected target navigated or closed"}
```

## Expected result
If the page is login-walled or otherwise unavailable for DOM extraction, the command should fail gracefully with a clear auth/login-wall error, or skip scrolling and return an empty result with a useful explanation.

## Analysis
The current implementation in `dist/clis/xiaohongshu/search.js` does this:
```js
await page.goto(`https://www.xiaohongshu.com/search_result?keyword=${keyword}&source=web_search_result_notes`);
await page.wait(3);
await page.autoScroll({ times: 2 });
const payload = await page.evaluate(`... const loginWall = /登录后查看搜索结果/.test(document.body.innerText || '') ...`);
```

So `autoScroll()` runs **before** checking `loginWall`.

At least in my environment, the search page lands on a login-walled state where DOM scrolling is not stable. A direct `page.evaluate(...)` check showed:
```json
{
  "count": 0,
  "loginWall": true,
  "results": [],
  "title": "中海阳光玫瑰园 - 小红书搜索"
}
```

The scroll helper in `dist/browser/dom-helpers.js` reads `document.body.scrollHeight` immediately:
```js
const lastHeight = document.body.scrollHeight;
window.scrollTo(0, lastHeight);
```
The stack suggests `document.body` can be absent / replaced during page transition or login-wall rendering, which makes the command crash before the auth-path can run.

## Suggested fix
One of these would likely make the command robust:
1. Check `loginWall` (or at least `document.body`) before calling `page.autoScroll()`.
2. Make `autoScrollJs()` defensive against missing `document.body`.
3. Skip scrolling entirely when the initial page already contains no `section.note-item` results and matches the login-wall text.
4. Optionally retry once when the page is in a transient navigation state that currently produces `Inspected target navigated or closed`.

Happy to test a patch if useful.

kagura-agent:
## Summary

Fixes #597. The `search` command crashes on Xiaohongshu when a login wall is presented because `autoScroll()` runs before the login-wall check, and `document.body` can be null in that state.

## Changes

1. **`search.ts`** — Added early login-wall detection (`page.evaluate` with regex for `登录后查看搜索结果`) **before** `autoScroll()`. Throws `AuthRequiredError` immediately if detected.

2. **`dom-helpers.ts`** — Added `if (!document.body) return;` guard in `autoScrollJs` as defensive fallback for all adapters.

3. **`search.test.ts`** — Updated login-wall test to verify `autoScroll` is never called when wall is detected.

4. **`dom-helpers.test.ts`** — Added test for null `document.body` handling.

## Testing

All 397 tests pass. `tsc --noEmit` clean.

### DIFF
--- a/src/browser/dom-helpers.test.ts
+++ b/src/browser/dom-helpers.test.ts
@@ -1,5 +1,19 @@
 import { describe, it, expect } from 'vitest';
-import { waitForCaptureJs, waitForSelectorJs } from './dom-helpers.js';
+import { autoScrollJs, waitForCaptureJs, waitForSelectorJs } from './dom-helpers.js';
+
+describe('autoScrollJs', () => {
+  it('returns early without error when document.body is null', async () => {
+    const g = globalThis as any;
+    const origDoc = g.document;
+    g.document = { body: null, documentElement: {} };
+    g.window = g;
+    const code = autoScrollJs(3, 500);
+    // Should resolve without throwing
+    await expect(eval(code)).resolves.not.toThrow();
+    g.document = origDoc;
+    delete g.window;
+  });
+});
 
 describe('waitForCaptureJs', () => {
   it('returns a non-empty string', () => {
--- a/src/browser/dom-helpers.ts
+++ b/src/browser/dom-helpers.ts
@@ -109,6 +109,7 @@ export function scrollJs(direction: string, amount: number): string {
 export function autoScrollJs(times: number, delayMs: number): string {
   return `
     (async () => {
+      if (!document.body) return;
       for (let i = 0; i < ${times}; i++) {
         const lastHeight = document.body.scrollHeight;
         window.scrollTo(0, lastHeight);
--- a/src/clis/xiaohongshu/search.test.ts
+++ b/src/clis/xiaohongshu/search.test.ts
@@ -41,15 +41,16 @@ describe('xiaohongshu search', () => {
     expect(cmd?.func).toBeTypeOf('function');
 
     const page = createPageMock([
-      {
-        loginWall: true,
-        results: [],
-      },
+      // First evaluate: early login-wall check (returns true)
+      true,
     ]);
 
     await expect(cmd!.func!(page, { query: '特斯拉', limit: 5 })).rejects.toThrow(
       'Xiaohongshu search results are blocked behind a login wall'
     );
+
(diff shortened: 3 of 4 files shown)


jackwener:
Merged after review. The fix direction is correct: detect the Xiaohongshu login wall before autoScroll, and keep autoScrollJs resilient when document.body is absent. I also validated with targeted Vitest, typecheck, and build before merge. Thanks for the contribution.

PR #615, opened 2026-03-30
fix(xiaohongshu): support full URL/short link and fix video extraction

geegewu:
## Summary

- **URL handling**: accept full URLs (with `xsec_token`) and short links (`xhslink.com`) in addition to bare note IDs. Bare note IDs construct `explore/{noteId}` which now lacks the `xsec_token` required by Xiaohongshu, causing all downloads to fail with "No media found".
- **Video extraction**: extract real video URLs from `window.__INITIAL_STATE__` and inline script JSON instead of relying on DOM `<video>` elements which return `blob:` URLs that cannot be downloaded via HTTP.

## Reproduction

```bash
# Before: fails — page loads but xsec_token missing, no content rendered
opencli xiaohongshu download 69bc166f000000001a02069a

# After: works with short link
opencli xiaohongshu download "http://xhslink.com/o/4MKEjsZnhCz"

# After: works with full URL
opencli xiaohongshu download "https://www.xiaohongshu.com/explore/69bc166f000000001a02069a?xsec_token=..."
```

## Test result

Successfully downloaded a 21.5 MB MP4 video note via short link.

### DIFF
--- a/src/clis/xiaohongshu/download.test.ts
+++ b/src/clis/xiaohongshu/download.test.ts
@@ -1,0 +1,96 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { IPage } from '../../types.js';
+
+const { mockDownloadMedia, mockFormatCookieHeader } = vi.hoisted(() => ({
+  mockDownloadMedia: vi.fn(),
+  mockFormatCookieHeader: vi.fn(() => 'a=b'),
+}));
+
+vi.mock('../../download/media-download.js', () => ({
+  downloadMedia: mockDownloadMedia,
+}));
+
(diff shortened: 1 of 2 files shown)


geegewu:
💡 Note: For users who just need to download XHS videos, [yt-dlp](https://github.com/yt-dlp/yt-dlp) also supports Xiaohongshu out of the box:

```bash
yt-dlp "https://www.xiaohongshu.com/explore/69bc166f000000001a02069a?xsec_token=..."
```

This PR is still valuable for the opencli ecosystem since it fixes the built-in download command to work with the current XHS auth flow (xsec_token) and resolves the blob: URL extraction issue.

PR #623, opened 2026-03-30
feat(instagram): add media download command

fii6:
## Summary
- add an `instagram dl` command to download media from `/p/`, `/reel/`, and `/tv/` links using the active browser session
- fetch Instagram media metadata inside the logged-in browser context, then download the resolved image/video URLs locally
- default downloads to `/sdcard/Pictures/Instagram`, support `--path` to override the destination
- keep command output minimal: no result table, no progress spam, only `📁 saved: ...` after completion
- place the new command after `saved` in Instagram help output and keep the command description in English

## Validation
- `npm run build`
- `vitest run src/clis/instagram/download.test.ts --pool=threads --maxWorkers=1`
- real download test via CDP-backed browser session:
  - `opencli instagram dl "https://www.instagram.com/p/DWUR_azCWbN/" --path ~/tmp/instagram-dl-alias-test`

### DIFF
--- a/src/build-manifest.ts
+++ b/src/build-manifest.ts
@@ -47,8 +47,6 @@ export interface ManifestEntry {
   modulePath?: string;
   /** Pre-navigation control — see CliCommand.navigateBefore */
   navigateBefore?: boolean | string;
-  /** Lower values appear earlier in command listings/help. */
-  order?: number;
 }
 
 import { type YamlCliDefinition, parseYamlArgs } from './yaml-schema.js';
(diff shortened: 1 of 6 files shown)


jackwener:
Merged after review. I specifically looked at the design intent here: the new Instagram media command should fit the existing OpenCLI command model instead of introducing a one-off UX path. I renamed it to the repository-wide download convention, removed the global ordering change, switched the default output directory to a cross-platform Downloads path, and validated with targeted Vitest, commanderAdapter/build-manifest coverage, typecheck, and build before merge. Thanks for the contribution.

PR #628, opened 2026-03-31
fix(weread): recover book details from cached shelf fallback

Astro-Han:
## Description

Fix WeRead `book` detail fallback when private API auth has expired.

This PR improves the detail fallback chain in two cases:
- reuse the first parsed shelf snapshot instead of reloading `/web/shelf` before search fallback
- allow a safe raw-order fallback when `shelfIndexes` never hydrate but rendered shelf reader links still cover the cached entries

It also adds regression tests for:
- ordinary book fallback through public search
- rejecting same-title results when the visible author does not match
- raw-order fallback for cached entries when shelf indexes remain empty

Related issue:
- #610

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

```bash
$ node dist/main.js weread shelf --limit 200 -f json
⚠  WeRead private API auth expired; showing cached shelf data from localStorage. Results may be stale, and detail commands may still require re-login.

$ node dist/main.js weread book 29196155 -f json
[
  {
    "title": "文明、现代化、价值投资与中国",
    "author": "李录",
    "publisher": "中信出版集团"
  }
]
```

### DIFF
--- a/src/clis/weread/book.ts
+++ b/src/clis/weread/book.ts
@@ -1,7 +1,13 @@
 import { cli, Strategy } from '../../registry.js';
 import { CliError } from '../../errors.js';
 import type { IPage } from '../../types.js';
-import { fetchPrivateApi, resolveShelfReaderUrl } from './utils.js';
+import {
+  fetchPrivateApi,
+  fetchWebApi,
+  resolveShelfReader,
+  WEREAD_UA,
+  WEREAD_WEB_ORIGIN,
+} from './utils.js';
 
 interface ReaderFallbackResult {
   title: string;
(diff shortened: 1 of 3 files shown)


jackwener:
Merged after review. I specifically looked at the design intent here: make weread book detail resilient when the private API path is unavailable, while keeping fallback resolution conservative and safe. I validated with targeted Vitest, typecheck, and build before merge. Thanks for the contribution.

PR #655, opened 2026-04-01
fix(browser): normalize wait timeouts

hiSandog:
## Summary
- normalize \ in both daemon and direct CDP paths so adapters can pass seconds while older millisecond-style callers keep working
- switch \ from a fixed sleep to \ plus a selector wait, matching the newer intercept semantics
- make \ use an explicit 10-second selector timeout and add regression coverage for the new wait helper and 36kr article flow

## Testing
- \
> @jackwener/opencli@1.5.5 test
> vitest run --project unit src/browser/wait.test.ts src/browser/dom-helpers.test.ts src/browser/cdp.test.ts

 RUN  v4.1.1 /Users/sandog/sandog/code/kaiyuan-space/opencli

 Test Files  3 passed (3)
      Tests  13 passed (13)
   Start at  14:46:42
   Duration  349ms (transform 82ms, setup 0ms, import 113ms, tests 259ms, environment 0ms)
- \
> @jackwener/opencli@1.5.5 test:adapter
> vitest run --project adapter src/clis/36kr/article.test.ts src/clis/36kr/hot.test.ts

 RUN  v4.1.1 /Users/sandog/sandog/code/kaiyuan-space/opencli

 Test Files  2 passed (2)
      Tests  4 passed (4)
   Start at  14:46:42
   Duration  79ms (transform 36ms, setup 0ms, import 42ms, tests 10ms, environment 0ms)
- \
> @jackwener/opencli@1.5.5 typecheck
> tsc --noEmit

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "OpenCLI",
-  "version": "1.6.8",
+  "version": "1.6.10",
   "description": "Browser automation bridge for the OpenCLI CLI tool. Executes commands in isolated Chrome windows via a local daemon.",
   "permissions": [
     "debugger",
--- a/extension/package.json
+++ b/extension/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opencli-extension",
-  "version": "1.6.8",
+  "version": "1.6.10",
   "private": true,
   "type": "module",
   "scripts": {


hiSandog (commented):
Good point — I agree keeping a single consistent unit (seconds) is simpler and less error-prone. I reverted the normalizeWaitTimeoutMs helper entirely and kept the original (timeout ?? default) * 1000 pattern. The only functional changes are:

- **36kr/article**: improved wait to use waitForCapture + selector wait instead of a blind sleep
- **sinafinance/rolling-news**: fixed the wrong unit from 10000 (ms) → 10 (s)

Tests pass. Thanks for the review!

PR #666, opened 2026-04-01
feat(twitter): add --images flag to post command

artshooter:
## Summary

- Add `--images` option to `twitter post` command, supporting up to 4 images (comma-separated paths)
- Uses existing `page.setFileInput()` (CDP `DOM.setFileInputFiles`) to attach files to Twitter's file input
- Includes file validation (`statSync` checks), graceful error handling for older extensions, and polling-based upload readiness detection

## Usage

```bash
# Single image
opencli twitter post "Hello world" --images /path/to/image.png

# Multiple images (comma-separated, max 4)
opencli twitter post "Hello world" --images /path/a.png,/path/b.jpg,/path/c.png
```

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npm test` passes (456 passed)
- [x] Manual test: posted tweet with single image successfully
- [x] Manual test: post tweet with multiple images
- [x] Manual test: verify error message when file not found
- [x] Manual test: verify error message when >4 images provided

### DIFF
--- a/src/clis/twitter/post.test.ts
+++ b/src/clis/twitter/post.test.ts
@@ -18,6 +18,10 @@ vi.mock('node:path', async (importOriginal) => {
   return {
     ...actual,
     resolve: vi.fn((p: string) => `/abs/${p}`),
+    extname: vi.fn((p: string) => {
+      const m = p.match(/\.[^.]+$/);
+      return m ? m[0] : '';
+    }),
   };
 });
 
@@ -38,8 +42,8 @@ describe('twitter post command', () => {
     const command = getCommand();
     const page = makePage({
       evaluate: vi.fn()
-        .mockResolvedValueOnce({ ok: true })                              // type text
-        .mockResolvedValueOnce({ ok: true, message: 'Tweet posted successfully.' }), // click post
+        .mockResolvedValueOnce({ ok: true })
+        .mockResolvedValueOnce({ ok: true, message: 'Tweet posted successfully.' }),
     });
 
     const result = await command!.func!(page as any, { text: 'hello world' });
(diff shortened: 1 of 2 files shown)


Astro-Han:
One concrete issue here: the new "wait for image upload" check is using the tweet button enabled state as the upload-complete signal, but `text` is still a required argument and the code pastes the text *before* starting the image upload.

That means `[data-testid="tweetButton"]` can already be enabled just because the tweet has text, even while the image upload is still in progress. In that case the `uploaded` polling loop returns `true` immediately, and the command goes on to click Post before the media is actually attached.

So for a text+image tweet this can silently post the text without the images, or with only a partially finished upload, while still reporting success. I think the wait condition needs to watch an actual media-upload completion signal in the DOM, not the generic button enabled state.

There also does not seem to be any test coverage for the new image path. The diff only changes `src/clis/twitter/post.ts`, with no added or updated tests around `--images`, file validation, upload waiting, or the "post only after media is attached" behavior. Given the change is adding a new side-effecting publish path, I think it should come with at least one focused test for the upload flow.

artshooter:
@Astro-Han  Thanks for the review! Addressed both points:

1. Upload detection fix: Replaced the tweet-button-only polling with a dual-condition check — now waits for   [data-testid="attachments"] to appear with the correct number of [role="group"] elements matching the uploaded image count, AND   the tweet button to be enabled. Also increased the timeout from 20s to 30s. 

2. Test coverage: Added 8 unit tests covering: text-only post, image upload success/timeout, file validation (>4 images,   missing files), missing setFileInput support, and no-browser-session error.

PR #669, opened 2026-04-01
feat(cursor): add chat command with CDP native key events

cjb2667:
## Summary

- Add `opencli cursor chat "prompt"` command that opens a new Cursor chat session and sends a prompt
- Uses CDP native `Input.dispatchKeyEvent` instead of JS `dispatchEvent` to submit messages, fixing compatibility with Cursor v2.6.22+ Lexical editor
- Supports `--timeout` parameter (default 60s) to control how long to wait for AI response

## Motivation

The existing `send` and `ask` commands use JavaScript `dispatchEvent(new KeyboardEvent(...))` to simulate Enter key press. However, Cursor's Lexical editor (since v2.6+) does not respond to synthetic keyboard events — only CDP-level native input events trigger the submit action.

This new `chat` command:
1. Opens a fresh chat via `Cmd+N` (the correct shortcut for new chat, not `Cmd+L`)
2. Injects text via `document.execCommand('insertText')`
3. Submits via CDP `Input.dispatchKeyEvent` (keyDown + keyUp for Enter)
4. Polls for AI response from `[data-message-role="ai"]` elements

## Test plan

- [x] `opencli cursor chat "请回复 hello world"` → AI responds `Hello world!`
- [x] `opencli cursor chat "请回复数字 42"` → AI responds `42`
- [x] Timeout parameter works: `--timeout 120`
- [x] New chat session is created each time (message count resets to 0)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/clis/cursor/chat.ts
+++ b/src/clis/cursor/chat.ts
@@ -1,0 +1,135 @@
+import { cli, Strategy } from '../../registry.js';
+import { SelectorError } from '../../errors.js';
+import type { IPage } from '../../types.js';
+
+export const chatCommand = cli({
+  site: 'cursor',
+  name: 'chat',
+  description: 'Open a new Cursor chat and send a prompt via CDP native key events',
+  domain: 'localhost',
+  strategy: Strategy.UI,
+  browser: true,
+  args: [
+    { name: 'text', required: true, positional: true, help: 'Prompt to send' },
+    { name: 'timeout', required: false, help: 'Max seconds to wait for response (default: 60)', default: '60' },
+  ],
+  columns: ['Role', 'Text'],
+  func: async (page: IPage, kwargs: any) => {
+    const text = kwargs.text as string;
+    const timeout = parseInt(kwargs.timeout as string, 10) || 60;
+    const isMac = process.platform === 'darwin';
+
+    // Cmd+N 打开新对话
+    await cdpKeyCombo(page, isMac ? 'Meta' : 'Control', 'n');
+    await page.wait(2);
+
+    // 查找输入框并注入文本
+    const injected = await page.evaluate(
+      `(function(text) {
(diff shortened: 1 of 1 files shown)


Astro-Han (commented):
A few concerns:

1. **Why not fix the existing commands?** The PR says `send`/`ask` broke because Cursor's Lexical editor ignores synthetic JS keyboard events. The CDP `Input.dispatchKeyEvent` fix here is the right technique, but it should be applied to the existing `send`/`ask` commands rather than adding a parallel `chat` command that does the same thing differently. Users shouldn't need to know which command works with which Cursor version.

2. **`(page as any).bridge` bypasses the IPage abstraction.** `cdpPressEnter` and `cdpKeyCombo` reach into the CDPBridge internal instance directly. This only works in direct CDP mode — in daemon mode (the default), `bridge` doesn't exist and this will throw at runtime with no useful error message. If `Input.dispatchKeyEvent` is needed, it should go through the IPage interface so both transport paths can support it.

3. **No tests.** Even a basic unit test mocking the page object and verifying the call sequence (new chat → inject text → press Enter → poll for response) would catch regressions.

jackwener:
基于当前 main(43d07222) 重新核对后关闭这条。\n\n当前 Cursor adapter 已经有 `clis/cursor/new.js`、`send.js`、`ask.js`、`read.js`、`history.js` 等命令；`page.pressKey()` 在 BrowserBridge/CDP 路径会走 CDP `Input.dispatchKeyEvent`，已经覆盖这条 PR 里为 Lexical editor 引入 native Enter 的核心问题。\n\n这条 PR 还新增在旧路径 `src/clis/cursor/chat.ts`，当前 adapter 布局已迁到 package-root `clis/cursor/*.js`。继续合入会产生重复命令和旧目录结构。按 stale/obsolete PR 关闭。

PR #688, opened 2026-04-02
fix(twitter): resolve article ID to tweet ID before GraphQL query

gucasbrg:
## Summary

- `twitter article` command fails with "Article not found" for all `x.com/i/article/{id}` URLs
- Root cause: article IDs are different from tweet IDs, but the code uses the article ID directly in `TweetResultByRestId` GraphQL query

## Fix

Navigate to the article page first, extract the associated tweet ID from DOM links (`a[href*="/status/"]` or `og:url` meta tag), then use that resolved tweet ID for the GraphQL query.

## Before
```
$ opencli twitter article https://x.com/i/article/2038726133869625344
💥 Article not found
```

## After
```
$ opencli twitter article https://x.com/i/article/2038726133869625344
| title | author | content |
| My Own AI Hedge Fund... | Alexo_0x | # Why Polymarket is... |
```

## Test plan
- [x] Tested with `x.com/i/article/2038726133869625344` — full article content returned
- [x] Existing tweet-based articles (status URLs) still work via the existing `urlMatch` extraction

### DIFF
--- a/src/clis/twitter/article.ts
+++ b/src/clis/twitter/article.ts
@@ -16,32 +16,44 @@ cli({
   ],
   columns: ['title', 'author', 'content', 'url'],
   func: async (page, kwargs) => {
-    // Extract tweet ID from URL if needed
+    // Extract tweet ID from URL if needed.
+    // Article URLs (x.com/i/article/{articleId}) use a different ID than
+    // tweet status URLs — the GraphQL endpoint needs the parent tweet ID.
     let tweetId = kwargs['tweet-id'];
+    const isArticleUrl = /\/article\/\d+/.test(tweetId);
     const urlMatch = tweetId.match(/\/(?:status|article)\/(\d+)/);
     if (urlMatch) tweetId = urlMatch[1];
 
-    // Navigate to article page first, then resolve the associated tweet ID.
-    // Article IDs differ from tweet IDs — we need to visit the article page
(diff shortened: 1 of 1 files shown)


Astro-Han (commented):
Good catch on the article ID vs tweet ID distinction. The fix works for article URLs, but the navigation change from `/i/status/` to `/i/article/` is unconditional, which breaks the other two input paths:

- **Status URL** (`x.com/user/status/123456`): extracts `123456` (a tweet ID), then navigates to `/i/article/123456` which doesn't exist as an article. The DOM extraction could pick up an unrelated `/status/` link from the error page (sidebar, trending, etc.) and silently return the wrong article.
- **Bare tweet ID** (`123456`): same issue.

The article-page-then-resolve approach should only apply when the input is actually an article URL. Something like:

```typescript
const isArticleInput = /\/article\//.test(kwargs['tweet-id']);
if (isArticleInput) {
  await page.goto(`https://x.com/i/article/${tweetId}`);
  await page.wait(3);
  const resolvedTweetId = await page.evaluate(/* ... */);
  if (resolvedTweetId) tweetId = resolvedTweetId;
} else {
  await page.goto(`https://x.com/i/status/${tweetId}`);
  await page.wait(3);
}
```

This way article URLs get the new resolution logic while status URLs and bare IDs keep working as before.

PR #691, opened 2026-04-02
fix(twitter): resolve article ID to tweet ID for article URLs

jackwener:
## Summary

- Article URLs (`x.com/i/article/{articleId}`) use a different ID than tweet status URLs
- The GraphQL `TweetResultByRestId` endpoint requires the parent tweet ID, not the article ID
- **Supersedes PR #688** — fixes the critical bug where all inputs (including status URLs) were routed through the article page

## Fix

- Detect article URLs upfront via regex before extracting the numeric ID
- For article URLs: navigate to article page, extract parent tweet ID from DOM links (`a[href*="/status/"]`) or `og:url` meta tag
- For status URLs or bare IDs: use directly as before (no behavior change)
- Throw explicit error if article→tweet resolution fails (instead of silently falling through)

## Test plan
- [ ] `opencli twitter article https://x.com/i/article/2038726133869625344` — should resolve to tweet ID and return article
- [ ] `opencli twitter article https://x.com/user/status/123456` — should work as before (no article resolution)
- [ ] `opencli twitter article 123456` — bare ID should work as before

### DIFF
--- a/src/clis/twitter/article.ts
+++ b/src/clis/twitter/article.ts
@@ -16,10 +16,40 @@ cli({
   ],
   columns: ['title', 'author', 'content', 'url'],
   func: async (page, kwargs) => {
-    // Extract tweet ID from URL if needed
+    // Extract tweet ID from URL if needed.
+    // Article URLs (x.com/i/article/{articleId}) use a different ID than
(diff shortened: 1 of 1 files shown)


jackwener:
Fixed directly on PR #688's branch instead.

PR #695, opened 2026-04-02
fix(twitter): use search input for SPA navigation instead of pushState

### LINKED ISSUE #690: [Bug]: SPA navigation to /search failed. Final path: (empty). Twitter may have changed its routing.   Add --verbose for details, or report: https://github.com/jackwener/opencli/issues
### Description

SPA navigation to /search failed. Final path: (empty). Twitter may have changed its routing.
  Add --verbose for details, or report: https://github.com/jackwener/opencli/issues

### Steps to Reproduce

SPA navigation to /search failed. Final path: (empty). Twitter may have changed its routing.
  Add --verbose for details, or report: https://github.com/jackwener/opencli/issues

### Expected Behavior

SPA navigation to /search failed. Final path: (empty). Twitter may have changed its routing.
  Add --verbose for details, or report: https://github.com/jackwener/opencli/issues

### OpenCLI Version

1.5.8

### Node.js Version

20.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

0xsline:
## Summary

- The `pushState + popstate` approach works in most environments but fails intermittently for some users (#690)
- Added a **fallback** strategy: when pushState fails after 2 retries, type query into the search input on `/explore` and press Enter
- Both strategies use **selector-based waiting** (`[data-testid="primaryColumn"]`) with graceful fallthrough on timeout
- The pushState approach remains the **primary** strategy

## Root Cause

The failure is **intermittent**, not permanent. Likely causes include Twitter A/B tests, timing race conditions (pathname not updated when checked), or browser/session state differences. In my testing, pushState failed consistently at first, then succeeded consistently later — consistent with server-side variation.

## Changes

- Wrap the `primaryColumn` selector wait in a try/catch so a timeout doesn't throw immediately — it falls through to the path check
- After 2 failed pushState attempts, fall back to the search input approach (nativeSetter + Enter keydown)
- The fallback also uses selector-based waiting, not fixed delays
- For `live` filter with the fallback path, click the "Latest" tab after navigation

## Testing

- Original pushState: failed 2/2, then later passed 3/3 (confirms intermittent nature)
- Search input fallback: passed 2/2 when pushState was failing
- Combined (this PR): covers both scenarios

Fixes #690

### DIFF
--- a/src/clis/twitter/search.test.ts
+++ b/src/clis/twitter/search.test.ts
@@ -153,6 +153,88 @@ describe('twitter search command', () => {
     expect(pushStateCall).toContain('f=top');
   });
 
+  it('falls back to search input when pushState fails twice', async () => {
+    const command = getRegistry().get('twitter/search');
+    expect(command?.func).toBeTypeOf('function');
+
+    const evaluate = vi.fn()
+      .mockResolvedValueOnce(undefined)     // pushState attempt 1
+      .mockResolvedValueOnce('/explore')    // pathname check 1 — not /search
+      .mockResolvedValueOnce(undefined)     // pushState attempt 2
+      .mockResolvedValueOnce('/explore')    // pathname check 2 — still not /search
+      .mockResolvedValueOnce({ ok: true })  // search input fallback succeeds
+      .mockResolvedValueOnce('/search');    // pathname check after fallback
+
+    const page = {
+      goto: vi.fn().mockResolvedValue(undefined),
+      wait: vi.fn().mockResolvedValue(undefined),
+      installInterceptor: vi.fn().mockResolvedValue(undefined),
+      evaluate,
+      autoScroll: vi.fn().mockResolvedValue(undefined),
+      getInterceptedRequests: vi.fn().mockResolvedValue([
+        {
+          data: {
+            search_by_raw_query: {
+              search_timeline: {
+                timeline: {
+                  instructions: [
+                    {
+                      type: 'TimelineAddEntries',
+                      entries: [
+                        {
+                          entryId: 'tweet-99',
+                          content: {
+                            itemContent: {
+                              tweet_results: {
+                                result: {
+                                  rest_id: '99',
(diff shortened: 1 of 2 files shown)


0xsline:
Thanks @Astro-Han for the thorough review, and @jackwener for testing. You were both right — I've updated the PR based on your feedback.

**What changed in v2:**

1. **Root cause corrected:** The failure is intermittent, not permanent. In my testing, pushState failed consistently at first (2/2), then succeeded consistently later (3/3) — consistent with Twitter A/B testing or timing races, not a complete router change.

2. **pushState remains primary:** The `pushState + popstate` approach is kept as the primary strategy with 2 retries, exactly as before.

3. **Selector-based wait restored:** Both the primary and fallback paths now use `wait({ selector: '[data-testid="primaryColumn"]' })` instead of fixed delays. The selector wait is wrapped in try/catch so a timeout falls through gracefully instead of throwing.

4. **Search input as fallback only:** The search input approach is only attempted after pushState fails twice. Regarding commit #105 — the difference is that it was used as the *primary* approach before, whereas here it's a last-resort fallback. If pushState works (which it does most of the time), the search input path is never reached.

The net effect: existing behavior is preserved for the majority of users, and the fallback covers the intermittent failures reported in #690.

Astro-Han:
Thanks for the update — v2 is much cleaner. Root cause correction, selector-based wait restored, pushState kept as primary — all good.

One thing I'm still curious about: commit d1986f0 (#105) moved *away* from the search input + Enter approach because the synthetic KeyboardEvent was ignored by React. The code here looks similar. Do you know what changed on Twitter's side that makes this work now? Even a brief note in the comment would help future maintainers understand why the fallback is expected to be reliable.

Not a blocker since the fallback can't break existing behavior, but worth documenting.

jackwener:
@0xsline can you contact me with Wechat?

My email is jakevingoo@gmail.com.

PR #732, opened 2026-04-03
fix(zhihu): make question runtime-compatible

### LINKED ISSUE #604: zhihu question command fails with 'Not logged in' despite valid session
## Bug Description

`opencli zhihu question <id>` always fails with `🔒 Not logged in to www.zhihu.com`, even when the user is logged in to Zhihu in the browser profile where the Browser Bridge extension is installed.

Other Zhihu commands (e.g. `zhihu search`) work correctly with the same login session.

## Environment

- opencli: v1.5.6
- Extension: v1.5.5
- OS: macOS (Darwin 25.4.0)
- Node: v22.22.1

## Steps to Reproduce

1. Install Browser Bridge extension in a Chrome profile that is logged in to zhihu.com
2. Verify `opencli zhihu search "any query"` works (it does)
3. Run `opencli zhihu question 2021881398772981878`
4. Result: `🔒 Not logged in to www.zhihu.com`

## Root Cause

In `src/clis/zhihu/question.ts`, the command calls `page.evaluate()` with `fetch(..., {credentials: 'include'})` directly, without first navigating to the Zhihu domain. Since the browser context is not on `www.zhihu.com`, cookies are not attached to the fetch request, and the Zhihu API returns an auth error.

In contrast, `src/clis/zhihu/search.yaml` includes a `navigate: https://www.zhihu.com` step before the evaluate, which establishes the correct domain context for cookies to be sent.

## Suggested Fix

Add a navigation step in `question.ts` before the `page.evaluate()` call:

```typescript
await page.goto('https://www.zhihu.com');
```

This matches the pattern used in `search.yaml` and should allow `credentials: 'include'` to work correctly.

KehaoC:
## Summary
Fix `zhihu question` so it works in the real browser runtime again.

The command was using function-style `page.evaluate(...)`, but the runtime only supports string-based `evaluate`. This PR switches it to the supported form, navigates to the question page first, and keeps the existing auth / fetch error mapping.

## Validation
- `npx vitest run src/clis/zhihu/question.test.ts`
- `npm run build`
- `node dist/main.js zhihu question 2021881398772981878 --limit 2 -f json`

Closes #604

### DIFF
--- a/src/clis/zhihu/question.test.ts
+++ b/src/clis/zhihu/question.test.ts
@@ -89,7 +89,7 @@ describe('zhihu question', () => {
       cmd!.func!(page, { id: '2021881398772981878', limit: 3 }),
     ).rejects.toMatchObject({
       code: 'FETCH_ERROR',
-      message: 'Zhihu question answers request failed Failed to fetch',
+      message: 'Zhihu question answers request failed: Failed to fetch',
     });
   });
 
--- a/src/clis/zhihu/question.ts
+++ b/src/clis/zhihu/question.ts
@@ -45,10 +45,10 @@ cli({
       if (result?.status === 401 || result?.status === 403) {
         throw new AuthRequiredError('www.zhihu.com', 'Failed to fetch question data from Zhihu');
       }
-      const detail = result?.status > 0 ? `with HTTP ${result.status}` : (result?.error ?? '');
+      const detail = result?.status > 0 ? ` with HTTP ${result.status}` : (result?.error ? `: ${result.error}` : '');
       throw new CliError(
         'FETCH_ERROR',
-        `Zhihu question answers request failed ${detail}`.trim(),
+        `Zhihu question answers request failed${detail}`,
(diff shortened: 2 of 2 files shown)


KehaoC:
截屏2026-04-03 17 10 32

Sorry, here is the image when running.

Astro-Han:
## Minor: error message formatting for fetch exceptions

When `status === 0` and `error` is present, the current message reads:

> `Zhihu question answers request failed Failed to fetch`

Two consecutive "failed" without a separator reads awkwardly. Suggested fix in `question.ts`:

```ts
const detail = result?.status > 0
  ? `with HTTP ${result.status}`
  : (result?.error ? `: ${result.error}` : '');
```

This produces `Zhihu question answers request failed: Failed to fetch`, which is clearer. The corresponding test expectation in `question.test.ts` should also be updated to match.

Otherwise the PR looks solid — runtime fix is correct, input validation and security hardening are well done. 👍

PR #735, opened 2026-04-03
fix(gemini): stabilize ask reply state handling

Astro-Han:
## Description

Stabilize `gemini ask` by replacing the old reply detection path with a conservative reply-state model.

This change tightens three parts of the Gemini adapter:
- confirms message submission before reading a reply
- assigns reply ownership from the submitted user turn instead of stale indices or assistant counts
- keeps transcript parsing as a low-trust fallback instead of the primary path

Related issue:
#390

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [x] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Checks:
- `npm test`
- `npm run typecheck`
- `npm run test:adapter -- src/clis/gemini/utils.test.ts src/clis/gemini/reply-state.test.ts src/clis/gemini/ask.test.ts`

Manual verification:
- `npm run dev -- gemini ask "请只回复：RV1" --timeout 20 -f plain` -> `💬 RV1`
- `npm run dev -- gemini ask "请只回复：RV1" --timeout 20 -f plain` -> `💬 RV1`
- `npm run dev -- gemini ask "请只回复：RV3" --timeout 20 --new true -f plain` -> `💬 RV3`

### DIFF
--- a/src/clis/gemini/utils.ts
+++ b/src/clis/gemini/utils.ts
@@ -148,22 +148,12 @@ export function collapseAdjacentGeminiTurns(turns: GeminiTurn[]): GeminiTurn[] {
   return collapsed;
 }
 
-function areGeminiTurnsEqual(left: GeminiTurn[], right: GeminiTurn[]): boolean {
-  if (left.length !== right.length) return false;
-  return left.every((turn, index) => turn.Role === right[index]?.Role && turn.Text === right[index]?.Text);
-}
-
 function hasGeminiTurnPrefix(before: GeminiTurn[], current: GeminiTurn[]): boolean {
   if (before.length > current.length) return false;
   return before.every((turn, index) => (
     turn.Role === current[index]?.Role
     && turn.Text === current[index]?.Text
   ));
-}
-
-function areGeminiLinesEqual(left: string[], right: string[]): boolean {
-  if (left.length !== right.length) return false;
-  return left.every((line, index) => line === right[index]);
 }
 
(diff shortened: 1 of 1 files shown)


Astro-Han:
A quick follow-up on what this revision improves.

This version does not keep layering more heuristics onto the old path. Instead, it tightens `gemini ask` around a more conservative submission and reply-state model, with three main changes:

1. Confirm submission before waiting for a reply
   After sending the prompt, the command no longer treats the latest visible text as the answer immediately. It first confirms that Gemini has actually accepted the message.

2. Assign reply ownership from the current-round user turn
   Instead of relying on stale indices or assistant counts to guess which round a reply belongs to, it captures the current-round user anchor at submission time and only accepts assistant content that belongs after that anchor.

3. Demote transcript parsing to a true fallback path
   Transcript parsing is now a low-trust fallback instead of the primary source of truth, which greatly reduces cases where welcome text, page chrome, or partial streamed text gets mistaken for the final answer.

This revision also adds state-driven tests and command-level verification for the main edge cases, especially:
- running the same prompt twice in a row
- the `--new true` path
- cases where older history is loaded later and must not be mistaken for the current-round reply

In one sentence: this revision makes `gemini ask` much more reliable at distinguishing between “the message was actually submitted” and “the visible reply really belongs to this round.”

Issue #791, opened 2026-04-05
[Windows] Adapter modules fail to load - incorrect import paths (src/ instead of dist/src/)


ufomaker:
# GitHub Issue: OpenCLI Windows Adapter Path Problem

**仓库:** https://github.com/jackwener/opencli  
**Issue Title:** `[Windows] Adapter modules fail to load - incorrect import paths (src/ instead of dist/src/)`

---

## Bug Description

On Windows, after installing `@jackwener/opencli@latest` via npm, all adapter modules fail to load with "Cannot find module" errors. The adapters are importing from `src/registry.js` and `src/errors.js`, but these files should be in `dist/src/` after compilation.

## Environment

- **OS:** Windows 10/11
- **Node.js:** v22.16.0
- **npm:** latest
- **opencli:** v1.6.3 (installed via `npm install -g @jackwener/opencli@latest`)

## Installation Steps

```powershell
# Clean install
npm uninstall -g @jackwener/opencli
npm cache clean --force
npm install -g @jackwener/opencli@latest
```

## Error Output

When running any opencli command (e.g., `opencli --version` or `opencli list`):

```
⚠  Failed to load module C:\Users\Administrator\.opencli\clis\1688\item.js: Cannot find module 'C:\Users\Administrator\.opencli\src\errors.js' imported from C:\Users\Administrator\.opencli\clis\1688\item.js
⚠  Failed to load module C:\Users\Administrator\.opencli\clis\amazon\bestsellers.js: Cannot find module 'C:\Users\Administrator\.opencli\src\registry.js' imported from C:\Users\Administrator\.opencli\clis\amazon\bestsellers.js
... (hundreds of similar errors for all adapters)
```

## Expected Behavior

The adapters should import from the compiled `dist/src/` directory, not the source `src/` directory.

## Root Cause Analysis

The adapter files in `clis/` directory have import statements like:
```javascript
import { cli } from 'src/registry.js';
```

But after npm install, the compiled files should be in:
```
C:\Users\Administrator\AppData\Roaming\npm\node_modules\@jackwener\opencli\dist\src\registry.js
```

## Possible Solutions

1. **Fix import paths in adapters** - Change imports from `src/` to `dist/src/` for published packages
2. **Use package.json exports** - Define proper exports in package.json to resolve imports correctly
3. **Postinstall script** - Fix import paths during postinstall on Windows

## Additional Context

- The issue appears on Windows only (Unix-like systems may handle relative paths differently)
[...]

Issue #788, opened 2026-04-04
bug: `opencli --version` reports `0.0.0` when installed from source + `Failed to load module` for user clis


AhogeK:
## Description

When installing from source (as documented in the README under **For Developers**), two issues occur simultaneously:

1. `opencli --version` always reports `0.0.0` instead of the actual version.
2. Every command in `~/.opencli/clis/` fails with `Failed to load module ... Cannot find module '/Users/<user>/.opencli/src/registry.js'`.

---

## Environment

- **Install method:** from source (`git clone` → `npm install` → `npm run build` → `npm link` / `pnpm link --global`)
- **Node:** v24.14.1 (via nvm)
- **npm:** latest
- **OS:** macOS
- **opencli version (package.json):** 1.6.3

---

## Steps to Reproduce

```bash
git clone git@github.com:jackwener/opencli.git && cd opencli
npm install
npm run build
npm link   # or: pnpm link --global
opencli --version
opencli doctor
```

---

### Bug 1 — `version` reports `0.0.0`

`src/version.ts` resolves `package.json` with one level up from `__dirname`:

```ts
// src/version.ts
const pkgJsonPath = path.resolve(__dirname, '..', 'package.json');
```

At runtime after `npm run build`, `__dirname` is `dist/src/`. One level up is `dist/`, where **no `package.json` exists**. The `try/catch` silently falls back to `'0.0.0'`.

This works fine for `npm install -g` users because npm automatically includes the root `package.json` in the published package, placing it at `<package-root>/package.json` — exactly one level above `dist/src/`. But when running from a cloned repo, the root `package.json` is two levels above `dist/src/`.

**Fix:**

```diff
- const pkgJsonPath = path.resolve(__dirname, '..', 'package.json');
+ const pkgJsonPath = path.resolve(__dirname, '..', '..', 'package.json');
```

### Bug 2 — `Failed to load module ~/.opencli/clis/*.js`

The `postinstall` / `fetch-adapters.js` script downloads pre-compiled `.js` adapter files into `~/.opencli/clis/`. These files contain hardcoded absolute-style imports:

```
Cannot find module '/Users/<user>/.opencli/src/registry.js'
Cannot find module '/Users/<user>/.opencli/src/errors.js'
```

However, `~/.opencli/src/` is never created — the core runtime lives inside the npm package's `dist/src/`, not under `~/.opencli/`. These pre-compiled clis appear to have been built against an older layout where the source was symlinked or copied into `~/.opencli/src/`.

[...]

Issue #792, opened 2026-04-05
[Bug]: [Windows]Failed to load module C:\Users\windows\.opencli\src


huuhi:
### Description

when npm install -g @jackwener/opencli then opencli doctor  error Failed to load module C:\Users\windows\.opencli\src

### Steps to Reproduce

1. Run `opencli ...`
2. ...
3. See error

### Expected Behavior

C:\Users\windows> opencli --version
ℹ  First run detected — copying adapters (one-time setup)...
[opencli] Installed 1261 adapter files to C:\Users\windows\.opencli\clis
⚠  Failed to load module C:\Users\windows\.opencli\clis\36kr\article.js: Cannot find module 'C:\Users\windows\.opencli\src\registry.js' imported from C:\Users\windows\.opencli\clis\36kr\article.js
⚠  Failed to load module C:\Users\windows\.opencli\clis\1688\item.js: Cannot find module 'C:\Users\windows\.opencli\src\errors.js' imported from C:\Users\windows\.opencli\clis\1688\item.js
⚠  Failed to load module C:\Users\windows\.opencli\clis\1688\search.js: Cannot find module 'C:\Users\windows\.opencli\src\errors.js' imported from C:\Users\windows\.opencli\clis\1688\search.js
⚠  Failed to load module C:\Users\windows\.opencli\clis\36kr\news.js: Cannot find module 'C:\Users\windows\.opencli\src\registry.js' imported from C:\Users\windows\.opencli\clis\36kr\news.js
⚠  Failed to load module C:\Users\windows\.opencli\clis\36kr\search.js: Cannot find module 'C:\Users\windows\.opencli\src\registry.js' imported from C:\Users\windows\.opencli\clis\36kr\search.js
⚠  Failed to load module C:\Users\windows\.opencli\clis\amazon\bestsellers.js: Cannot find module 'C:\Users\windows\.opencli\src\registry.js' imported from C:\Users\windows\.opencli\cli

### OpenCLI Version

1.6.3

### Node.js Version

22.x

### Operating System

Windows

### Logs / Screenshots

```shell

```

Astro-Han:
Thanks for reporting this. This looks closely related to #788, specifically the adapter/shim path mismatch in 1.6.3.

[...]

PR #795, opened 2026-04-05
refactor: migrate adapter imports to package exports

jackwener:
## Summary

Replaces all relative imports across 484 adapter files with Node.js package exports (`@jackwener/opencli/registry`, `@jackwener/opencli/errors`, etc.).

**Before**: adapters used fragile relative paths (`../../src/registry.js`) that broke whenever the internal directory structure changed (caused #788, #791).

**After**: adapters use stable package exports that resolve through Node.js module resolution, completely decoupled from internal layout.

### What changed
- **package.json**: added 14 sub-path exports covering all public modules
- **484 adapter files**: `../../src/registry.js` → `@jackwener/opencli/registry`, etc.
- **discovery.ts**: simplified `ensureUserCliCompatShims` from ~80 lines of shim generation to ~30 lines (just symlink + package.json)
- **registry-api.ts**: exported `CommandArgs` type (was missing)
- **Removed**: root-level shim directories (`browser/`, `download/`, `pipeline/`) and their tsconfig/package.json entries

### User adapter authoring (new pattern)
```typescript
import { cli, Strategy } from '@jackwener/opencli/registry';
import { CliError } from '@jackwener/opencli/errors';
import type { IPage } from '@jackwener/opencli/types';
```

Works in both `~/.opencli/clis/` (via symlink) and the package itself (via self-referencing).

## Test plan
- [x] Build succeeds (497 entries manifest)
- [x] All 496 unit tests pass (42 files)
- [x] Compiled adapters use `@jackwener/opencli/...` imports

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,7 +73,7 @@ See [`hackernews/top.yaml`](clis/hackernews/top.yaml) for a real example.
 Create a file like `clis/<site>/<command>.ts`:
 
 ```typescript
-import { cli, Strategy } from '../../src/registry.js';
+import { cli, Strategy } from '@jackwener/opencli/registry';
 
 cli({
   site: 'mysite',
--- a/docs/developer/contributing.md
+++ b/docs/developer/contributing.md
@@ -76,8 +76,8 @@ See [`hackernews/top.yaml`](https://github.com/jackwener/opencli/blob/main/clis/
 Create a file like `clis/<site>/<command>.ts`:
 
 ```typescript
-import { cli, Strategy } from '../../src/registry.js';
-import { CommandExecutionError, EmptyResultError } from '../../src/errors.js';
+import { cli, Strategy } from '@jackwener/opencli/registry';
+import { CommandExecutionError, EmptyResultError } from '@jackwener/opencli/errors';
 
 cli({
   site: 'mysite',
--- a/docs/developer/ts-adapter.md
+++ b/docs/developer/ts-adapter.md
@@ -5,8 +5,8 @@ Use TypeScript adapters when you need browser-side logic, multi-step flows, DOM 
 ## Basic Structure
 
 ```typescript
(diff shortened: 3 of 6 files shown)


Astro-Han:
设计很干净。package exports + self-referencing 是正解，也为后续 monorepo 迁移铺好了路。

提一个加固建议：shim 移除后，symlink 成了用户 CLI 解析的唯一通道。当前如果 symlink 位置已经是普通目录而非 symlink，`unlink()` 删不掉目录，`symlink()` 因 EEXIST 失败，外层 catch 静默吞掉异常，用户侧无任何提示。

建议把 `unlink` 换成 `rm`：

```typescript
try { await fs.promises.rm(symlinkPath, { recursive: true, force: true }); } catch { /* doesn't exist */ }
```

外层 catch 也值得加一行 `console.warn`，方便排障。

jackwener:
Thanks for the review! Both suggestions applied:

1. `unlink()` → `rm({ recursive: true, force: true })` — handles stale directories that `unlink` can't remove
2. Outer catch now logs a warning via `log.warn()` instead of silently swallowing the error

Pushed in 4877213.

Astro-Han:
Post-merge follow-up: the symlink approach has a gap I missed during review.

The symlink at `~/.opencli/node_modules/@jackwener/opencli` correctly resolves package exports (`@jackwener/opencli/registry`, etc.), but adapters also import third-party packages as bare specifiers (`chalk`, `turndown`). These resolve by walking up from `~/.opencli/clis/<site>/`, which never reaches the global `node_modules/` where the actual packages live.

This is now surfacing as the remaining failures in #792 (1.6.5).

Suggested fix: extend `ensureUserCliCompatShims` to also symlink each dependency from `package.json` into `~/.opencli/node_modules/`, derived from the package root's parent directory (the global modules tree).

Sorry for missing this in the original review.

PR #801, opened 2026-04-05
fix: stop modifying shell rc files in postinstall

### LINKED ISSUE #788: bug: `opencli --version` reports `0.0.0` when installed from source + `Failed to load module` for user clis

jackwener:
## Summary
- Remove all `.zshrc` / `.bashrc` modification logic from `postinstall.js`
- Only write completion files to disk (`~/.zsh/completions/_opencli`, `~/.bash_completion.d/opencli`, fish)
- Print clear setup instructions telling the user what to add to their rc file

## Motivation
The previous approach parsed `.zshrc` to find a `compinit` line and spliced `fpath=(...)` before it. This broke multi-line shell commands — e.g. zinit blocks where `zicompinit` matched `/compinit/` inside a backslash-continued line, causing the insertion to split the command block and corrupt the user's shell config (#788).

Rather than trying to fix the parser (which can never be fully safe against arbitrary shell syntax), we adopt the approach used by rustup, homebrew, and other CLI tools: **write the file, print the instructions, let the user decide where to put it**.

## Before → After

| Shell | Before | After |
|-------|--------|-------|
| zsh | Wrote `_opencli` + modified `.zshrc` (splice/append) | Writes `_opencli` + prints fpath/compinit instructions |
| bash | Wrote `opencli` + modified `.bashrc` (append) | Writes `opencli` + prints source instruction |
| fish | Wrote `opencli.fish` (no rc change) | Same — no change |

## Test plan
- [ ] `npm install -g` prints completion setup instructions for zsh
- [ ] `.zshrc` is not modified
- [ ] `.bashrc` is not modified  
- [ ] Completion works after user manually adds the printed lines

Closes #788

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,11 +1,13 @@
 #!/usr/bin/env node
 
 /**
- * postinstall script — automatically install shell completion files.
+ * postinstall script — install shell completion files and print setup instructions.
  *
  * Detects the user's default shell and writes the completion script to the
- * standard system completion directory so that tab-completion works immediately
- * after `npm install -g`.
+ * standard completion directory.  For zsh and bash, the script prints manual
+ * instructions instead of modifying rc files (~/.zshrc, ~/.bashrc) — this
+ * avoids breaking multi-line shell commands and other fragile rc structures.
+ * Fish completions work automatically without rc changes.
  *
  * Supported shells: bash, zsh, fish.
  *
@@ -13,7 +15,7 @@
  * the main source tree) so that it can run without a build step.
  */
 
-import { mkdirSync, writeFileSync, existsSync, readFileSync, appendFileSync } from 'node:fs';
+import { mkdirSync, writeFileSync, existsSync } from 'node:fs';
 import { join } from 'node:path';
 import { homedir } from 'node:os';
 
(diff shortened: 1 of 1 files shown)


jackwener:
Pushed the changes to PR #796 fork branch instead, as requested.

Issue #810, opened 2026-04-05
[Feature]: unify network interception across explore, diagnostic, and operate


Astro-Han:
End-to-end tested the full adapter lifecycle (explore → generate → plugin → operate → diagnostic → repair) on v1.6.7. The individual pieces mostly work, but they don't share browser instrumentation, which limits the AI agent story.

All claims below have been verified by both static code analysis and runtime reproduction on v1.6.7.

## What was tested

| Step | Result | Notes |
|------|--------|-------|
| explore | Runs, but 0 API endpoints on HN and Lobsters | Compensation strategy (iframe re-fetch) exists but rarely triggers |
| generate | Fails (0 candidates) | Depends on explore finding APIs; confirmed: synthesize hackernews → 0 |
| plugin create | Scaffold OK | YAML template uses nonexistent `extract` pipeline step (separate bug) |
| plugin install (TS) | Compile OK, runtime OK | `linkHostOpencli()` in `plugin.ts` already creates `node_modules/@jackwener/opencli` symlink |
| operate open/state/eval | All work | |
| operate network | 0 requests after open | Interceptor injected after goto + wait, misses initial load. Confirmed at runtime. |
| operate network (post-eval) | Works | Manual fetch after open is captured correctly (visible with `--all`) |
| operate init + verify | Works | Template uses correct package exports |
| diagnostic (TS) | Works | error.code and source both correct |
| diagnostic (YAML) | source missing | Covered by #808 |
| diagnostic (browser page state) | Minimal info | networkRequests 0, consoleErrors [] (see notes below) |
| repair loop (manual) | Each step works individually | diagnostic → operate explore → patch → verify all functional |

> **Note on `consoleErrors: []`**: `base-page.ts` `consoleMessages()` returns an empty array unconditionally with no subclass override. So diagnostic's `consoleErrors: []` does not mean "no errors on page"; it means console errors are not collected at all. This further weakens diagnostic info quality.

## Core finding: fragmented network interception paths

The codebase has at least four independent network capture mechanisms (three relevant to the explore/diagnostic/operate story, plus one in record):

| System | Global var | Used by | Captures |
|--------|-----------|---------|----------|
| `NETWORK_INTERCEPTOR_JS` (cli.ts:310) | `__opencli_net` | operate | fetch/XHR with method, status, body. Injected after page load, so misses initial requests. |
[...]

PR #812, opened 2026-04-05
fix: replace nonexistent extract step in plugin scaffold

jackwener:
## Summary
- Plugin scaffold YAML template used `extract` pipeline step which doesn't exist in the registry
- Replaced with `select` (JSONPath selection), which is the correct step for this operation
- Fixes the `Unknown pipeline step "extract"` error when running scaffolded plugins

Part of #810

## Test plan
- [ ] `opencli plugin create test-plugin` generates valid YAML
- [ ] Running the scaffolded `hello` command works end-to-end

### DIFF
--- a/src/plugin-scaffold.ts
+++ b/src/plugin-scaffold.ts
@@ -92,9 +92,7 @@ pipeline:
   - fetch:
       url: "https://httpbin.org/get?greeting=hello"
       method: GET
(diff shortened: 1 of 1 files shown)


jackwener:
Closing in favor of opus0's #814 (same fix, both approved)

PR #828, opened 2026-04-06
feat(doubao): add --thread support for ask/send/read

hanxuanliang:
## Description

Add optional `--thread <id|url>` support to `opencli doubao ask`, `send`, and `read` so they can target a specific existing Doubao conversation instead of relying on the currently reused browser tab.

This keeps the semantics aligned with `opencli doubao detail <id>`:
- accepts either a numeric conversation id
- or a full `https://www.doubao.com/chat/<id>` URL

Also included:
- focused adapter tests for `ask`, `send`, and `read`
- updated adapter docs with `--thread` examples

Related issue:
- N/A

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [x] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

```bash
$ npx vitest run clis/doubao/*.test.ts

 RUN  v4.1.1 /Users/hanfox/Desktop/code/self/agent/agentx/iamwho/opencli

 Test Files  6 passed (6)
      Tests  11 passed (11)
```

### DIFF
--- a/clis/doubao/ask.ts
+++ b/clis/doubao/ask.ts
@@ -1,6 +1,14 @@
 import { cli, Strategy } from '@jackwener/opencli/registry';
 import type { IPage } from '@jackwener/opencli/types';
-import { DOUBAO_DOMAIN, getDoubaoTranscriptLines, getDoubaoVisibleTurns, sendDoubaoMessage, waitForDoubaoResponse } from './utils.js';
+import {
+  DOUBAO_DOMAIN,
+  getDoubaoTranscriptLines,
+  getDoubaoVisibleTurns,
+  navigateToConversation,
+  parseDoubaoConversationId,
+  sendDoubaoMessage,
+  waitForDoubaoResponse,
+} from './utils.js';
 
 export const askCommand = cli({
   site: 'doubao',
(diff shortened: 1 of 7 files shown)


Astro-Han:
Left one P2 note on `--thread` validation. The issue is not isolated to one command, it comes from a shared helper path used by `ask`, `send`, and `read`: malformed or partial thread values can fall through, be treated as a match for the current URL, and end up operating on the wrong conversation instead of failing early.

Other than that, this looks solid to me. The scope is tight, the command wiring is consistent across `ask`, `send`, and `read`, and the added tests cover the main happy path well.

I'm reviewing this as a contributor, not a maintainer, but from that angle I think the PR is in good shape once that shared edge case is addressed.

clis/doubao/ask.ts
```diff
@@ -13,12 +21,19 @@ export const askCommand = cli({
...
     { name: 'timeout', required: false, help: 'Max seconds to wait (default: 60)', default: '60' },
   ],
   columns: ['Role', 'Text'],
   func: async (page: IPage, kwargs: any) => {
     const text = kwargs.text as string;
+    const thread = typeof kwargs.thread === 'string' ? kwargs.thread.trim() : '';
     const timeout = parseInt(kwargs.timeout as string, 10) || 60;
+
+    if (thread) {
+      await navigateToConversation(page, parseDoubaoConversationId(thread));
```
Astro-Han, line 34: If `--thread` is malformed or only a partial id, `parseDoubaoConversationId()` falls back to the raw input. That becomes risky here because `navigateToConversation()` treats `currentUrl.includes(`/chat/${conversationId}`)` as an exact match, so a value like `123` will match an existing `/chat/1234567890123` URL and skip navigation. In that case `ask` can post into the wrong conversation instead of failing fast. I think this should validate the parsed id before navigation, or tighten the current URL check to compare the full path segment.

PR #834, opened 2026-04-06
feat(browser): unify page capture across daemon and CDP

Astro-Han:
## Description

Implement the unified page capture work for issue #810 across daemon-backed and direct CDP pages.

This wires structured capture into execution and diagnostics, preserves daemon capture intent across navigation reattach, restores stale-extension compatibility paths, and adds regression coverage for the new lifecycle and fallback behavior.

Related issue:
- #810

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [x] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

```text
npm test
Test Files  46 passed (46)
Tests 578 passed | 1 skipped (579)

npx tsc --noEmit
passed

cd extension && npm run build
vite build completed successfully
```


Astro-Han (commented):
### Context

This PR implements T3-1 (unified IPage capture interface) and T3-4 (real consoleMessages collection) from the #810 roadmap, following:

- #829 (T2-1): diagnostic reads intercepted payloads (merged)
- #831 (T2-2): explore contentType URL inference (open)

### What this does for diagnostic and auto-fix

The goal of #810 is making `opencli fix` work better. When a command fails, the fix loop needs to know what happened in the browser. Before this PR, that data was poor:

- Network requests came from the Performance API (URL and timing only, no headers or response body). The real CDP capture existed but was split across daemon and CDP paths, and got wiped on every navigation.
- `consoleMessages()` returned `[]` in daemon mode. Uncaught exceptions and failed API calls were invisible.

Now, diagnostic reads from `readNetworkCapture()` first (full request/response with headers and body preview), falling back to the Performance API only when capture data is empty. `consoleMessages('error')` returns real warn + error entries from CDP Runtime events in both modes. Capture is command-scoped: starts before `runCommand`, stops in `finally`, and survives mid-command navigation via the `captureIntents` persistence mechanism.

In practice, `opencli fix` now sees the actual 401/403/500 responses, CORS errors, and console stack traces instead of guessing from a DOM snapshot.

### Review notes

I fixed five items in v2 after self-review:

1. Preferred-tab drift guard for owned sessions (resolveTab now validates window ID)
2. Doctor version check relaxed to major.minor (exact match was too noisy)
3. Console non-drain semantics documented (readConsoleCapture is intentionally a rolling buffer, unlike readNetworkCapture which drains)
4. `ensureAttached` early-return dependency comment (armCapture handles Runtime.enable, so the generic call is skipped)
5. Workspace tab cache atomic writes (temp file + rename)

Capture methods stay optional on `IPage` (with a `CaptureCapablePage` utility type) to preserve backward compatibility for external implementers.

### Follow-ups

This PR is the infrastructure layer. Still ahead from #810:
- Migrate `explore` and `operate` to use `readNetworkCapture` as primary data source
- T3-2 (adapter-repair loop) and T3-3 (adapter generation preset) need design discussion

Astro-Han:
The CI failure came from a change pulled in from main, not from the unified capture work in this PR. GitHub runs checks on the PR’s merge commit, and once we merged main to clear conflicts, the failing path was `src/clis/binance/commands.test.ts`. The error was a broken Binance test import, and the stack trace never touched the capture files changed here. That’s why I reverted the Binance test fix instead of keeping an unrelated change in this PR.

Astro-Han:
Latest round is pushed.

Final fixes in this update:
- hardened capture lifecycle across detach, reattach, and stale-attach recovery
- fixed stale-extension fallback so interceptor install/read/uninstall works across fresh CLI invocations
- restored `browser open` capture boundary semantics by resetting capture before each open
- kept cleanup from retrying `stopCapture()` after window close on the success path

Verification rerun after the last fixes:
- `npm test` passed: 194 files, 1506 passed, 2 skipped
- `npx tsc --noEmit` passed
- final fresh-eye review found no P0/P1 issues

Residual risk remains the already documented navigation detach-window gap in the spec, and I did not find a different regression beyond that accepted limitation.

PR #839, opened 2026-04-06
fix(xiaohongshu): scope note interaction selectors to .interact-container

y4h2:
## Summary

`xiaohongshu note` was returning wrong like/collect/comment counts for any note that had user comments.

## Root cause

The DOM selectors `.like-wrapper .count` / `.collect-wrapper .count` / `.chat-wrapper .count` are not unique to the post's main interaction bar. Every comment in the comment section also renders a `.like-wrapper > .count` for its own like button (and the comment count input has `.chat-wrapper`). `document.querySelector` returns the **first** match, which on a note with comments is a comment's count, not the post's.

In practice the post's true counts live inside `.interact-container`, while comment-level counts are scattered throughout the comment list above/below it. Whether the bug fired depended on DOM order, which is why the bug surfaced inconsistently across different notes.

## Fix

Scope each selector to `.interact-container`:

```diff
- const likes    = clean(document.querySelector('.like-wrapper .count'))
- const collects = clean(document.querySelector('.collect-wrapper .count'))
- const comments = clean(document.querySelector('.chat-wrapper .count'))
+ const likes    = clean(document.querySelector('.interact-container .like-wrapper .count'))
+ const collects = clean(document.querySelector('.interact-container .collect-wrapper .count'))
+ const comments = clean(document.querySelector('.interact-container .chat-wrapper .count'))
```

## Verification

Tested on three notes that previously returned wrong values:

| Note | Before (likes) | After (likes) | Real value |
|---|---|---|---|
| A (notId 69d15115...) | 2 | **74** | 74 ✓ |
| B (noteId 69c176ea...) | 1 | **796** | 796 ✓ |
| C (noteId 69cc56c3...) | 1 | **269** | 269 ✓ |

All three now match the values displayed on xiaohongshu.com.

## Test plan

- [x] Run `opencli xiaohongshu note <url>` against several notes with comments
- [x] Verify likes / collects / comments match what's shown on the page
- [ ] Maintainer: run existing test suite (`note.test.ts` if any)

### DIFF
--- a/clis/xiaohongshu/note.test.ts
+++ b/clis/xiaohongshu/note.test.ts
@@ -199,6 +199,20 @@ describe('xiaohongshu note', () => {
     expect(result.find((r: any) => r.field === 'comments')!.value).toBe('0');
   });
 
+  it('scopes metric selectors to .interact-container to avoid matching comment like buttons', async () => {
+    const page = createPageMock({
+      loginWall: false, notFound: false,
+      title: 'Test', desc: '', author: 'Author', likes: '10', collects: '5', comments: '3', tags: [],
+    });
+
+    await command!.func!(page, { 'note-id': 'abc123' });
+
+    const evaluateScript: string = (page.evaluate as any).mock.calls[0][0];
+    expect(evaluateScript).toContain('.interact-container .like-wrapper .count');
+    expect(evaluateScript).toContain('.interact-container .collect-wrapper .count');
+    expect(evaluateScript).toContain('.interact-container .chat-wrapper .count');
+  });
+
   it('omits tags row when no tags present', async () => {
     const page = createPageMock({
       loginWall: false, notFound: false,


Astro-Han:
This fix changes the selector inside `page.evaluate`, but the current test still only asserts the mocked `evaluate` result, so the old broken selectors could slip back in without any test failing.

It may be worth adding a small regression check on `page.evaluate.mock.calls[0][0]` to verify the metric selectors stay scoped to `.interact-container`. The repo already uses that pattern in other Xiaohongshu tests.

y4h2:
@Astro-Han Thanks for the suggestion! Added a regression test as advised, following the same pattern used in comments.test.ts.

Astro-Han:
LGTM from my side. The follow-up test covers the selector regression I called out.

PR #840, opened 2026-04-07
fix: fallback to undiciFetch when globalThis.fetch is unavailable

deep-innovator:
## Problem
On Windows with older Node.js versions, `globalThis.fetch` is not available at module load time, causing the original `nativeFetch` definition to fail immediately.

## Solution
- Keep `nativeFetch` as a direct request function, preserving the original signature `(input, init) => Promise<Response>`.
- Fall back to `undiciFetch` synchronously when `globalThis.fetch` is unavailable.
- This maintains the existing no-proxy path behavior while fixing the startup error on Windows.

## Verification
- No breaking changes to the call site `return nativeFetch(input, init)`.
- Works both with native `fetch` and the `undiciFetch` fallback.

### DIFF
--- a/src/node-network.ts
+++ b/src/node-network.ts
@@ -39,7 +39,9 @@ interface ProxyConfig {
 let installed = false;
 const directDispatcher = new Agent();
 const proxyDispatcherCache = new Map<string, Dispatcher>();
-const nativeFetch = globalThis.fetch.bind(globalThis);
+const nativeFetch = globalThis.fetch
+  ? globalThis.fetch.bind(globalThis)
+  : (input, init) => undiciFetch(input, init);
 
 function readEnv(env: NodeJS.ProcessEnv, lower: string, upper: string): string | undefined {
   const lowerValue = env[lower];


Astro-Han:
I reproduced the current failure by deleting `globalThis.fetch` before importing `src/node-network.ts`. On `main`, that throws at the eager `bind`, so this PR is fixing the right spot.

One thing still missing is a regression test for the no-`fetch` module-load path. The current tests cover proxy decisions after import, but not the startup case this change is meant to fix.

Also worth confirming the support target here. `package.json` still says `node >=20`, so this is either a best-effort compatibility fix for unsupported runtimes or a sign that the engine floor should be revisited.

deep-innovator:
Thanks for the review! I'll wait for your further guidance on this



---- Replied Message ----
| From | ***@***.***> |
| Date | 04/07/2026 20:43 |
| To | ***@***.***> |
| Cc | ***@***.***>***@***.***> |
| Subject | Re: [jackwener/opencli] fix: fallback to undiciFetch when globalThis.fetch is unavailable (PR #840) |
Astro-Han left a comment (jackwener/opencli#840)

I reproduced the current failure by deleting globalThis.fetch before importing src/node-network.ts. On main, that throws at the eager bind, so this PR is fixing the right spot.

One thing still missing is a regression test for the no-fetch module-load path. The current tests cover proxy decisions after import, but not the startup case this change is meant to fix.

Also worth confirming the support target here. package.json still says node >=20, so this is either a best-effort compatibility fix for unsupported runtimes or a sign that the engine floor should be revisited.

src/node-network.ts
```diff
@@ -39,7 +39,9 @@ interface ProxyConfig {
 let installed = false;
 const directDispatcher = new Agent();
 const proxyDispatcherCache = new Map<string, Dispatcher>();
-const nativeFetch = globalThis.fetch.bind(globalThis);
+const nativeFetch = globalThis.fetch
```
Astro-Han, line 42: Would it be simpler to make `nativeFetch` a tiny wrapper instead of choosing once at module load? As written, if `fetch` is missing during import but gets installed later, the direct path is still pinned to `undiciFetch` for the rest of the process.

Issue #722, opened 2026-04-03
[Bug]: cannot find undici when install plugin.


kirin233x:
### Description

here is my package.json of my plugin.
{
  "name": "opencli-plugin-codehub",
  "version": "1.0.1",
  "type": "module",
  "peerDependencies": {
    "@jackwener/opencli": ">=1.0.0"
  },
  "dependencies": {
    "undici": "^7.0.0"
  },
  "devDependencies": {
    "@types/node": "^25.5.0",
    "typescript": "^6.0.2"
  }
}

when I install plugin
Plugin codehub/commands.js: Cannot find package 'undici' imported from /Users/xxxxxx/.opencli/monorepos/opencli-plugins-app-team/packages/codehub/codehub.js

### Steps to Reproduce

1. Run `opencli ...`
2. ...
3. See error

### Expected Behavior

install success.

### OpenCLI Version

1.6.1

### Node.js Version

22.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

Astro-Han:
Thanks for the report. I looked through the current plugin install flow, and this seems to be a real gap in OpenCLI's monorepo plugin handling, not just a one-off `undici` problem.

From the current implementation, monorepo sub-plugins are installed with this assumption:

1. OpenCLI clones the monorepo once
2. It runs `npm install --omit=dev` only at the monorepo root
3. It then symlinks each sub-plugin into `~/.opencli/plugins/...`
4. For TS plugins, it links the host `@jackwener/opencli` package and transpiles `.ts` to `.js`

The important part is step 2: OpenCLI does not install dependencies inside each sub-plugin directory. It assumes the monorepo root is already set up to provide those dependencies, typically via npm workspaces.

That assumption is currently not validated.

So if a sub-plugin has its own runtime dependencies, for example `undici` in its `package.json`, but the monorepo root is not correctly configured to install and expose that package for the sub-plugin, the install step can appear to succeed, while the plugin later fails at runtime with:

`Cannot find package 'undici'`

That matches the behavior in this issue.

Concretely, the code today does the following:

- `installDependencies(dir)` runs `npm install --omit=dev` for the given directory if it has a `package.json`
- standalone plugins call that on the plugin directory
- monorepo plugins call that only once on the repo root, not on each sub-plugin
- monorepo sub-plugin validation checks plugin structure, but does not check whether the root workspace setup will actually install the sub-plugin's dependencies

Because of that, the current behavior can become:

- install succeeds
[...]

PR #852, opened 2026-04-07
fix: resolve plugin host root from published dist layout

### LINKED ISSUE #843: [Bug]: install plugins get a warning, cannot find @jackwener/opencli/registry
### Description

⚠  Plugin xxx-dts/dts.js: Cannot find module '/Users/xxxx/.opencli/plugins/xxxx-dts/node_modules/@jackwener/opencli/registry' imported from /Users/xxxx/.opencli/plugins/xxx-dts/dts.js
Did you mean to import "@jackwener/opencli/dist/src/registry-api.js"?

### Steps to Reproduce

1. Run `opencli  plugin install xxxx`

### Expected Behavior

no errors， no warnings.

### OpenCLI Version

1.6.8

### Node.js Version

22.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

KehaoC:
## Summary

Fixes plugin host package linking for published builds by resolving the real `@jackwener/opencli` package root from the running module path instead of assuming it is always one directory up.

In published installs, `src/plugin.ts` compiles under `dist/src/plugin.js`, so the old `dirname + '..'` logic could point plugin `node_modules/@jackwener/opencli` at `dist/` instead of the package root. That breaks package export resolution for imports like `@jackwener/opencli/registry`.

## Changes

- Add `resolveHostOpencliRoot()` to walk upward until it finds the host package `package.json`.
- Use that helper for plugin host symlinking and esbuild binary lookup.
- Add a regression test for resolving from a compiled `dist/src/plugin.js` layout.

## Validation

- `npm test -- --run src/plugin.test.ts`
- `npm run typecheck`

Fixes #843
Refs #722

### DIFF
--- a/src/plugin.test.ts
+++ b/src/plugin.test.ts
@@ -27,6 +27,7 @@ const {
   _readLockFile,
   _readLockFileWithWriter,
   _resolveEsbuildBin,
+  _resolveHostOpencliRoot,
   uninstallPlugin,
   updatePlugin,
   _parseSource,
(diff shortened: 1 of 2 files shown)


jackwener (approved):
LGTM. Clean fix.

**Problem**: `path.dirname(thisFile) + '..'` assumed the module is always one level deep (src/plugin.ts or dist/plugin.js). But in published builds, the compiled file lives at `dist/src/plugin.js` — two levels deep — so `..` resolves to `dist/` instead of the package root, breaking plugin host symlinks and esbuild resolution.

**Solution**: `resolveHostOpencliRoot()` walks upward from the current file until it finds a `package.json` with `name: '@jackwener/opencli'`. Falls back to the old `dirname + '..'` if nothing matches, preserving backward compatibility.

**Review notes**:
- Walk-up logic is correct — checks `pkg.name` to avoid stopping at nested package.json files (e.g. in node_modules)
- The `catch {}` on JSON parse is appropriate — a malformed package.json shouldn't halt the search
- Both call sites (`linkHostOpencli` and `resolveEsbuildBin`) are updated
- Regression test covers the exact scenario (dist/src layout)
- CI all green

No issues found.

PR #855, opened 2026-04-07
fix(twitter): always use compose URL for reply to fix missing textarea

### LINKED ISSUE #848: twitter reply fails with: 'Could not find the reply text area'
## Repro

On a logged-in Chrome session where `opencli doctor` passes and `opencli twitter post "..."` works, run:

```bash
opencli twitter reply "https://x.com/<user>/status/<tweet_id>" "hello"
```

Observed output:

```text
- status: failed
  message: Could not find the reply text area. Are you logged in?
```

## What I checked

As of **April 7, 2026**, the current source already calls `page.goto(kwargs.url)` in `clis/twitter/reply.ts` for text-only replies, so this does **not** look like a simple “missing navigation” bug.

The more likely problem is the wait condition:

- text-only flow does:
  - `await page.goto(kwargs.url)`
  - `await page.wait({ selector: '[data-testid="primaryColumn"]' })`
  - then immediately queries `document.querySelector('[data-testid="tweetTextarea_0"]')`
- but on X, `primaryColumn` is not specific to the reply composer
- the inline reply textarea can appear later than `primaryColumn`, or the SPA can still be settling when the selector check runs
- result: reply fails with `Could not find the reply text area` even though the user is logged in

## Suggested fix

Use the dedicated reply composer for **all** replies (not only image replies), and wait for the actual reply textarea before typing:

```diff
diff --git a/clis/twitter/reply.ts b/clis/twitter/reply.ts
@@
-      if (localImagePath) {
-        await page.goto(buildReplyComposerUrl(kwargs.url));
-        await page.wait({ selector: '[data-testid="tweetTextarea_0"]' });
-        await page.wait({ selector: REPLY_FILE_INPUT_SELECTOR, timeout: 20 });
-        await attachReplyImage(page, localImagePath);
-      } else {
-        await page.goto(kwargs.url);
-        await page.wait({ selector: '[data-testid="primaryColumn"]' });
-      }
+      await page.goto(buildReplyComposerUrl(kwargs.url), { waitUntil: 'load', settleMs: 2500 });
+      await page.wait({ selector: '[data-testid="tweetTextarea_0"]', timeout: 8 });
+      if (localImagePath) {
+        await page.wait({ selector: REPLY_FILE_INPUT_SELECTOR, timeout: 20 });
+        await attachReplyImage(page, localImagePath);
+      }
```

## Why this helps

- avoids relying on the inline tweet-page reply box
- waits on the actual composer needed by the command
- makes text-only and image reply paths consistent
- matches the user-visible failure mode more closely than a login problem

## Validation

I applied the patch locally and the existing adapter test file still passes after updating expectations:

```bash
npx vitest run --project adapter clis/twitter/reply.test.ts
```

Result:

```text
Test Files  1 passed (1)
Tests       7 passed (7)
```

winfred5432:
## Problem

When calling `opencli twitter reply` without an image, the code navigated to the tweet page URL and waited for `[data-testid="primaryColumn"]`. The inline reply composer is not reliably mounted in that context, causing:

```
Error: Could not find reply text area
```

This was a silent regression — the image path already used the reply composer URL correctly, but the text-only path did not.

## Fix

Unconditionally navigate to the reply composer URL (`/compose/post?in_reply_to=<tweet_id>`) for all replies, then wait for `[data-testid="tweetTextarea_0"]` before proceeding. The image attachment flow is unchanged.

## Testing

Verified locally with `opencli twitter reply --url <tweet_url> --text "..."` — reply posts successfully on both text-only and image paths.

Fixes #848

### DIFF
--- a/clis/twitter/reply.test.ts
+++ b/clis/twitter/reply.test.ts
@@ -53,8 +53,11 @@ describe('twitter reply command', () => {
       text: 'text-only reply',
     });
 
-    expect(page.goto).toHaveBeenCalledWith('https://x.com/_kop6/status/2040254679301718161?s=20');
-    expect(page.wait).toHaveBeenCalledWith({ selector: '[data-testid="primaryColumn"]' });
+    expect(page.goto).toHaveBeenCalledWith(
+      'https://x.com/compose/post?in_reply_to=2040254679301718161',
+      { waitUntil: 'load', settleMs: 2500 }
+    );
+    expect(page.wait).toHaveBeenCalledWith({ selector: '[data-testid="tweetTextarea_0"]', timeout: 8 });
     expect(result).toEqual([
       {
         status: 'success',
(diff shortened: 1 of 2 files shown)


Astro-Han:
I reviewed this locally. It matches the failure mode I reproduced, and `npx vitest run clis/twitter/reply.test.ts` passes on the PR branch. Looks good to me.

jackwener:
Superseded by #860 (same fix, already merged) + #862 (timeout relaxation). Thanks for the contribution!

PR #859, opened 2026-04-07
fix(antigravity): implement configurable timeout and auto-reconnect for serve

### LINKED ISSUE #861: Antigravity serve 120s Timeout & CDP Connection Instability
## 1. 问题描述
在使用 \opencli antigravity serve\ 时，观察到以下两个主要问题：
1. **硬超时限制**: 代理服务器内部 \waitForReply\ 逻辑硬编码了 120 秒超时。对于耗时较长的生成任务（如复杂代码重构），会导致客户端收到超时错误，即使 IDE 仍在生成。
2. **连接不稳定性**: 当其他进程（如 Pager 的 Fallback 机制）调用 \Antigravity dump\ 或 \
ead\ 时，由于 CDP 的排他性，正在运行的 \serve\ 会丢失连接。目前 \waitForReply\ 缺乏重试机制，导致请求直接失败。

## 2. 改进方向
- **配置化超时**: 引入命名参数 \--timeout\ 和环境变量 \OPENCLI_ANTIGRAVITY_TIMEOUT\ 以便灵活调整单个任务的最大存活时间，默认为 120s。
- **鲁棒的轮询逻辑**: 在 \waitForReply\ 轮询循环中加入 \	ry-catch\ 并在报错（如 \Session closed\，\CDP connection lost\）时调用 \opts.reconnect\ 函数适配器，底层使用 \nsureConnected()\ 在保持页面内容和 DOM 不变的情况下恢复监控通信。

deepziyu:
Resolves #861

Resolves the 120s hardcoded timeout issue and CDP connection exclusion problem in \ntigravity serve\ command. Adds \--timeout\ option and \OPENCLI_ANTIGRAVITY_TIMEOUT\ support. Implements auto-reconnect logic via \nsureConnected\ during polling.

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -201,7 +201,7 @@ npm link
 | **discord-app** | `status` `send` `read` `channels` `servers` `search` `members` | 桌面端 |
 | **v2ex** | `hot` `latest` `topic` `node` `user` `member` `replies` `nodes` `daily` `me` `notifications` | 公开 / 浏览器 |
 | **xueqiu** | `feed` `hot-stock` `hot` `search` `stock` `comments` `watchlist` `earnings-date` `fund-holdings` `fund-snapshot` | 浏览器 |
-| **antigravity** | `status` `send` `read` `new` `dump` `extract-code` `model` `watch` | 桌面端 |
+| **antigravity** | `status` `send` `read` `new` `dump` `extract-code` `model` `watch` `serve` | 桌面端 |
 | **chatgpt-app** | `status` `new` `send` `read` `ask` `model` | 桌面端 |
 | **xiaohongshu** | `search` `notifications` `feed` `user` `download` `publish` `creator-notes` `creator-note-detail` `creator-notes-summary` `creator-profile` `creator-stats` | 浏览器 |
 | **xiaoe** | `courses` `detail` `catalog` `play-url` `content` | 浏览器 |
--- a/clis/antigravity/serve.js
+++ b/clis/antigravity/serve.js
@@ -13,7 +13,6 @@ import { createServer } from 'node:http';
 import { CDPBridge } from '@jackwener/opencli/browser/cdp';
 import { resolveElectronEndpoint } from '@jackwener/opencli/launcher';
 import { EXIT_CODES, getErrorMessage } from '@jackwener/opencli/errors';
-import { parseEnvTimeout, parseTimeoutValue } from '../../src/runtime.js';
 // ─── Helpers ─────────────────────────────────────────────────────────
 function generateMsgId() {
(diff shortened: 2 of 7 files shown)


Astro-Han:
I think this PR is moving in the right direction overall. Making the `antigravity serve` timeout configurable and trying to recover from dropped CDP sessions both make sense for a long-running local proxy.

My main concern is the new reconnect path in `clis/antigravity/serve.ts`, inside the `catch (err)` block in `waitForReply()`, especially the `if (opts.reconnect) { ... continue; }` branch. Right now it treats any polling error from `isGenerating(page)` or `getConversationText(page)` as a reconnect case. That is broader than the problem this PR is trying to solve. A lost CDP session should reconnect, but a DOM change, selector break, or logic bug should still fail directly. Otherwise the original error gets hidden and the request eventually turns into a generic timeout.

So from a design point of view, I like the direction of configurable timeout plus reconnect, but I think the reconnect policy needs a tighter boundary. I would limit retry to explicit transport or session-loss errors and rethrow everything else.

deepziyu:
Refined the reconnection logic in waitForReply as suggested:

1. **Narrowed error scope**: The catch block now specifically filters for CDP transport or session loss errors (regex: /closed|lost|not open|websocket/i). Genuine logical or execution errors will be rethrown immediately to avoid masking bugs.
2. **Added retry limit**: Reconnection is now capped at a maximum of 2 attempts (maxtime = 2) per polling session.
3. **Regression Check**: Verified locally that 14 overall unit test failures match the 12 baseline failures plus environment-specific noise (consistent with main branch state on this machine).

Astro-Han:
I checked the latest head. The reconnect change looks much better now. Restricting retries to session-loss-style errors and rethrowing other failures addresses the concern I raised earlier about masking DOM or selector bugs behind a later timeout.

I still see one separate issue with the new timeout handling. In `clis/antigravity/serve.ts`, `OPENCLI_ANTIGRAVITY_TIMEOUT` is parsed with `parseInt(...) * 1000`, and in `src/cli.ts` the `--timeout` value is forwarded with `parseInt(opts.timeout)`. Invalid values like `abc` turn into `NaN`, and negative values are accepted too. That can make the proxy fail immediately or behave oddly instead of falling back cleanly. Since `src/runtime.ts` already has validated timeout parsing, I think this path should follow the same pattern.

deepziyu:
Thank you for the detailed feedback, @Astro-Han! 

I have standardized the timeout parsing as suggested:
1. **Refactored Runtime Utility**: Exported and generalized the parsing logic in \src/runtime.ts\ into a new \parseTimeoutValue\ function that handles \NaN\, \

PR #930, opened 2026-04-10
fix(security): escape codegen strings and redact diagnostic body

jackwener:
## Summary

Fixes two bugs identified during post-migration first-principles review:

- **candidateToJs string injection**: `site`, `name`, `domain`, and arg `name`/`type` fields were written into generated JS adapter files without escaping single quotes. If any of these contained `'`, the generated adapter would have a syntax error. Now all string literals in codegen use `.replace(/'/g, "\\'")`
- **diagnostic body data leak**: `redactNetworkRequest` truncated the `body` field but did not pass it through `redactText()`, meaning JWTs or bearer tokens in response bodies could leak into AI repair context. `responseBody` and `responsePreview` already used `sanitizeCapturedValue` → `redactText`, but `body` was missed

## Test plan

- [x] `src/generate-verified.test.ts` — 28 tests passed
- [x] `src/diagnostic.test.ts` — 15 tests passed

### DIFF
--- a/src/diagnostic.ts
+++ b/src/diagnostic.ts
@@ -183,9 +183,9 @@ function redactNetworkRequest(req: unknown): unknown {
     redacted.responseHeaders = redactHeaders(redacted.responseHeaders as Record<string, string>);
   }
 
-  // Truncate response body
+  // Redact and truncate response body
   if (typeof redacted.body === 'string') {
-    redacted.body = truncate(redacted.body, MAX_REQUEST_BODY_CHARS);
+    redacted.body = redactText(truncate(redacted.body, MAX_REQUEST_BODY_CHARS));
   }
   if ('responseBody' in redacted) {
     redacted.responseBody = sanitizeCapturedValue(redacted.responseBody);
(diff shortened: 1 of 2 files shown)


hiSandog:
Reviewed the diff against local code. The fix correctly addresses both issues:

1. **codegen string escaping** in `candidateToJs`: The patch adds `.replace(/'/g, "\\'")` to `site`, `name`, `domain`, `name` (arg), and `type` (arg) fields. Verified these were the only unescaped string interpolations in the function — `description` and `help` already had escaping.

2. **diagnostic body redaction**: Wrapping `truncate()` with `redactText()` in `redactNetworkRequest` closes the gap where `body` bypassed the JWT/bearer token sanitization that `responseBody` and `responsePreview` already get via `sanitizeCapturedValue`.

One minor suggestion: consider adding a unit test that feeds a candidate with single quotes in `site`/`name`/arg `name` to `candidateToJs` and asserts the generated JS is syntactically valid (e.g., via `new Function()` or `acorn.parse()`). The PR mentions tests pass but doesn't show an explicit injection test case for these newly-escaped fields.

PR #957, opened 2026-04-11
feat(scys): add SCYS browser adapters

warkcod:
## Summary
- add SCYS browser adapters for course, toc, read, feed, opportunity, activity, and article extraction
- add shared SCYS extraction, course image download, and normalization helpers with adapter test coverage
- add SCYS adapter docs, schema guidance, and regenerate the CLI manifest

## Test Plan
- [x] npm run build
- [x] npm run typecheck
- [x] npm test
- [x] npx vitest run --project adapter
- [x] bash scripts/check-doc-coverage.sh --strict


Astro-Han:
There is a lot of work here, and the adapter itself looks substantial. My only review note is about deliverables around discoverability.

From what I can tell, the PR adds the SCYS adapter page and the adapters index entry, but not the other repo-level updates that usually ship with a new adapter, namely the VitePress sidebar and the command tables in `README.md` and `README.zh-CN.md`. I think it would be worth adding those before merge so the new adapter is discoverable from the main docs and README surfaces.

PR #958, opened 2026-04-11
feat(twitter): add lists command to retrieve user lists

isanwenyu:
## Summary
- Add `twitter/lists` command to retrieve Twitter/X lists for a user
- Supports fetching lists with member/follower counts and privacy mode

## Usage
```bash
opencli twitter lists                # Get your lists
opencli twitter lists elonmusk     # Get another user's lists
opencli twitter lists sundyme      # Get lists for specified user
```

## Output columns
- `name`: List name
- `members`: Number of members
- `followers`: Number of followers  
- `mode`: "public" or "private"

## Implementation
Uses DOM text extraction since the lists page doesn't expose a GraphQL API via network interception. The adapter parses the rendered page text to extract list information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ To load the source Browser Bridge extension:
 | **bilibili** | `hot` `search` `history` `feed` `ranking` `download` `comments` `dynamic` `favorite` `following` `me` `subtitle` `user-videos` |
 | **tieba** | `hot` `posts` `search` `read` |
 | **hupu** | `hot` `search` `detail` `mentions` `reply` `like` `unlike` |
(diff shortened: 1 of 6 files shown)


Astro-Han:
Thanks for adding this. I think the command is useful, but I’m not fully comfortable with the current extraction approach yet.

The biggest concern is that the parser seems tied to Chinese UI text such as `位成员`, `位关注者`, `锁定列表`, and `你的`. If someone is using X in English or another locale, I think this would likely return `Could not parse list data` even though the page loaded normally.

I also think `--limit` may be misleading right now. From what I can tell, the command reads the page text once after the initial render and does not scroll or paginate, so larger list sets would get truncated to whatever happened to be visible at first load. I’d be more comfortable once the extraction is locale-independent and can actually load enough entries to satisfy `--limit`.

PR #963, opened 2026-04-11
fix(zsxq): accept topic_id as string in getTopicFromResponse

fgksgf:
## Bug

`opencli zsxq topic <id>` returns "⚠️" (NOT_FOUND) for all valid topic IDs because the ZSXQ API returns `topic_id` as a string, but `getTopicFromResponse()` only checked for `typeof === 'number'`.

## Fix

Extend the type guard in `getTopicFromResponse()` to also accept strings:

```diff
- if (typeof data.topic_id === 'number')
+ if (typeof data.topic_id === 'number' || typeof data.topic_id === 'string')
```

This allows the function to correctly recognize topics when the API returns `topic_id` as a string (e.g., `"22255514214444820"`).

## Testing

Confirmed the bug exists in v1.7.0. Fix compiled and tested locally against the actual API response structure.


Astro-Han:
This looks good to me. The change is small, the bug report matches the diff, and I did not spot a regression risk from this patch.

PR #968, opened 2026-04-11
feat(doubao): add --thread support for ask send read

hanxuanliang:
## Description

Add optional `--thread <id|url>` support to `opencli doubao ask`, `send`, and `read` so they can target a specific existing Doubao conversation instead of relying on the currently reused browser tab.

This addresses the real workflow issue where the current Doubao tab can disappear and OpenCLI may reopen a fresh tab, which otherwise starts a new conversation and loses the prior context.

This version also fixes the validation concern raised on the previous PR:
- malformed or partial `--thread` values now fail fast with `INVALID_INPUT`
- thread navigation now compares the exact `/chat/<id>` path instead of using a substring match

Included:
- focused adapter tests for `ask`, `send`, and `read`
- shared utility tests for strict thread parsing and exact navigation matching
- updated Doubao adapter docs with `--thread` examples
- updated manifest entries for the three affected commands only

Supersedes the earlier closed PR: #828

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [x] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

### Verification

```bash
npm run test:adapter -- clis/doubao/*.test.js

Test Files  6 passed (6)
Tests      17 passed (17)
```

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -4224,6 +4224,12 @@
         "help": "Prompt to send"
       },
       {
+        "name": "thread",
+        "type": "str",
+        "required": false,
+        "help": "Conversation ID (numeric or full URL)"
+      },
+      {
         "name": "timeout",
         "type": "str",
         "default": "60",
@@ -4381,7 +4387,14 @@
     "domain": "www.doubao.com",
     "strategy": "cookie",
     "browser": true,
-    "args": [],
+    "args": [
+      {
+        "name": "thread",
+        "type": "str",
+        "required": false,
+        "help": "Conversation ID (numeric or full URL)"
+      }
+    ],
     "columns": [
       "Role",
       "Text"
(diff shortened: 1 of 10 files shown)


Astro-Han:
Nice improvement overall. I think there may still be one small CLI parsing edge case worth a look.

`--thread` is added in `clis/doubao/ask.js`, `clis/doubao/send.js`, and `clis/doubao/read.js`, but I do not see `valueRequired: true` on those args. If I am reading `src/commanderAdapter.ts` correctly, that means Commander will register the option as `--thread [value]` rather than `--thread <value>`.

That leaves a case where `opencli doubao ... --thread` with no value may be accepted by the CLI and then fall back to the current tab, instead of failing during argument parsing. If that reading is right, making `--thread` require a value, plus one test that goes through the actual CLI parsing path, should close the gap.

PR #974, opened 2026-04-12
feat(bilibili): add feed-detail and enhance feed command

harveyyue:
## Summary
- Add new `feed-detail` command to view dynamic details (supports exclusive content)
- Enhance `feed` command to support viewing specific user's dynamics via uid parameter
- Add support for multiple dynamic types: video, article, draw, text, forward
- Add pagination support with `pages` parameter


Astro-Han:
Nice expansion of the bilibili adapter. I think the new commands are useful.

A couple of follow-ups, neither feels blocking to me:

`feed-detail` expects a dynamic id, but `feed` currently rewrites some rows to the target `jump_url` for videos and articles. That makes the documented `feed -> feed-detail` path hard to use for common item types, because the original `https://t.bilibili.com/<id>` link is no longer in the output. I think it would be safer to keep the dynamic URL as the main `url`, or add a separate `dynamic_id` / `dynamic_url` field.

Also, this PR includes the `ke` docs and index entry, which looks unrelated to the bilibili change. I would probably split that into a separate PR so this one stays focused.

One more small doc sync item: since `feed-detail` is user-facing, `README.md` and `README.zh-CN.md` should probably list it too.

PR #984, opened 2026-04-13
feat(extension): add open-user-tab support

sangnt1552314:
## Description
Adds support for opening a new tab in an existing Chrome window instead of only using the dedicated automation window.

## Changes
- Add `open-user-tab` operation to the extension protocol
- Implement extension handling to find a non-automation Chrome window and open a tab there
- Add `openUserTab()` to the browser page interface and implementation
- Add `open-user-tab` as a pipeline step
- Mark the step as browser-only for capability routing

## Why
Some commands should open content for the user in their normal Chrome window without being closed by the automation window lifecycle.

## Testing
- Reload the browser extension after building
- Run a command that uses `open-user-tab`
- Verify the URL opens as a new tab in an existing Chrome window
- Verify the automation window behavior remains unchanged for normal browser actions

### DIFF
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -595,6 +595,20 @@ async function handleTabs(cmd: Command, workspace: string): Promise<Result> {
       await chrome.tabs.update(target.id, { active: true });
       return { id: cmd.id, ok: true, data: { selected: target.id } };
     }
+    case 'open-user-tab': {
+      if (!cmd.url || !isSafeNavigationUrl(cmd.url)) {
+        return { id: cmd.id, ok: false, error: 'Missing or invalid URL' };
+      }
+      // Find a user window that is NOT an automation window
+      const automationWindowIds = new Set([...automationSessions.values()].map(s => s.windowId));
+      const allWindows = await chrome.windows.getAll({ windowTypes: ['normal'] });
+      const userWindow = allWindows.find(w => w.id !== undefined && !automationWindowIds.has(w.id));
+      if (!userWindow?.id) {
+        return { id: cmd.id, ok: false, error: 'No existing Chrome window found. Open a Chrome window first.' };
+      }
+      const tab = await chrome.tabs.create({ windowId: userWindow.id, url: cmd.url, active: true });
+      return { id: cmd.id, ok: true, data: { tabId: tab.id, windowId: userWindow.id } };
+    }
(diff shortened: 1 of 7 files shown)


jackwener:
Thanks for the contribution. I reviewed the design against the current browser architecture, and I do not think we should merge this PR as-is.

The underlying need makes sense: sometimes a command wants to hand a URL back to the user instead of keeping it inside OpenCLI's automation window. But I think `open-user-tab` as a pipeline/browser primitive is the wrong boundary for that.

Our current model separates browser ownership explicitly:

- OpenCLI-owned automation windows/tabs are for extraction and automation. OpenCLI can navigate them, close them, and manage their lifecycle.
- User-owned Chrome tabs/windows should only be touched when the user explicitly opts into that ownership boundary.

For pure information extraction, OpenCLI does not need to open anything in the user's normal Chrome window. It should keep using the automation window/session and return structured output.

For cases where the user already has a page open and wants OpenCLI to operate on that page, the current `opencli browser bind` flow is the clearer model: the user explicitly binds a `bound:*` workspace to an existing Chrome tab, and follow-up browser commands operate on that bound page with stricter navigation/tab-mutation safeguards.

Putting `open-user-tab` into the pipeline lets an adapter create visible user-window side effects implicitly. That is a weaker ownership model, and it also makes the pipeline step less natural: pipeline steps should primarily transform/fetch data, not unexpectedly open user-facing browser tabs.

There are also a few implementation issues that would need fixing even if we kept this direction:

- `page!.openUserTab?.(url)` can silently do nothing if the page implementation does not support the method.
- The new step is not added to `src/validate.ts`'s known step list, so `opencli validate` would warn on it.
- There are no tests for the extension handler, pipeline step, or routing.
- No built-in adapter uses this step, so the real usage shape is not demonstrated.

If we want this product capability later, I would prefer a more explicit user-triggered design, for example a command-level `--open` / post-action that opens a returned URL only when the user asked for that side effect. For browser-controlled user tabs, `browser bind` should remain the primary path.

Given that, I recommend closing this PR rather than extending the pipeline with `open-user-tab`.

sangnt1552314:
Hi jackwener,

Thank you for your detailed comment and for pointing out all the possible issues that could happen. This helps me better understand OpenCLI and its development direction. I am closing the PR now.

PR #1015, opened 2026-04-14
feat(extension): make WINDOW_IDLE_TIMEOUT configurable via OPENCLI_WINDOW_IDLE_TIMEOUT env var

### LINKED ISSUE #1014: [Feature]: Make extension WINDOW_IDLE_TIMEOUT configurable via daemon
## Problem

The Browser Bridge extension has a hardcoded `WINDOW_IDLE_TIMEOUT = 30000` (30 seconds) in `extension/src/background.ts`. When no commands arrive within 30 seconds, the automation window is automatically closed via `chrome.windows.remove()`.

This causes a critical usability issue for **`opencli browser` subcommands** (open, eval, click, type, etc.) used by AI agents:

1. AI agents typically need 30-60+ seconds between steps to analyze page state, generate scripts, and plan next actions
2. The 30s idle timeout closes the window before the agent can send the next command
3. Unlike adapter commands (which call `closeWindow()` explicitly per #915), `opencli browser` subcommands rely entirely on this idle timer for window lifecycle

**This is especially problematic because the CLI-side timeout (`OPENCLI_BROWSER_COMMAND_TIMEOUT`) defaults to 60s, which is already longer than the extension-side idle timeout of 30s** — a clear inconsistency.

Related: #638 (Chrome window closes after daemon idle timeout)

## Environment

- OS: macOS
- OpenCLI: v1.7.3
- Extension: v1.0.0
- Node.js: v22.22.0

## Steps to Reproduce

1. Run `opencli browser open "https://example.com"`
2. Wait 30+ seconds without running another command
3. The automation window closes automatically

## Expected Behavior

The idle timeout should be configurable, and the default should be much longer (e.g., 10 minutes) to support multi-step AI agent workflows.

## Proposed Solution

Make the extension read the idle timeout from the daemon via the existing WebSocket channel:

1. **Daemon side**: Read `OPENCLI_WINDOW_IDLE_TIMEOUT` env var (default: 600 seconds)
2. **Protocol**: Include `windowIdleTimeout` in the per-command payload (similar to how `windowFocused` is already passed)
3. **Extension side**: Use the received timeout value, falling back to a sensible default (600s)

This approach:
- Is fully backward-compatible (extension falls back to default if field is missing)
- Follows the existing pattern established by `OPENCLI_WINDOW_FOCUSED` in #915
- Lets users configure via environment variable without modifying extension source code
- Keeps the extension stateless (timeout is refreshed per-command from daemon)

## Alternative (Simpler)

If the per-command approach is too complex, simply increasing the hardcoded default from 30s to 600s in the extension would solve 95% of use cases. The `close-window` action and adapter auto-close (#915) already handle cleanup for short-lived commands.

xzygis:
## Summary

Make the extension's automation window idle timeout configurable via the `OPENCLI_WINDOW_IDLE_TIMEOUT` environment variable, following the same per-command relay pattern used by `OPENCLI_WINDOW_FOCUSED` (#915).

Closes #1014

## Problem

The extension's `WINDOW_IDLE_TIMEOUT` is hardcoded to 30 seconds. When `opencli browser` subcommands are used by AI agents, the automation window closes between commands because agents typically need 30-60+ seconds to analyze state and plan the next action.

Unlike adapter commands (which explicitly call `closeWindow()` per #915), `opencli browser` subcommands rely entirely on the idle timer for window lifecycle — making the 30s default too aggressive.

## Changes

**3 files, +15/-6 lines** — minimal and focused:

### `extension/src/protocol.ts`
- Add `windowIdleTimeout?: number` to `Command` interface

### `src/browser/daemon-client.ts`
- Add `windowIdleTimeout?: number` to `DaemonCommand` interface
- Read `OPENCLI_WINDOW_IDLE_TIMEOUT` env var (seconds, min 5), convert to ms, attach per-command

### `extension/src/background.ts`
- Replace hardcoded `WINDOW_IDLE_TIMEOUT = 30000` with `DEFAULT_WINDOW_IDLE_TIMEOUT = 600_000` (10 min)
- Use mutable `windowIdleTimeout` variable updated from per-command payload
- All idle timer code uses the variable instead of the constant

## Design Decisions

1. **Per-command relay** (not stored in `chrome.storage`): Follows the `windowFocused` pattern from #915. Each command carries the timeout value, keeping the extension stateless.

2. **Default raised to 10 minutes**: Adapter commands already close windows immediately (#915), so the idle timer only matters for `opencli browser` sessions. 10 minutes gives AI agents ample time between steps.

3. **Minimum 5 seconds**: Prevents misconfiguration from making windows uncloseable.

4. **Fully backward-compatible**: If `windowIdleTimeout` is absent from the command payload, the extension uses the new 10-minute default. Old CLI versions work unchanged.

## Usage

```bash
# Set idle timeout to 5 minutes (300 seconds)
export OPENCLI_WINDOW_IDLE_TIMEOUT=300
opencli browser open "https://example.com"
# Window stays alive for 5 min between commands

# Default (no env var): 10 minutes
opencli browser open "https://example.com"
```

## Testing

Tested locally on macOS with Chrome + Browser Bridge Extension v1.0.0:
- ✅ Without env var: window survives 60s+ idle (was closing at 30s)
- ✅ With `OPENCLI_WINDOW_IDLE_TIMEOUT=10`: window closes after ~10s idle
- ✅ `opencli browser close` still works for explicit cleanup
- ✅ Adapter commands still auto-close immediately (unchanged behavior)

### DIFF
--- a/src/browser/daemon-client.ts
+++ b/src/browser/daemon-client.ts
@@ -143,7 +143,9 @@ async function sendCommandRaw(
     const id = generateId();
     const wf = process.env.OPENCLI_WINDOW_FOCUSED;
     const windowFocused = (wf === '1' || wf === 'true') ? true : undefined;
-    const command: DaemonCommand = { id, action, ...params, ...(windowFocused && { windowFocused }) };
+    const wit = process.env.OPENCLI_WINDOW_IDLE_TIMEOUT;
+    const idleTimeout = wit ? Math.max(parseInt(wit, 10), 5) : undefined;
+    const command: DaemonCommand = { id, action, ...params, ...(windowFocused && { windowFocused }), ...(idleTimeout && { idleTimeout }) };
     try {
       const res = await requestDaemon('/command', {
         method: 'POST',


xzygis:
Closing this PR — upstream has since implemented a comprehensive per-workspace idle timeout system (IDLE_TIMEOUT_DEFAULT/IDLE_TIMEOUT_INTERACTIVE + workspaceTimeoutOverrides + per-command idleTimeout field), which fully addresses the original problem of AI agents being disconnected within 30s. The remaining env var convenience is not worth the maintenance overhead. Thanks!

PR #1051, opened 2026-04-16
fix(browser): reuse last network snapshot for browser network --detail

freemandealer:
## Summary
- cache the last `browser network` snapshot and reuse it for `browser network --detail <index>`
- add coverage for cached detail lookup and out-of-range indexes

## Problem
`browser network --detail <index>` could consume a fresh capture batch instead of the list the user had just seen, so the selected index could immediately stop matching the request they meant to inspect.

## Validation
- `npx vitest run --project unit src/cli.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -1,4 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
 import * as path from 'node:path';
 import type { IPage } from './types.js';
 
@@ -13,6 +15,9 @@ const {
   mockRenderCascadeResult,
   mockGetBrowserFactory,
   mockBrowserSession,
+  mockBrowserConnect,
+  mockBrowserClose,
+  browserState,
 } = vi.hoisted(() => ({
   mockExploreUrl: vi.fn(),
   mockRenderExploreSummary: vi.fn(),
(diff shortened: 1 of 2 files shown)


jackwener:
Thanks @freemandealer for spotting this — the index-mismatch symptom is real and I hit it myself this week while building a Twitter adapter. The cache approach here correctly identifies that `--detail` needs to refer back to what the user *just saw*, not a fresh capture batch.

We've decided to fold this into a larger refactor of `browser network` that addresses several related issues together:

1. **Stable request references** — `--detail <key>` using URL path + operation name instead of array index, so references survive across `network` re-runs
2. **Body shape output** in the default list view (JSON path tree + types), so agents can identify the target request without needing `--detail` round-trips
3. **Remove arbitrary truncations** on URL (80 char), body preview (60 char), and `get html` (50000 char) that hurt agent consumers
4. **TTL / invalidation** for the cache so stale snapshots from previous sessions don't leak
5. **Per-tab workspace keys** for future multi-tab support

The new PR will be co-authored with you (using the commit email from this branch) since you identified the root problem and the cache-to-disk direction is correct. Closing this one in favor of the consolidated version — will link it here when it's up.

jackwener:
Heads up — opened #1100 as the full-featured replacement (per discussion on closing this one). You are credited as co-author on the commit. The new PR keeps your TTL/cache idea, adds stable keys (GraphQL operationName or METHOD host+pathname), and swaps the default listing from full bodies to a compact shape-only preview so agents are not paying body tokens unless they explicitly --detail the key. Thanks for the prototype — it shaped the direction.

PR #1079, opened 2026-04-18
fix(zsxq): separate content field from title, remove title truncation

Pandas886:
## Summary
- Split `getTopicText` to return only title, add `getTopicContent` for body text
- Remove `.slice(0, 120)` that was truncating titles
- `content` field now contains full body text instead of duplicating title

## Problem
- `title` was being truncated to 120 characters
- `content` field was identical to `title` (both used the same `getTopicText` function)

## Solution
- `getTopicText` now returns only `topic.title`
- New `getTopicContent` function extracts body text from `talk.text`, `question.text`, etc.
- `toTopicRow` uses `getTopicContent` for the `content` field

## Testing
Verified with json/yaml/plain/md/csv formats - `content` now shows full text.

### DIFF
--- a/clis/zsxq/dynamics.js
+++ b/clis/zsxq/dynamics.js
@@ -37,7 +37,7 @@ cli({
                 time: d.create_time || topic.create_time || '',
                 group: topic.group?.name || '',
                 author: getTopicAuthor(topic),
-                title: getTopicText(topic).slice(0, 120),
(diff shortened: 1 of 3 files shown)


Pandas886:
Without the full content, truncated titles alone have limited usefulness. This fix ensures users get actual body text when querying topics.

PR #1086, opened 2026-04-19
revert: roll back output truncation/key-value heuristic from #1081

jackwener:
## Summary

Revert `3bbea01` from PR #1081.

## Why

Follow-up discussion after merge concluded that the change took the wrong first-principles direction:
- it treated output design as a table-renderer truncation problem
- it hardcoded single-row => key/value as a global renderer heuristic
- the desired direction is to define presentation by result semantics / command shape first, then render

So this PR removes the truncation + single-row heuristic patch from `src/output.ts` and restores the prior behavior.

## Verification

- `node node_modules/vitest/vitest.mjs run src/output.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/output.test.ts
+++ b/src/output.test.ts
@@ -1,13 +1,8 @@
 import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest';
 import { render } from './output.js';
 
-function stripAnsi(str: string): string {
-  return str.replace(/\u001B\[[0-9;]*m/g, '');
-}
-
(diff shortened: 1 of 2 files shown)


jackwener:
Closing as duplicate of #1085 (merged first, CI CLEAN). Thanks @codex-mini1 for the fast reaction — #1085 was opened 29s earlier by @First-principles-1 per task #59. Only one revert needed.

PR #1088, opened 2026-04-19
feat(deepseek): add DeepSeek browser adapter

### LINKED ISSUE #548: [Feature]: support to access deepseek with browser style
### Feature Description

support to browse deepseek.com with opencli

### Use Case

OpenCli can be used to serve as local llm.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

Benjamin-eecs:
## Summary

Closes #548.

Adds a full DeepSeek (chat.deepseek.com) browser adapter with 5 commands:

| Command | Description |
|---------|-------------|
| `deepseek ask <prompt>` | Send a prompt, supports `--model`, `--think`, `--search` |
| `deepseek new` | Start a new conversation |
| `deepseek status` | Check login state and page availability |
| `deepseek read` | Read current conversation messages |
| `deepseek history` | List conversation history from sidebar |

## Features

- Model selection via `--model instant/expert` (default: instant)
- DeepThink mode via `--think`
- Web search mode via `--search`
- All toggles are synced on every invocation: omitting a flag resets it to default
- `--new` flag to start fresh conversations
- Configurable timeout via `--timeout`
- Fail-fast via `CommandExecutionError` when mode toggle or send fails
- Sidebar auto-expand for `history` command
- Retry logic for CDP context invalidation from SPA router transitions

## Addressed review feedback

- `cli-manifest.json` regenerated with all 5 deepseek entries
- `--model` / `--think` / `--search` now fail-fast with `CommandExecutionError` instead of silently degrading
- Toggle state detection uses `ds-toggle-button--selected` class

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `opencli deepseek ask "hello" --new` returns response
- [x] `opencli deepseek ask "..." --model expert` switches to Expert model
- [x] `opencli deepseek ask "..." --think` enables DeepThink mode
- [x] `opencli deepseek ask "..." --search` enables web search
- [x] All 7 mode combinations tested (instant/expert x think/search)
- [x] Toggle reset verified: `--think` then no flag correctly disables DeepThink
- [x] Model reset verified: `--model expert` then no flag correctly reverts to Instant
- [x] `opencli deepseek status` shows login state
- [x] `opencli deepseek new` starts new conversation
- [x] `opencli deepseek history --limit 5` lists conversations (works with collapsed sidebar)
- [x] `opencli deepseek read` reads current conversation

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -4066,6 +4066,147 @@
     "navigateBefore": true
   },
   {
+    "site": "deepseek",
+    "name": "ask",
+    "description": "Send a prompt to DeepSeek and get the response",
+    "domain": "chat.deepseek.com",
+    "strategy": "cookie",
+    "browser": true,
+    "args": [
+      {
+        "name": "prompt",
+        "type": "str",
+        "required": true,
+        "positional": true,
+        "help": "Prompt to send"
+      },
+      {
+        "name": "timeout",
+        "type": "int",
+        "default": 120,
+        "required": false,
+        "help": "Max seconds to wait for response"
+      },
+      {
+        "name": "new",
+        "type": "boolean",
+        "default": false,
+        "required": false,
+        "help": "Start a new chat before sending"
+      },
+      {
+        "name": "model",
+        "type": "str",
+        "default": "instant",
+        "required": false,
+        "help": "Model to use: instant or expert",
+        "choices": [
+          "instant",
+          "expert"
+        ]
+      },
+      {
+        "name": "think",
+        "type": "boolean",
+        "default": false,
+        "required": false,
+        "help": "Enable DeepThink mode"
+      },
+      {
+        "name": "search",
+        "type": "boolean",
+        "default": false,
+        "required": false,
+        "help": "Enable web search"
+      }
+    ],
+    "columns": [
+      "response"
+    ],
+    "timeout": 180,
+    "type": "js",
+    "modulePath": "deepseek/ask.js",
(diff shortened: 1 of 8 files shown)


jackwener:
Hi @Benjamin-eecs — thanks for this adapter. Review concluded **blocked** at head `6108a6e1`. Two concrete blockers that need a fix pass before we can merge:

### 1. `cli-manifest.json` is missing `deepseek/*` entries
`src/discovery.ts:95-106` prefers the manifest fast path and skips the filesystem scan when the manifest is available. Because no `deepseek` entries are in `HEAD:cli-manifest.json`, the published artifact will not register `deepseek ask` / `new` / `status` / `read` / `history`, and they won't appear in completion either.

**Suggested fix:** regenerate or hand-patch `cli-manifest.json` to include all five commands (same shape as the other `browser: true` adapters).

### 2. `--model` / `--think` / `--search` silently degrade
`selectModel()` / `enableFeature()` return `{ selected: false }` / `{ enabled: false }` when the UI control toggle fails, but `clis/deepseek/ask.js:39-50` ignores that result and continues to submit the prompt. The user sees a normal response and mistakenly believes the mode switched to `expert` / DeepThink / Search. Since the command surface exposes these options, the execution path should tighten the contract.

**Suggested fix:** fail-fast — when the control toggle fails, either error out explicitly or return a clear structured failure so the user doesn't get a silently-wrong-mode reply.

Once you push the fix, `@codex-mini0` will re-review. Appreciate the patience — this adapter is otherwise in good shape.

PR #1101, opened 2026-04-20
fix(gemini+grok): restore ask on current UIs

Tchaiku:
Two adapters drifted away from their targets' current DOMs. Both
`gemini ask` and `grok ask` return `[NO RESPONSE]` or timeout until
these patches.

## gemini (3 bugs)

Against `gemini.google.com/app` on opencli v1.7.4 + extension v1.0.1.

1. **Turn-selector pollution in zero-state.** The `/app` welcome
   (`Hi <user>, Where should we start?`) renders inside
   `.visible-primary-message` with a `.message-text` span, which matches
   the `[class*="message"]` turn selector. That gave `before.turns =
   [{Role:'Assistant', Text:'Where should we start?'}]`. After sending,
   the first real turn is User, so `hasGeminiTurnPrefix` returned false
   and the whole structured diff path died.
   Fix: require turn elements to live inside `.conversation-container`
   / `chat-history`, and exclude zero-state greeting containers.

2. **`structuredTurnsTrusted` flipped false on zero-state.** With
   `turns.length === 0` and welcome transcript lines present, the
   structured diff was short-circuited as untrusted. Now that (1)
   removes the false turn, this signal no longer carries information —
   trust the structured diff unconditionally.

3. **Sticky `isGenerating` after response completes.** Gemini's
   send-button keeps `aria-label="Stop response"` long after streaming
   ends, so `waitForGeminiResponse` never exits via
   `!isGenerating && stable >= 2`. Fallback: accept `stable >= 4` (~8s
   at 2s polls) as a completion signal regardless of `isGenerating`.

Verified: 5/5 `opencli gemini ask` calls succeed in 13–14s end-to-end on Fast mode.

## grok (1 bug, two symptoms)

Against `grok.com` on the same setup.

- grok.com migrated from `<textarea>` to a `<div contenteditable="true">`
  composer. Old adapter only queried `textarea`, missing the real
  composer. `document.execCommand('insertText')` on the wrong element
  produced stale partial content, leaving Submit disabled.
- Current Submit button uses `aria-label="Submit"`; adapter only looked
  for `aria-label="提交"` first.

Fix:
- Prefer `div[contenteditable="true"]`, fall back to `textarea`.
- Use `page.nativeType` (CDP `Input.insertText`) so React's input
  listener fires and re-enables Submit.
- Check `aria-label="Submit"` before `aria-label="提交"`.

Verified: 3/3 `opencli grok ask` calls succeed in 12–13s end-to-end.
Before: 3/3 timed out at 53s.

## Tests

```
$ npm test -- clis/gemini/   # 5 files, 64 tests pass
$ npm test -- clis/grok/     # 2 files, 19 tests pass
```

### DIFF
--- a/clis/gemini/utils.js
+++ b/clis/gemini/utils.js
@@ -299,7 +299,11 @@ function readGeminiSnapshotScript() {
         transcriptLines,
         composerHasText: composerText.length > 0,
         isGenerating,
-        structuredTurnsTrusted: turns.length > 0 || transcriptLines.length === 0,
+        // After filtering out zero-state greetings, empty turns means
+        // "no actual conversation yet" — still trust the structured diff.
+        // Previously this flipped to false on the /app root because the
+        // welcome text produced transcriptLines > 0.
+        structuredTurnsTrusted: true,
       };
     })()
   `;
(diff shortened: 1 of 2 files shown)


jackwener:
**Status: Superseded — closing**

Since this PR was opened (2026-04-20), the gemini and grok adapters were substantially rewritten:

- **grok** — full baseline rebuild in PRs #1392, #1397 (typed errors), #707cebd0 (Enter-key fallback), #5a82ecfd (export), #a1555e8f (conversation management)
- **gemini** — the bug patterns this PR targeted no longer exist in current code:
  - `visible-primary-message` / `Where should we start` zero-state selectors → no longer in `clis/gemini/*.js`
  - `structuredTurnsTrusted` is now normally `true` across all tests; the "flipped false on zero-state" path is gone
  - `isGenerating` sticky logic was rewritten

The specific bugs you identified are fixed, just via different code paths. The PR is also CONFLICTING with current main.

Closing as superseded. Thanks for the original deep-dive analysis — it's still useful as a historical record of what the failure looked like. If `gemini ask` or `grok ask` are still broken on current main + extension v1.0.18+, please file a fresh issue with the current `-v` trace.

PR #1122, opened 2026-04-21
feat(cli): add --live and --focus flags for automation window lifecycle

jackwener:
## Summary
- `--live` / `OPENCLI_LIVE=1` keeps the automation window open after an adapter command finishes (default: close immediately as before)
- `--focus` / `OPENCLI_WINDOW_FOCUSED=1` already existed as env var; now also a CLI flag so users don't have to shell-export
- Both parsed early and stripped from argv → work on any subcommand (adapter or browser), in any position

## Why
@WAWQAQ asked for a way to stop commands from auto-closing, so the page stays visible for inspection or follow-up commands. `--focus` was also brought up and turned out to already exist as an env var — exposing it as a flag makes it actually discoverable.

## Test plan
- [x] New `execution.test.ts` cases: `OPENCLI_LIVE=1` skips `closeWindow()` on both success and failure paths
- [x] Existing `calls closeWindow on browser command failure` case still passes (default behavior unchanged)
- [x] `npx vitest run src/execution.test.ts src/commanderAdapter.test.ts src/cli.test.ts` — 99 passed
- [x] `npm run build` green
- [x] Smoke: `opencli --live list` and `opencli list --focus` both parse correctly

## Not in scope
- No auto-teardown timer for the lingering window — the extension's existing 30s idle timeout still applies if the user doesn't explicitly close. Can revisit if agents end up leaking windows across long sessions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,8 @@ OpenCLI is not only for websites. It can also:
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `OPENCLI_DAEMON_PORT` | `19825` | HTTP port for the daemon-extension bridge |
-| `OPENCLI_WINDOW_FOCUSED` | `false` | Set to `1` to open automation windows in the foreground (useful for debugging) |
+| `OPENCLI_WINDOW_FOCUSED` | `false` | Set to `1` to open automation windows in the foreground (useful for debugging). The `--focus` flag sets this. |
+| `OPENCLI_LIVE` | `false` | Set to `1` to keep the automation window open after an adapter command finishes (useful for inspection). The `--live` flag sets this. |
 | `OPENCLI_BROWSER_CONNECT_TIMEOUT` | `30` | Seconds to wait for browser connection |
 | `OPENCLI_BROWSER_COMMAND_TIMEOUT` | `60` | Seconds to wait for a single browser command |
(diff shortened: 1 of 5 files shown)


Henry-G-H-Huang:
I checked PR #1122 more closely, and I think the keep-alive fix may only apply to adapter execution, not to `opencli browser ...` built-in subcommands.
From the diff in PR #1122:
- `src/main.ts` only parses `--live` / `--focus` and converts them into env vars
- `src/execution.ts` is the only place that actually consumes `OPENCLI_LIVE`
- the change there only skips `page.closeWindow()` after command execution
However, PR #1122 does not seem to modify the built-in browser command path at all:
- `src/cli.ts`
- `src/browser/page.ts`
- `src/browser/bridge.ts`
- `src/browser/daemon-client.ts`
So unless I’m missing something, `--live` appears to affect adapter cleanup, but not `opencli browser open/state/click/tab list/...`.
That seems consistent with what I’m observing in practice:
- `opencli browser --live ...` is accepted
- but multi-step browser flows can still drift back to `about:blank` / lose the active target
- so the browser command group itself still does not seem to have true keep-alive semantics
Could you clarify whether PR #1122 was intended to cover `opencli browser ...` subcommands too?
If yes, could you point to the code path where `OPENCLI_LIVE` is consumed for browser subcommands?
If not, then I think #924 may only be partially resolved: adapter commands are covered, but the browser command group is not.

PR #1142, opened 2026-04-22
fix(deepseek): separate thinking process from response in --think mode

### LINKED ISSUE #1124: [Feature]: deepseek ask --think should separate thinking process from final response
### Feature Description

When `deepseek ask --think` is used, the response mixes chain-of-thought reasoning with the final answer in a single `response` string:

```json
[
  {
    "response": "已思考（用时 1 秒）\n\nWe need to answer \"what is 2+2? just the number\" so answer is 4.\n\n4"
  }
]
```

The thinking process and final answer should be structured separately for easy parsing.

### Use Case

opencli is designed to be consumed by agents and scripts. A mixed string forces downstream consumers to regex-parse thinking vs answer, which is fragile across languages (English: "Thought for 1 second", Chinese: "已思考（用时 1 秒）") and subject to break when DeepSeek changes formatting.

Structured output makes the response directly usable without post-processing.

### Proposed Solution

Open for discussion. One option is adding extra fields when `--think` is enabled:

```json
[
  {
    "response": "4",
    "thinking": "We need to answer \"what is 2+2? just the number\" so answer is 4.",
    "thinking_time": "1s"
  }
]
```

This keeps `response` as a clean string (backward compatible), with optional `thinking` and `thinking_time` fields. Agents that don't care about CoT just read `response` as before.

Other approaches welcome.

### Alternatives Considered

1. Embedding a structured object inside `response` (e.g. `response: { answer, thinking }`), but this changes the type from string to object and breaks existing consumers.

2. Encoding the structured data as a JSON string inside `response` (e.g. `response: "{\"answer\":\"4\",\"thinking\":\"...\"}"`) to keep the field type as string. But this makes `response` not human-readable, and consumers need to detect whether `response` is plain text or JSON before parsing, adding unnecessary complexity.

kagura-agent:
## Summary

Fixes #1124

When `deepseek ask --think` is used, the thinking process and final answer are now returned as separate structured fields instead of being mixed in a single `response` string.

### `clis/deepseek/utils.js`
- Added `parseThinkingResponse()` function that extracts thinking content, thinking time, and final answer from the raw response text
- Supports both English ("Thought for X seconds") and Chinese ("已思考（用时 X 秒）") thinking header patterns
- Updated `waitForResponse()` to accept optional `parseThinking` parameter

### `clis/deepseek/ask.js`
- Added `thinking` and `thinking_time` to output columns
- Updated both response paths (normal and file-upload) to use structured parsing when `--think` is enabled

## Output Format

When `--think` is enabled:
```json
[{
  "response": "4",
  "thinking": "We need to answer 'what is 2+2? just the number' so answer is 4.",
  "thinking_time": "1"
}]
```

When `--think` is off, `thinking` and `thinking_time` are `null` — backward compatible.

## Testing

- 6 new tests for `parseThinkingResponse()` (English/Chinese patterns, edge cases)
- 3 new tests for `--think` flag behavior in ask command
- All 1889 project tests pass
- `npx tsc --noEmit` clean

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -4409,7 +4409,9 @@
       }
     ],
     "columns": [
-      "response"
+      "response",
+      "thinking",
+      "thinking_time"
     ],
     "timeout": 180,
     "type": "js",
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -4408,11 +4408,6 @@
         "help": "Attach a file (PDF, image, text) with the prompt"
       }
     ],
-    "columns": [
-      "response",
-      "thinking",
-      "thinking_time"
-    ],
     "timeout": 180,
     "type": "js",
     "modulePath": "deepseek/ask.js",
--- a/clis/deepseek/ask.js
+++ b/clis/deepseek/ask.js
@@ -23,7 +23,7 @@ export const askCommand = cli({
         { name: 'search', type: 'boolean', default: false, help: 'Enable web search' },
         { name: 'file', help: 'Attach a file (PDF, image, text) with the prompt' },
     ],
-    columns: ['response', 'thinking', 'thinking_time'],
+    // columns omitted: derived from row keys so non-think output shows only 'response'
 
     func: async (page, kwargs) => {
         const prompt = kwargs.prompt;
--- a/clis/deepseek/ask.js
+++ b/clis/deepseek/ask.js
@@ -23,7 +23,7 @@ export const askCommand = cli({
         { name: 'search', type: 'boolean', default: false, help: 'Enable web search' },
         { name: 'file', help: 'Attach a file (PDF, image, text) with the prompt' },
     ],
-    columns: ['response'],
+    columns: ['response', 'thinking', 'thinking_time'],
 
     func: async (page, kwargs) => {
         const prompt = kwargs.prompt;
@@ -71,11 +71,14 @@ export const askCommand = cli({
                 if (!String(err?.message || err).includes('Promise was collected')) throw err;
             }
             await page.wait(3);
-            const response = await waitForResponse(page, baseline, prompt, timeoutMs);
-            if (!response) {
+            const result = await waitForResponse(page, baseline, prompt, timeoutMs, wantThink);
+            if (!result) {
                 return [{ response: `[NO RESPONSE] No reply within ${kwargs.timeout}s.` }];
             }
-            return [{ response }];
+            if (wantThink && typeof result === 'object' && result.response !== undefined) {
+                return [result];
+            }
+            return [{ response: result }];
         }
 
         const baseline = await withRetry(() => getBubbleCount(page));
@@ -84,11 +87,14 @@ export const askCommand = cli({
             throw new CommandExecutionError(sendResult?.reason || 'Failed to send message');
         }
 
-        const response = await waitForResponse(page, baseline, prompt, timeoutMs);
-        if (!response) {
+        const result = await waitForResponse(page, baseline, prompt, timeoutMs, wantThink);
+        if (!result) {
             return [{ response: `[NO RESPONSE] No reply within ${kwargs.timeout}s.` }];
         }
 
-        return [{ response }];
+        if (wantThink && typeof result === 'object' && result.response !== undefined) {
+            return [result];
+        }
+        return [{ response: result }];
     },
 });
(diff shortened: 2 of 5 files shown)


jackwener:
Hi @kagura-agent — thanks for the PR. pr-monitor reviewers (B 组) ran two independent reviews on head `555924a` and both blocked. Consolidated blockers below.

## 🛑 Blocker 1 — silent-corruption in `parseThinkingResponse()` (`clis/deepseek/utils.js:114–129`)

The current heuristic — `split(/\n\n+/)` and treat the last segment as the final `response` — misroutes data whenever the streamed text contains multiple paragraphs but **hasn't reached the final answer yet** (which is exactly what happens during mid-stream polling in `waitForResponse`).

Minimal reproduction (run in `clis/deepseek/utils.test.js`):

```js
parseThinkingResponse('Thought for 1.2 seconds\n\nFirst paragraph.\n\nSecond paragraph.')
// Current output:
// { thinking: 'First paragraph.', response: 'Second paragraph.', thinking_time: '1.2' }
// Expected: both paragraphs are still thinking; response should be '' or buffered
```

Also reproducible when the **final answer itself has multiple paragraphs**:
```js
parseThinkingResponse('Thought for 3 seconds\n\nreasoning\n\nAnswer para 1.\n\nAnswer para 2.')
// Current output splits Answer para 1 into thinking, Answer para 2 becomes response.
```

Because `ask.js:78` / `ask.js:95` directly return `[result]` when the parsed object has `response !== undefined`, the CLI will ship the misrouted segment as the canonical answer — silent data corruption, not a display glitch.

**Suggested directions (pick one that fits your knowledge of DeepSeek's UI):**
- **A. DOM-level separation (preferred):** DeepSeek's UI renders thinking and final-answer in distinct DOM nodes. Instead of splitting scraped text, walk the bubble's DOM in `waitForResponse` (or a new helper) and read the two regions separately. This removes the heuristic entirely.
- **B. Stronger boundary marker:** If DeepSeek inserts a specific separator (horizontal rule / role marker / CSS boundary) between CoT and final answer, use that as the split point rather than `\n\n`.
- **C. Stream-safe fallback:** If mid-stream separation cannot be reliably detected, only return a structured row **after** the response is stable AND a clear final-answer region exists; otherwise keep returning plain `{ response }` with the raw text (no thinking field populated). This favors BC-safety over feature completeness.

## ⚠️ Blocker 2 — `columns` regression on non-`--think` path (`clis/deepseek/ask.js:26`)

`columns` is now statically `['response', 'thinking', 'thinking_time']`. Because `src/commanderAdapter.ts:102–105` forwards `resolved.columns` to the renderer unconditionally and `src/output.ts:26,63–67` emits a cell per declared column, every `opencli deepseek ask ...` invocation (including the existing non-`--think` usage) now renders table/csv/md output with two empty trailing columns. That's a breaking change to the existing command contract.

**Suggested fix:** make `columns` a function of `kwargs.think` (or conditionally set a narrower column list in the non-`--think` return path). The JSON renderer is already fine since it only serializes the actual row keys.

## 🧪 Test gaps

Please add coverage for:
- Multi-paragraph thinking **without** a final answer (the stream-mid reproduction above).
- Multi-paragraph final answer (ensures whichever fix you pick keeps the full answer intact).
- Non-`--think` output: assert that table/csv rendering does NOT add `thinking` / `thinking_time` columns (regression guard for Blocker 2).

## Next steps

Once you push the rework, ping this thread — reviewers will do an incremental pass on the new head and pr-monitor will merge when both reviewers are `green` + CI clean.

Reviewers sign-off (both blocked on `555924a`): codex-mini1 (lead), First-principles-1 (aux).

kagura-agent:
Thanks for the thorough review! Both blockers addressed in `c79fe91`:

**Blocker 1 — silent-corruption fix:**
- Replaced the fragile `split(/\\n\\n+/)` heuristic with **DOM-level separation** (Option A). `waitForResponse()` now queries distinct DOM nodes (`.ds-markdown--think` for thinking, `.ds-markdown` for the final answer) directly from the page, avoiding any text-based split.
- The text-level `parseThinkingResponse()` fallback now treats everything after the header as thinking content (no split), so mid-stream or multi-paragraph content is never misrouted.
- Added tests for: multi-paragraph thinking without final answer, multi-paragraph final answer.

**Blocker 2 — columns regression fix:**
- Removed the static `columns: ['response', 'thinking', 'thinking_time']` from `askCommand`. The renderer now infers columns from row keys: non-think output only has `response`, think output has all three fields.
- Added test asserting non-think rows only contain the `response` key.

All 16 deepseek tests pass. Ready for re-review.

PR #1144, opened 2026-04-22
fix(browser): fall back to legacy networkRequests() when CDP and JS capture return empty

Benjamin-eecs:
Closed. Wrong approach.


Benjamin-eecs:
Closing this. The legacy networkRequests() fallback uses performance.getEntriesByType which lacks response body and content-type, so it can't serve as a proper fallback for the new capture system. The root cause seems to be that CDP readNetworkCapture() returns empty despite startNetworkCapture() succeeding, which may be an extension-side issue.

PR #1162, opened 2026-04-23
Fix concurrent command execution

yaokuku123:
## Description
Fix "Detached while handling command" error when running multiple opencli commands concurrently in the same workspace.

Root cause : closeWindow() in page.ts sends close-window without the page identity ( targetId ), so the extension's handleCloseWindow cannot identify which tab to close and falls back to detaching all tabs — breaking other concurrent commands.

Changes :

- src/browser/page.ts : Use _cmdOpts() instead of _wsOpt() in closeWindow() so the page identity is included
- extension/src/background.ts :
  - Add withWorkspaceLock to prevent race condition in automation window creation
  - Modify resolveTab to create a new tab per command when no specific page is requested
  - Modify handleCloseWindow to close only the tab identified by cmd.page instead of detaching all tabs
Related issue:

## Type of Change
- 🐛 Bug fix
- ✨ New feature
- 🌐 New site adapter
- 📝 Documentation
- ♻️ Refactor
- 🔧 CI / build / tooling
## Checklist
- I ran the checks relevant to this PR
- I updated tests or docs if needed
- I included output or screenshots when useful
- Added doc page under docs/adapters/ (if new adapter)
- Updated docs/adapters/index.md table (if new adapter)
- Updated sidebar in docs/.vitepress/config.mts (if new adapter)
- Updated README.md / README.zh-CN.md when command discoverability changed
- Used positional args for the command's primary subject unless a named flag is clearly better
- Normalized expected adapter failures to CliError subclasses instead of raw Error
## Screenshots / Output

E2E tests with concurrency=4:
Before fix 38.5% (Detached errors)  
After fix 100% (matching serial baseline)  Detached errors None

### DIFF
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -137,6 +137,22 @@ const IDLE_TIMEOUT_INTERACTIVE = 600_000; // 10min — human-paced browser:* / o
 /** Per-workspace custom timeout overrides set via command.idleTimeout */
 const workspaceTimeoutOverrides = new Map<string, number>();
 
+/**
+ * Per-workspace mutex for getAutomationWindow().  Without this, concurrent
+ * commands targeting the same workspace can race: both see no session,
+ * both create a new window, and the second overwrites the first — leaving
+ * the first command's tab in an orphaned window that the session no longer
+ * points to, causing "Detached while handling command" errors.
+ */
+const workspaceLocks = new Map<string, Promise<void>>();
+
+function withWorkspaceLock<T>(workspace: string, fn: () => Promise<T>): Promise<T> {
+  const prev = workspaceLocks.get(workspace) ?? Promise.resolve();
+  const next = prev.then(() => fn(), () => fn());
+  workspaceLocks.set(workspace, next.then(() => {}, () => {}));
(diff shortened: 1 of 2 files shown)


jackwener:
Closing as stale/superseded by current main.

I re-checked the actual codepath this PR was trying to fix. Current main already isolates browser adapter runs differently:

- `src/execution.ts` now gives each browser adapter invocation a unique workspace (`site:${cmd.site}:${crypto.randomUUID()}`) instead of reusing one shared site workspace.
- `extension/src/background.ts` closes/releases by workspace lease (`handleCloseWindow()` -> `releaseWorkspaceLease(workspace, ...)`) instead of the older "detach everything in the shared site window" behavior.
- `extension/src/background.test.ts` now has explicit coverage for concurrent execution across workspaces and for closing one workspace lease without tearing down another workspace.

So the underlying concurrent-detach bug is already addressed on main, but through a newer workspace/lease model rather than the exact patch shape in this PR. At this point the branch is outdated relative to the current design, so I am closing it instead of trying to merge it on top of the newer concurrency model.

PR #1180, opened 2026-04-25
fix(twitter): preserve text when posting images

hanzili:
## Summary
- upload images before inserting tweet text so X's composer re-render cannot erase Draft.js content
- use the current `/compose/post` route and explicit composer/file input readiness checks
- insert text with native browser input when available, fall back to DOM events, and verify text before clicking Post
- wait for submission completion instead of reporting success immediately after click
- add adapter regression coverage for image posting order, upload timeout, fallback text insertion, and validation paths

## Test Plan
- [x] `npx vitest run --project adapter clis/twitter/post.test.js clis/twitter/reply.test.js`
- [x] `npm run build`
- [x] `git diff --check`

## Background
While using `opencli twitter post "text" --images image.jpg`, the command could report success while the resulting tweet was missing text/media. X's composer can reset the Draft.js editor while media is attaching, so the command now attaches media first and verifies the final composer state before posting.

### DIFF
--- a/clis/twitter/post.js
+++ b/clis/twitter/post.js
@@ -2,10 +2,18 @@ import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { cli, Strategy } from '@jackwener/opencli/registry';
 import { CommandExecutionError } from '@jackwener/opencli/errors';
+
 const MAX_IMAGES = 4;
 const UPLOAD_POLL_MS = 500;
 const UPLOAD_TIMEOUT_MS = 30_000;
+const COMPOSER_POLL_MS = 250;
+const COMPOSER_TIMEOUT_MS = 10_000;
+const SUBMIT_POLL_MS = 500;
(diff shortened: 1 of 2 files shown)


hanzili:
Live-tested the patched local build against X with an actual image upload.

Result verified on the profile:
- text preserved
- media attached
- URL: https://x.com/hanzi_li/status/2048095554240913879

The live test exposed one extra edge case: X can publish successfully while stale composer/media nodes remain visible long enough for the CLI's submission-complete heuristic to time out. I updated the PR to:
- prefer the Page `nativeType` CDP insertion helper before the older bridge `insert-text` path
- treat X's success toast as completion even when stale composer nodes remain
- add regression coverage for both cases

Updated local verification:
- `npm run build`
- `npx vitest run --project adapter clis/twitter/post.test.js clis/twitter/reply.test.js` → 20/20 passing
- `git diff --check`

Issue #1265, opened 2026-05-03
Feature Request: Support Input.insertText for DraftJS-compatible Text Input


JustToFaith:
### Problem

`browser type` uses `document.execCommand('insertText')` to insert text. However, React DraftJS editors (used by Twitter/X, Notion, and many other web apps) do not capture `execCommand` events.

Text appears in the DOM but the application state never updates, causing automation to fail silently.

### Current Behavior

```
opencli browser type "[data-testid=\"tweetTextarea_0\"]" "Hello"
# Text appears in DOM but DraftJS state is empty
# Submit produces empty tweet
```

### Expected Behavior

Text inserted should update React/DraftJS state, not just the DOM. This requires using CDP's `Input.insertText`.

### Reference

CDP `Input.insertText` correctly triggers React synthetic events: https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-insertText

Benjamin-eecs:
Reproduced on Twitter / X with a real logged-in session.

Before, with `document.execCommand('insertText')`:

```
opencli browser type '[data-testid="tweetTextarea_0"]' "test framework bug"
# DOM text = "test framework bug"
# tweetBtn disabled = true, aria-disabled = "true"
```

The DraftJS internal state never updated, so the Post button stayed disabled and submit produced nothing.

PR #1272 routes `BasePage.typeText` through `page.nativeType` (CDP `Input.insertText`) when the backend supports it, with the legacy `execCommand` path kept as fallback for `CDPPage` and any backend without nativeType. Pattern mirrors the existing `tryNativeClick` escalation.

After the fix, same Twitter composer:

```
tweetBtn disabled = false, aria-disabled = null
```

Post button enabled, confirming the React-controlled state received the input.

Existing browser unit tests pass (275/275, `src/browser/`). No new unit tests added because `typeText` calls `runResolve()` first which requires deep resolver mocking to exercise the new branch in isolation; the live Twitter reproducer above is the practical signal.

PR #1273, opened 2026-05-03
docs(developer): add CDP method policy

jackwener:
## Summary

Codify what CDP methods we expose, what we intentionally don't, and the concrete failure-mode triggers that would change our minds. Triggered by issue #1265 — instead of evaluating CDP methods one at a time, capture the policy so future "should we add X?" requests get filtered through a written framework.

The doc has four parts:
- **Decision framework** — three-question filter (real failure mode? no existing primitive? maintenance value > cost?)
- **Currently exposed surface** — extension allowlist, daemon action table, IPage methods with rationale
- **Intentionally not exposed** — each row paired with a specific re-evaluation trigger (e.g. Fetch domain stays out until an adapter can only be verified by mutating a response)
- **Known parity gaps** — CDPPage trails Browser-Bridge Page on native input methods; closing this is in-scope for whichever PR touches native input next

## Test plan
- [x] `npm run docs:build` — passes
- [x] Sidebar entry visible under `/developer/` section

## Related

- Issue #1265 (CDP Input.insertText for DraftJS) — codex-coder is implementing the `browser type` smart-router fix in a separate PR (task A); this doc captures the policy that filtered the rest of @opencli-质量官's proposed CDP additions down to what's actually justified.
- Task #202 (#OpenCLI:239caba7) — splits as: codex-coder = task A implementation, this PR = policy doc, follow-up task B = JS dialog handling.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -158,6 +158,7 @@ export default defineConfig({
                 { text: 'YAML Adapter Guide (Deprecated)', link: '/developer/yaml-adapter' },
                 { text: 'TypeScript Adapter Guide', link: '/developer/ts-adapter' },
                 { text: 'AI Workflow', link: '/developer/ai-workflow' },
+                { text: 'CDP Method Policy', link: '/developer/cdp-method-policy' },
               ],
             },
           ],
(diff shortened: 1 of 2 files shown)


jackwener:
WAWQAQ pushed back: policy doc is itself over-engineering (preventing hypothetical future debate by writing a 26-row "not exposed" list with no current demand). Closing per first-principles.

PR #1313, opened 2026-05-04
test(dianping): JSDOM-against-frozen-fixture tests for in-browser extractors

jackwener:
## Summary

Follow-up to PR #1309 / #1312. Add CI-time coverage for the dianping in-browser extractors so silent DOM bugs (like the full-width 【】 vs ASCII [] mismatch fixed in #1312) fail in CI rather than only on live verify.

PR #1312 fixed two silent bugs that mocked `page.evaluate` tests could not catch:
1. shop title fallback split on ASCII `[]` while dianping renders full-width `【】`, so `name` was always empty.
2. headText `\s+` collapse fused rating `4.8` with reviews `21241条`, so a head-wide `/\d+条/` regex captured `4.821241` → 5.

Both surfaced only on live verify. Existing tests fed pre-baked results to the func; the real DOM walk never ran.

## Changes

- `clis/dianping/shop.js` / `search.js`: extract the IIFE bodies into top-level `extractShopFields()` / `extractSearchRows()` using bare `document` / `location`. The live adapters inject these via `page.evaluate(\`(${fn.toString()})()\`)`, so behavior is unchanged. Both commands re-verified end-to-end against live dianping post-refactor.
- `clis/dianping/__fixtures__/shop.html` (3.4KB), `search.html` (8.4KB): sanitized HTML snapshots. Scripts/styles/iframes/comments stripped, images placeholdered, only structural attrs kept. Trimmed to minimum subtree needed to exercise the extractors (search keeps 3 of 15 li cards; shop keeps `.shop-head` + `.desc-info` + `.review-title` plus full-width 【】 title and the rating/reviews fusion preserved).
- `clis/dianping/dianping.test.js`: new "extractors against frozen HTML fixtures" describe block. Loads fixtures via JSDOM, swaps `globalThis.document` / `globalThis.location`, asserts post-fix behavior on shop (`name=芈重山老火锅(五道口店)`, `reviewsRaw=21241条`, `rating=4.8`, breakdown, hours, rank, subway), search (3 rows with shop_ids / names / reviewsRaw / priceRaw / starClass + post-process round-trip), and `ok:false` branches.

## Verification

Manually verified the fixtures would catch the original bugs by running buggy extractor variants against `shop.html`:
- ASCII-bracket fallback → `name = "undefined"` (fails new assertion).
- Head-wide `/\d+条/` regex → `reviewsRaw = "821241条"` (fails new assertion).

Test suite grows 14 → 18 passing.

Live verify after refactor:
- `opencli browser verify dianping/shop H20FrTRI6kbXbgTu` → `name=芈重山老火锅(五道口店), rating=4.8, reviews=21241, price=109, ...` ✓
- `opencli browser verify dianping/search 火锅 --city beijing --limit 3` → 3 rows with correct ids ✓

## Scope intentionally limited to dianping

Pattern intentionally not pushed across sites yet. If other in-browser DOM adapters encounter similar silent bugs, this JSDOM-against-frozen-fixture pattern can be adopted per-site.

## Test plan

- [x] `pnpm vitest run clis/dianping/dianping.test.js` — 18/18 passing
- [x] Live verify dianping/shop with refactored extractor
- [x] Live verify dianping/search with refactored extractor
- [x] Confirm fixtures would catch the original PR #1312 bugs

### DIFF
--- a/clis/dianping/__fixtures__/search.html
+++ b/clis/dianping/__fixtures__/search.html
@@ -1,0 +1,372 @@
+<!DOCTYPE html><html class="trancy-und"><head>
+    
+    
+    <title>北京火锅相关搜索结果推荐-大众点评网</title>
+        
+        
+        
+        
+    
+
+    
+
+    
+    
+    
+    
+    
+    
+    
+    
+        
+    
+
+    
+    
+    
+    
+    
+    
+    
+    
+    
+    
+
+
+
+    
+</head>
+<body>
+    
+    
+    <div class="section Fix J-shop-search">
+        
+
+
+                            
+                        
+
+
+
+
+
+
+
+
+
+        
+        <div class="content-wrap">
+            <div class="shop-wrap">
+                <div class="content">
+                        
+                    
+<div class="shop-list J_shop-list shop-all-list" id="shop-all-list">
+  <ul>
+  <li class="">
+    <div class="pic">
+      <a href="https://www.dianping.com/shop/H20FrTRI6kbXbgTu" data-shopid="H20FrTRI6kbXbgTu" title="">
+        <img title="芈重山老火锅(五道口店)" alt="芈重山老火锅(五道口店)" src="placeholder.png">
+      </a>
+    </div>
+
+    <div class="txt">
+      <div class="tit">
+        <a data-shopid="H20FrTRI6kbXbgTu" title="芈重山老火锅(五道口店)" href="https://www.dianping.com/shop/H20FrTRI6kbXbgTu">
+            <h4>芈重山老火锅(五道口店)</h4>
+        </a>
+
+
+        <div class="promo-icon J_promo_icon">
+
+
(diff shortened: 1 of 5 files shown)


jackwener:
Review note (cannot use formal request-changes because this PR is under the same GitHub identity):

Blocking on one real issue: CI is red in `check:silent-column-drop`. The refactor changed the search extractor failure object from `{ ok, bodyLen, sample, url }` to `{ ok, sample, url }`, so the existing baseline entry `missing=[bodyLen, sample]` resolves and a new entry `missing=[sample]` appears. That is not a real adapter fix; it is signature drift from dropping `bodyLen` during extraction refactor.

Please preserve the previous diagnostic shape in the no-results branch, ideally in both search and shop extractors for consistency: compute `bodyLen: sampleText.length` and return it with `sample`/`url`. After that, rerun the silent-column-drop gate against current main.

I checked the two review-focus areas: fixture sizes are acceptable for the regression coverage, and the `globalThis.document/location` swap restores state in `afterEach`. No other blocker from my side.

PR #1333, opened 2026-05-05
fix(web/read): extract same-origin iframes outside contentEl

kevinapi:
## Problem

When `--frames same-origin` is enabled, `buildRenderAwareExtractorJs()` builds `originalFrames` by querying only within `contentEl` (the heuristically selected main content element):

```js
const originalFrames = Array.from(contentEl.querySelectorAll('iframe'));
```

Any same-origin iframe that exists **outside** `contentEl` — for example above the fold, in a header region, or in a sidebar — appears in `allFrames` and is reported in diagnostics (`frames: N, included_same_origin: 0`), but its content is never extracted or merged into the output. The result is silent content loss for pages where meaningful data lives in an iframe that the content-selection heuristic places outside the main element.

## Root cause

`allFrames` was only used to populate `result.diagnostics.frames`. It played no role in content extraction. The `originalFrames` loop, which does the actual extraction, only iterates over iframes inside `contentEl`.

## Fix

After the existing `originalFrames.forEach(...)` loop, add a second pass over `allFrames` that:

1. **Skips** any frame already contained within `contentEl` (to avoid duplication with the first loop).
2. **Filters** to accessible same-origin frames with at least 50 characters of text content — this threshold excludes navigation bars, ad slots, and other low-signal iframes without requiring any site-specific logic.
3. **Appends** the qualifying frame bodies as `<section data-opencli-iframe-source="...">` elements to `clone`, following the same pattern used by the first loop.

```js
// Also extract same-origin iframes that live outside contentEl.
allFrames.forEach((frame, index) => {
  if (contentEl.contains(frame)) return; // already handled above
  const desc = describeFrame(frame, index);
  if (!desc.sameOrigin || !desc.accessible || (desc.textLength || 0) < 50) return;
  try {
    const doc = frame.contentDocument;
    if (!doc?.body) return;
    const frameBody = doc.body.cloneNode(true);
    absolutizeTree(frameBody, desc.src || window.location.href);
    collectEmptyContainers(frameBody, 'iframe', desc.src);
    const section = document.createElement('section');
    section.setAttribute('data-opencli-iframe-source', desc.src);
    const heading = document.createElement('h2');
    heading.textContent = '来自 iframe: ' + (desc.src || frame.getAttribute('src') || ('#' + index));
    section.appendChild(heading);
    Array.from(frameBody.childNodes).forEach(node => section.appendChild(node));
    clone.appendChild(section);
    result.diagnostics.includedFrameCount += 1;
  } catch {}
});
```

## Verification

With this fix, `--diagnose` output changes from `included_same_origin: 0` to `included_same_origin: 1` (or more) for pages that have qualifying iframes outside `contentEl`, and the previously missing content appears in the Markdown output. Pages without such iframes are unaffected.

## Notes

- The 50-character `textLength` threshold is intentionally generic — it filters noise without encoding any site-specific knowledge.
- The fix follows the identical cloning/absolutizing/section-wrapping pattern already used in the `originalFrames` loop, so behavior is consistent.
- No changes to the diagnostics schema or CLI interface.

### DIFF
--- a/clis/web/read.js
+++ b/clis/web/read.js
@@ -277,6 +277,31 @@ function buildRenderAwareExtractorJs(options) {
               result.diagnostics.includedFrameCount += 1;
             } catch {}
           });
+
+          // Also extract same-origin iframes that live outside contentEl.
+          // originalFrames only covers iframes inside the selected content element;
+          // meaningful iframes elsewhere in the document (e.g. above the fold) are
+          // never processed by the loop above, so their content is silently dropped.
+          allFrames.forEach((frame, index) => {
+            if (contentEl.contains(frame)) return; // already handled above
+            const desc = describeFrame(frame, index);
+            if (!desc.sameOrigin || !desc.accessible || (desc.textLength || 0) < 50) return;
+            try {
+              const doc = frame.contentDocument;
+              if (!doc?.body) return;
+              const frameBody = doc.body.cloneNode(true);
+              absolutizeTree(frameBody, desc.src || window.location.href);
+              collectEmptyContainers(frameBody, 'iframe', desc.src);
+              const section = document.createElement('section');
+              section.setAttribute('data-opencli-iframe-source', desc.src);
+              const heading = document.createElement('h2');
+              heading.textContent = '来自 iframe: ' + (desc.src || frame.getAttribute('src') || ('#' + index));
+              section.appendChild(heading);
+              Array.from(frameBody.childNodes).forEach(node => section.appendChild(node));
+              clone.appendChild(section);
+              result.diagnostics.includedFrameCount += 1;
+            } catch {}
+          });
         }
 
         collectEmptyContainers(clone, 'main', window.location.href);


jackwener (changes requested):
Thanks for the PR. The root problem is real: `--frames same-origin` currently reports same-origin iframes outside the selected `contentEl` in diagnostics, but never merges them into output, so content can be silently lost when the main-content heuristic chooses a narrower element.

I don't think this should merge as-is because the new outside-frame path adds an asymmetric `textLength >= 50` gate. From the reader contract, `frames: same-origin` is already the policy boundary; whether an iframe is inside or outside the heuristic `contentEl` should not change whether its same-origin body is eligible for extraction. The 50-character threshold can still silently drop exactly the class of pages this feature is meant to diagnose, especially short data/status tables or empty-but-important data containers where `--diagnose` should surface `#gridDatas` / table/list hints.

Please adjust the second pass to use the same inclusion semantics as the existing in-content iframe path, or replace the text-length check with a structural signal that still includes likely data containers and records diagnostics before skipping. Also add a regression test with a selected `<main>` plus a same-origin iframe outside that main element, asserting the iframe content is merged. A second test for a short/table-like outside iframe would lock down the data-loss case.

Local check: `npm test -- --run clis/web/read.test.js` passes on the PR branch.

jackwener:
Closing as superseded. The root cause in this PR was valid, but the fix has already landed through follow-up PRs with a broader implementation and regression coverage:\n\n- #1371 fixed same-origin iframes outside the selected content element, added shared iframe section building, and added regression tests for outside/relevant iframe inclusion.\n- #1372 cleaned up duplicate iframe diagnostics and consolidated the iframe pass.\n- #1373 added the explicit exhaustive mode `--frames all-same-origin`, updated docs/help to distinguish relevant vs exhaustive iframe merging, and added the short pure-text outside-iframe reproducer test.\n\nSo this PR is no longer needed as a merge candidate. Thanks for identifying the original silent content-loss bug.

PR #1380, opened 2026-05-07
fix:(xiaohongshu) improve image publishing

E2ern1ty:
## Summary

Fixes Xiaohongshu image note publishing in the creator center UI.

## Changes

- Open the creator publish page directly in image mode with target=image.
- Avoid clicking broad parent containers when selecting the 图文 tab by preferring exact tab text first.
- Fall back to the existing DataTransfer upload path when CDP file injection returns Chrome Not allowed.

## Verification

- node --check clis/xiaohongshu/publish.js
- git diff --check
- npm run typecheck
- npm run build-manifest
- Verified locally by publishing a 5-image Xiaohongshu note after the fallback path was used.


E2ern1ty:
@jackwener 你好，这个 PR 修复了小红书图文笔记发布时两个实际问题：默认落到视频上传页，以及 CDP 文件注入返回 Not allowed 时没有走备用上传路径。我已经用真实发布流程验证过，麻烦帮忙看下，谢谢。

PR #1399, opened 2026-05-07
fix(browser): restart stale ready daemon

jackwener:
## Summary\n- check daemon version before the ready fast path in BrowserBridge\n- restart stale daemons even when the extension is already connected\n- add regression coverage for a ready-but-stale daemon\n\n## Test plan\n- npm test -- --run src/browser.test.ts\n- npm run typecheck\n- npm run build\n- git diff --check

### DIFF
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -205,6 +205,27 @@ describe('BrowserBridge state', () => {
 
     await expect(bridge.connect({ timeout: 0.1 })).rejects.toThrow('Stale daemon could not be replaced');
   });
+
+  it('attempts stale daemon replacement even when extension is connected', async () => {
+    vi.spyOn(daemonClient, 'getDaemonHealth').mockResolvedValue({
+      state: 'ready',
+      status: {
+        ok: true,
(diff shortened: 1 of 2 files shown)


jackwener:
LGTM. Ran browser.test.ts locally on this branch (20/20 pass). The relocated stale check correctly gates all fast-path branches with !staleDaemonReplaced and falls through to spawn when the port releases. Behavior change for stale + profile-required / profile-disconnected (now restarts instead of throwing) is desirable.

Nice-to-have (non-blocking): the new test only covers the failure path (shutdown returns false → throws). A happy-path test (shutdown succeeds → port released → fresh daemon spawns) would prove the actual user-visible fix end-to-end, but the existing tests already prove the stale check fires before the ready fast-path return, and the spawn path is shared with the existing no-extension+stale branch — regression risk is low.

Residual smoke gap: real-world `npm install -g @jackwener/opencli@latest` → next `opencli grok status` auto-replacing the daemon hasn't been validated in a live env. Worth a manual smoke before next release.

PR #1401, opened 2026-05-07
feat(help): make adapter help agent-friendly

jackwener:
## Summary\n- remove per-command `[options]` noise from adapter site help command lists\n- render site help with positional usage plus a single Common options block\n- render command help with Arguments / Command options / Common options groups\n- enrich structured help so site-level YAML exposes each command's usage, positionals, command_options, common_options, access, examples, and output metadata\n\n## Test plan\n- npm test -- --run src/help.test.ts src/commanderAdapter.test.ts src/cli.test.ts\n- npm run typecheck\n- npm run build\n- git diff --check\n- empty-HOME smoke: opencli twitter --help, opencli twitter notifications --help, opencli twitter --help -f yaml

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -247,6 +247,7 @@ describe('createProgram root help descriptions', () => {
         strategy: Strategy.PUBLIC,
         browser: false,
         args: [{ name: 'limit', type: 'int', default: 20, help: 'Number of videos' }],
+        columns: ['title', 'url'],
       });
 
       const program = createProgram('', '');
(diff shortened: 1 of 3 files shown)


jackwener:
Re-reviewed at e243845 — LGTM, all three gaps cleanly addressed.

**Verified locally on the updated branch:**
- Gap 1 (hybrid `[options]`): site listing shows `article <tweet-id>` (no opts) vs `notifications [options]` / `accept <query> [options]` / `bookmarks [options]` / `download [username] [options]` — exactly the 1-bit signal we wanted.
- Gap 2 (site YAML completeness): each command inlines `browser`, `domain`, `columns` — agent gets full planning data from one `opencli <site> --help -f yaml` call.
- Gap 3 (no redundant `args`): structured output has `positionals` + `command_options` only; new test `expect(data.commands[0]).not.toHaveProperty('args')` locks the invariant.

148/148 targeted tests, build clean (795 manifest entries unchanged).

Two non-blocking nice-to-haves for a future pass: (a) `positionals[].positional: true` is now permanently redundant once positionals/command_options are partitioned; (b) site YAML `command` field is a prefix of `usage` and could go. Neither blocks merge.

PR #1425, opened 2026-05-09
feat(daemon): add OPENCLI_DAEMON_HOST and OPENCLI_DAEMON_BIND envs

qinglu008:
## Summary
- add `OPENCLI_DAEMON_HOST` so the CLI can reach a daemon that is not on `127.0.0.1` (e.g. when calling out from a container to a daemon on the host)
- add `OPENCLI_DAEMON_BIND` so the daemon can listen on an interface other than loopback when the operator explicitly opts in
- add regression tests covering the default and the env override on the client side
- document both variables in `README.md` and `README.zh-CN.md`, with a security note on `OPENCLI_DAEMON_BIND` (the daemon has no built-in auth)

## Why
The daemon URL was a hardcoded `http://127.0.0.1:${OPENCLI_DAEMON_PORT}` and the listener was hardcoded to `127.0.0.1`. This makes it impossible to use OpenCLI from a container that wants to drive Chrome on the host: even when `host.docker.internal:19825` is reachable, the CLI still dials `127.0.0.1` inside the container and fails.

Both new env vars default to `127.0.0.1`, so all existing deployments are byte-for-byte unchanged.

## Verification
- `npx tsc --noEmit` — clean
- `npx vitest run --project unit src/browser/daemon-client.test.ts` — 13/13 pass, including the two new cases
- Manual: with `OPENCLI_DAEMON_BIND=0.0.0.0` on the host and `OPENCLI_DAEMON_HOST=host.docker.internal` inside an OrbStack container, `opencli twitter trending` returns data from the host's logged-in Chrome session
- The broader unit suite has unrelated failures in this dev env (project requires Node >= 21, my container has Node 20 → `undici` `webidl.util.markAsUncloneable` errors); confirmed identical failures on `main` before any change, so they are pre-existing and unrelated to this PR


jackwener:
关闭此 PR。current main 的安全模型明确把 Browser Bridge daemon 固定在 127.0.0.1:19825；docs/guide/remote-orchestration.md 也要求通过反向端口转发保持两端 loopback，而不是让无内建认证的 daemon 绑定 0.0.0.0。此 PR 的 OPENCLI_DAEMON_BIND/HOST 会把浏览器会话控制面暴露到网络，目标本身与现行安全边界冲突。

PR #1426, opened 2026-05-09
feat(runtime): add project-local adapter and plugin discovery under ./.opencli/

### LINKED ISSUE #1423: Feature: disable built-in adapters + project-level CLI isolation
## Problem

Currently opencli loads all built-in adapters unconditionally. There is no way to suppress them, which causes two practical issues:

### 1. Built-in adapters pollute the agent tool list

When using opencli as an AI agent tool (e.g. via the `smart-search` skill in Claude Code), **all 140+ built-in adapters are exposed to the model at once**. This bloats the tool listing, increases token usage, and makes it harder for the agent to pick the right tool. There is no mechanism to disable or hide specific sites.

**Request:** Add a config option to disable built-in adapters, e.g. in `~/.opencli/config.json`:

```json
{
  "disableBuiltins": true,
  "disabledSites": ["1688", "amazon", "bloomberg"]
}
```

Or an env var: `OPENCLI_DISABLE_BUILTINS=1`

---

### 2. No project-level CLI isolation

The current discovery order is fixed:
1. `BUILTIN_CLIS` (npm package)
2. `~/.opencli/clis/` (user-level)
3. `~/.opencli/plugins/` (user-level)

There is **no project-local discovery**. When working in a monorepo or a project that defines its own domain-specific tools, there is no way to:
- Scope a set of CLIs to a specific project directory
- Have different projects expose different tool sets to the agent
- Check project-local tools into version control alongside the code

**Request:** Add a project-level discovery step, loading from the current working directory:

```
./.opencli/clis/     ← project-local adapters
./.opencli/plugins/  ← project-local plugins
```

Combined with `disableBuiltins`, this would allow a project to define exactly the tools it needs and expose only those to the agent — clean, isolated, version-controlled.

---

## Use case

A project that communicates with internal APIs wants to expose 3 custom tools to an AI agent, without the agent seeing 140+ unrelated built-in adapters. Today this is impossible without patching the npm package directly.

## Version

`@jackwener/opencli` v1.7.14

Benjamin-eecs:
## Description

Adds a project-local discovery step that loads adapters from `./.opencli/clis/<site>/<command>.js` and plugins from `./.opencli/plugins/<plugin>/` when opencli runs inside a project. Mirrors the existing user-level layer at `~/.opencli/` (same import surface, same compat shim mechanism), so a repo can check project-specific adapters into VCS and they override built-ins for that working directory.

Closes #1423.

### Discovery order

Last wins on collision (mirrors existing `registerCommand` semantics in `src/registry.ts`):

```
built-in
  -> ~/.opencli/clis
  -> ./.opencli/clis
  -> ~/.opencli/plugins
  -> ./.opencli/plugins
```

A project adapter shadows a user adapter, which shadows a built-in. External CLIs (`gh`, `ntn`, `tg`, `discord`, `wx`, ...) are passthrough binaries on a separate registration path (`src/external.ts` / `external-clis.yaml`) and are unaffected.

### Scope

Implements the project-isolation half of #1423 only. The disable-builtins half is left out; external-CLI replacement (#1559, #1544) is already cutting the built-in surface, so a runtime flag can wait for a follow-up if it's still useful after that migration settles.

## Type of Change

- [x] ✨ New feature
- [ ] 🐛 Bug fix
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

### `src/discovery.ts`

- `projectOpenCliDir(cwd?)` / `projectClisDir(cwd?)` / `projectPluginsDir(cwd?)`: path helpers, default to `process.cwd()`. Lazy resolution so they pick up the actual working directory at invocation time.
- `ensureProjectCliCompatShims(cwd?)`: thin wrapper that calls existing `ensureUserCliCompatShims(baseDir)` with a project-local `baseDir`, but only when `./.opencli/` already exists. Silent no-op when the project has no `.opencli/` directory, so projects that don't use the feature pay zero startup cost.
- `discoverPlugins(dir = PLUGINS_DIR)`: widen the existing signature with an optional `dir` parameter so project-local plugins can be loaded from `./.opencli/plugins/` using the same flat-file scan.

### `src/main.ts`

- `PROJECT_CLIS` / `PROJECT_PLUGINS` constants.
- `ensureProjectCliCompatShims()` added to the parallel startup block alongside the existing user-level setup.
- `await discoverClis(PROJECT_CLIS)` after the user-CLI pass.
- `await discoverPlugins(PROJECT_PLUGINS)` after the default plugins pass.

### `src/discovery-project.test.ts`

Three tests covering:
- Path helpers resolve relative to a given `cwd`.
- `discoverClis(projectDir)` registers an adapter dropped into a project-local site directory.
- `discoverPlugins(projectDir)` registers a plugin dropped into a project-local plugin directory.

### `docs/guide/extending-opencli.md`

Adds a "Project-local adapters and plugins" section documenting the `./.opencli/` layout and the discovery order. Adds a row to the extension-paths table at the top.

## Screenshots / Output

```bash
$ mkdir -p my-project/.opencli/clis/hello
$ cat > my-project/.opencli/clis/hello/world.js <<'JS'
import { cli, Strategy } from '@jackwener/opencli/registry';
cli({
  site: 'hello', name: 'world', access: 'read',
  description: 'project-local greeting',
  strategy: Strategy.PUBLIC, browser: false,
  func: async () => [{ msg: 'hi from ./.opencli' }],
});
JS

$ cd my-project && node /path/to/OpenCLI/dist/src/main.js hello world -f json
[{ "msg": "hi from ./.opencli" }]
```

A project adapter with the same `site/command` as a built-in overrides the built-in for the duration of `cd my-project`; leave the directory and the built-in is back.

## Test plan

- [x] `npx vitest run src/discovery-project.test.ts`: 3 passed
- [x] `npm run build`: manifest compiles, 813 entries, no path leaks
- [x] `npx tsc --noEmit`: clean
- [x] Project-local adapter registers and runs when `./.opencli/clis/<site>/<cmd>.js` is present
- [x] Project without `./.opencli/` directory pays zero startup cost (early-return in `ensureProjectCliCompatShims`)
- [ ] Reviewer to confirm precedence behaviour on a project that shadows a built-in (`./.opencli/clis/twitter/search.js` should win over the bundled `twitter/search.js`).

### DIFF
--- a/docs/guide/extending-opencli.md
+++ b/docs/guide/extending-opencli.md
@@ -1,6 +1,6 @@
 # Extending OpenCLI
 
-OpenCLI has five extension paths. Pick the path based on where you want the source code to live and how you want commands to be shared.
+OpenCLI has six extension paths. Pick the path based on where you want the source code to live and how you want commands to be shared.
 
 | Goal | Use | Source location | Command surface |
 |------|-----|-----------------|-----------------|
@@ -8,6 +8,7 @@ OpenCLI has five extension paths. Pick the path based on where you want the sour
 | Quickly draft a private adapter on this machine | User adapter | `~/.opencli/clis/<site>/<command>.js` | `opencli <site> <command>` |
 | Edit an official adapter locally | Adapter override | `~/.opencli/clis/<site>/` | `opencli <site> <command>` |
 | Publish or install third-party commands | Plugin | Git repo, installed into `~/.opencli/plugins/` | `opencli <plugin> <command>` |
+| Scope commands to a single project | Project-local adapter / plugin | `./.opencli/clis/<site>/` or `./.opencli/plugins/<plugin>/` | `opencli <site> <command>` |
 | Wrap an existing local binary | External CLI | `~/.opencli/external-clis.yaml` | `opencli <tool> ...` |
 
 ## Personal commands in your own Git repo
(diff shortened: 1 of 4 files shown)


Benjamin-eecs:
Closing. The project-local discovery layer here extends framework-extension surface, which doesn't align with the current curated-CLI direction. #1423 stays open for the maintainer to decide on the disable-builtins half.

PR #1428, opened 2026-05-09
feat(reddit): add reply command for replying to comments

gauravsaxena1997:
## What and why

Reddit's adapter set had `comment` (for top-level post comments) but no `reply` command for responding to an existing comment. This gap meant any agent or script that needed to reply to a specific comment had no way to express that intent cleanly.

The `comment` adapter internally handles `t1_` prefixes, but its arg is named `post-id` with a help string of "Post ID" — so passing a comment ID is undiscoverable and confusing. An agent reasoning about `opencli reddit --help` would never infer that `comment` also handles replies.

## The gap this hit in practice

Running a batch of Reddit engagement tasks, 5 out of 6 commands failed with `error: unknown command 'reply'` because the script naturally expressed reply intent as `opencli reddit reply <comment-id> "..."`. There was no fallback — the command simply didn't exist. Adding `reply.js` to `~/.opencli/clis/reddit/` as a private adapter fixed all 5 instantly, which confirmed the implementation is correct.

## What this PR does

Adds `clis/reddit/reply.js` — a `Strategy.COOKIE` adapter that replies to a Reddit comment by its ID:

```bash
opencli reddit reply okf3s7u "Great point, here's my take..."
opencli reddit reply t1_okf3s7u "Also works with the t1_ fullname prefix"
```

**Implementation details:**
- Takes `comment-id` (positional) and `text` (positional) — mirrors the arg design of `comment.js`
- Normalises the input: bare ID (`okf3s7u`) or fullname (`t1_okf3s7u`) or a full comment URL are all accepted
- Uses Reddit's `/api/comment` endpoint with `parent=t1_<id>` — same API as top-level comments, just with the `t1_` prefix instead of `t3_`
- Fetches modhash from `/api/me.json` before posting (same pattern as `comment.js`)
- Full `try/catch` inside `page.evaluate` with structured error surfacing
- Returns `[{ status, message }]` — same column schema as `comment.js`

## Checklist

- [x] `opencli validate` passes (no errors on the new adapter)
- [x] `npx tsc --noEmit` clean
- [x] Manually tested: replied to 5 live Reddit comment threads successfully
- [x] Follows arg design convention (positional for primary args)
- [x] Consistent style with existing `clis/reddit/comment.js`


jackwener:
Pushed `25bdda81` directly (maintainer-can-modify) to upgrade the failure handling before merging — the original `reply.js` mirror-copied `comment.js`'s silent-sentinel row pattern, which is exactly what we should avoid in **new** files per typed-errors.md anti-patterns and Round 21 lesson (f) ("grandfathered-not-exempt + helper-refactor boundary is new"). The three sibling write adapters (`comment.js` / `upvote.js` / `save.js`) stay grandfathered until a separate cleanup PR.

### What changed

| Before | After |
|---|---|
| `[{ status: 'failed', message: 'HTTP 403' }]` on any failure | Throws `AuthRequiredError('reddit.com', ...)` for 401/403 + missing `me.data.name` |
| Reddit `json.errors` swallowed as failed row | Throws `CommandExecutionError('Reddit rejected reply: <reason>')` |
| Non-2xx HTTP swallowed as failed row | Throws `CommandExecutionError('HTTP <status> from <where>')` |
| `if (!page) throw ...` over-defensive | Removed (registry guarantees `page` when `browser: true`) |

Auth detection is two-pronged: 401/403 on `/api/me.json` **or** 200 with empty `data.name`. The empty-modhash check alone is too weak — Reddit sometimes returns 200 with an empty modhash for stale anonymous sessions, which would have produced misleading `'Comment posted'` rows on logged-out runs.

### Audit / lint

- `npx tsc --noEmit` — clean
- `node scripts/check-typed-error-lint.mjs` — 189 / 189, 0 new
- `node scripts/check-silent-column-drop.mjs` — 103 / 103, 0 new (intermediate `result` object uses `kind` / `detail` / `httpStatus` / `where` keys that don't overlap `columns: ['status','message']`, per PR #1329 sediment)
- `npx vitest run clis/reddit src/convention-audit` — 11 / 11 passing
- `node ./dist/src/main.js validate` — 0 errors

Success row is unchanged: still `[{ status: 'success', message: 'Reply posted on t1_<id>' }]`. Your "manually tested 5 live threads" coverage still applies.

Thanks for the contribution — the gap analysis ("`opencli reddit reply` returns `unknown command`, agents naturally express reply intent that way") is exactly the kind of agent-facing rough edge worth closing.

PR #1451, opened 2026-05-10
perf(twitter): drop redundant goto+wait — framework auto pre-navs (PR C)

jackwener:
## Summary

Twelve twitter read adapters did `await page.goto('https://x.com'); await page.wait(2~3)` purely to populate the cookie context for a `document.cookie` read. After #1450 (PR A) hoisted the cookie reads to `page.getCookies({url})` — which queries the CDP cookie store directly with no navigation requirement — the explicit goto+wait became dead.

The framework already pre-navigates to `https://${domain}` for any adapter that declares `Strategy.COOKIE + domain` (`src/registry.ts:191`), so by the time `func` runs the cookie store is already populated. The 2-3s `page.wait` was the slowest part of the redundant call — same speed-up family as the original `thread.js` 35s→9s in #1450.

This is the **PR C** half of the cookie-API plan from #OpenCLI:3889b5cf (PR A merged → PR C now → PR B `browserSession: { reuse: 'site' }` next).

## Scope

12 files, all read-only, all ct0/cookie-only adapters that paid the goto+wait tax purely for cookie context:

| File | Was | Now |
|---|---|---|
| `clis/twitter/bookmark-folder.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/bookmark-folders.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/bookmarks.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/following.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/likes.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/list-add.js` (L87 only) | goto+wait+getCookies | getCookies |
| `clis/twitter/list-remove.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/list-tweets.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/lists.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/thread.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/timeline.js` | goto+wait+getCookies | getCookies |
| `clis/twitter/tweets.js` | goto+wait+getCookies | getCookies |

Diff: 12 files, +4/-30 (net -26 lines including stale `// Navigate to x.com for cookie context` comments).

## Out of scope (intentional, kept as-is)

These do `page.goto` to a *specific* URL needed for content/SPA shell, not for cookie context:

- `trending.js` → `/explore/tabs/trending` (DOM scrape target)
[...]

PR #1450, opened 2026-05-10
perf(adapters): hoist cookie reads to page.getCookies (Tier 1, 25 files)

jackwener:
## Summary

Replaces `page.evaluate(\`document.cookie.split…\`)` round-trips with direct CDP `page.getCookies({domain})` reads across **25 files** in 4 sites (twitter / linkedin / maimai / youtube). Part of the goto-and-wait audit follow-up per #OpenCLI msg=fa209a2c.

`twitter thread 2051987950200648006` cold-run profile breakdown (with framework pre-nav already in place) showed `goto + wait(3) + page.evaluate(document.cookie)` together accounted for ~20s of a 35s run; this PR is the first surgical cut — removing the `document.cookie` extraction path. Sibling PRs B (`browserSession: { reuse: 'site' }` for read-only twitter) and C (delete the redundant in-func `goto + wait`) are scoped separately.

## Patterns

| Site | Files | Pattern |
|---|---|---|
| twitter | 13 | Cookie was extracted *outside* `page.evaluate` — direct 4-line replacement: `await page.getCookies({domain:'.x.com'})` + `cookies.find((c) => c.name === 'ct0')?.value` |
| twitter | 2 (article, profile) | Cookie was extracted *inside* `page.evaluate` — hoisted out and `AuthRequiredError` is now thrown upfront on the Node side, dropping the unreachable in-evaluate `error: 'No ct0 cookie'` branch + its outside re-throw |
| linkedin | 1 (search) | `JSESSIONID` was read inside the per-batch fetch loop — hoisted once before the loop, `csrf` interpolated into the fetch template as `JSON.stringify(csrf)` |
| maimai | 1 (search-talents) | csrftoken cookie hoisted; meta-tag fallback preserved inside the template. Also converted the `page.evaluate(async (body) => …, body)` Playwright-style call (which doesn't match OpenCLI's `evaluate(js: string)` signature) to a template-string form so the helper actually runs |
| youtube | 5 (utils + like/unlike/subscribe/unsubscribe) | `SAPISID_HASH_FN` reshaped to take `sapisid` as a parameter; new `readYoutubeSapisid(page)` helper reads `__Secure-3PAPISID` (preferred) / `SAPISID` via CDP. HMAC-SHA1 compute still happens browser-side (Web Crypto); only the cookie read is hoisted |

## Tests

[...]

PR #1454, opened 2026-05-10
perf(twitter): enable browserSession reuse:site on 17 read-only adapters (PR B)

jackwener:
## Summary

Read-only Twitter/X adapters now declare `browserSession: { reuse: 'site' }`, matching the LLM-site adapters (claude/gemini/yuanbao/etc.). This unblocks the perf wins WAWQAQ called out for the thread.js 35s→9s/3.4s progression (#OpenCLI:3889b5cf, msg=fa209a2c).

This is **PR B** in the agreed cookie-API/perf sweep:
- PR A #1450 (merged, `60dbbd4b`) — `document.cookie` → `page.getCookies({domain})` hoist
- PR C #1451 (merged, `ff7d741a`) — drop redundant `goto + wait` (framework auto pre-navs)
- **PR B (this) — `browserSession: { reuse: 'site' }` for read-only twitter**
- D-track (coding-claude-opus, parallel) — wait→event sweep across LLM sites

## What changes

- Tab lease shared across calls under `site:twitter` until idle expiry, so the second-and-later command pays no cold-start tab cost.
- Framework's domain-root pre-nav (`https://x.com`) is skipped on subsequent calls when the reused tab is already on x.com (`shouldRunPreNav` → `isDomainRootPreNav` + `urlMatchesDomain` short-circuit at `src/execution.ts:190`).

## Files (17 read-only adapters)

- **Strategy.COOKIE × 14**: article, bookmark-folder, bookmark-folders, bookmarks, download, following, likes, list-tweets, lists, profile, thread, timeline, trending, tweets
- **Strategy.UI × 1**: followers
- **Strategy.INTERCEPT × 2**: notifications, search

Insertion point in each file: after `browser: true,` (or after `strategy:` in download.js which omits the explicit `browser:` field), matching the convention used by yuanbao/read.js, claude/read.js, etc.

Manifest regenerated (`cli-manifest.json`: +85/-17 — 17 entries gain the `browserSession: { reuse: "site" }` block).

## Scope notes (intentionally NOT in this PR)

- **Write adapters** (post/reply/quote/like/retweet/bookmark/follow/list-add/list-remove/delete/hide-reply/block/accept/follow) are kept as one-shot by default — `reuse: 'site'` for write paths is a separate decision about action idempotency under tab reuse.
- **Comment polish on thread.js / timeline.js**: still says "Cookie context auto-established by framework pre-nav"; the deeper truth (CDP `getCookies({url})` is origin-independent — framing nit from coding-claude-opus on PR C) is left as a doc-only follow-up to keep this PR's diff focused on the perf gain.

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx vitest run clis/twitter --reporter=dot` → 218/218 pass (25 files)
- [x] `npx vitest run src/convention-audit.test.ts --reporter=dot` → 8/8 pass
- [x] `node scripts/check-typed-error-lint.mjs` → 189/189 baseline, no new violations
- [x] `node scripts/check-silent-column-drop.mjs` → 103/103 baseline, no new violations
- [x] `npm run build-manifest` → 801 entries, 17 manifest entries gain the new field
- [ ] Live verify on a couple of read commands back-to-back to observe the cold-start saving (requires logged-in twitter session — leaving for human verification per PR C path)

Refs: #OpenCLI:3889b5cf (WAWQAQ msg=fa209a2c, msg=35c90460, msg=838128ef "你们继续做啊… 后面还有那么多其他的东西呢")

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -22175,7 +22175,10 @@
     "type": "js",
     "modulePath": "twitter/article.js",
     "sourceFile": "twitter/article.js",
-    "navigateBefore": "https://x.com"
+    "navigateBefore": "https://x.com",
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
@@ -22273,7 +22276,10 @@
     "type": "js",
     "modulePath": "twitter/bookmark-folder.js",
     "sourceFile": "twitter/bookmark-folder.js",
-    "navigateBefore": "https://x.com"
+    "navigateBefore": "https://x.com",
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
@@ -22293,7 +22299,10 @@
     "type": "js",
     "modulePath": "twitter/bookmark-folders.js",
     "sourceFile": "twitter/bookmark-folders.js",
-    "navigateBefore": "https://x.com"
+    "navigateBefore": "https://x.com",
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
@@ -22332,7 +22341,10 @@
     "type": "js",
     "modulePath": "twitter/bookmarks.js",
     "sourceFile": "twitter/bookmarks.js",
-    "navigateBefore": "https://x.com"
+    "navigateBefore": "https://x.com",
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
@@ -22406,7 +22418,10 @@
     "type": "js",
     "modulePath": "twitter/download.js",
     "sourceFile": "twitter/download.js",
-    "navigateBefore": "https://x.com"
+    "navigateBefore": "https://x.com",
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
@@ -22466,7 +22481,10 @@
     "type": "js",
     "modulePath": "twitter/followers.js",
     "sourceFile": "twitter/followers.js",
-    "navigateBefore": true
+    "navigateBefore": true,
+    "browserSession": {
+      "reuse": "site"
+    }
   },
   {
     "site": "twitter",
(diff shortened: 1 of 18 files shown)


jackwener (commented):
LGTM (comment-mode, same-identity blocks formal approve).

Diff shape verified against PR description:
- 17 adapter files × 1 line `browserSession: { reuse: 'site' }` insertion after `browser: true,` (or after `strategy:` for download.js which omits the explicit browser line).
- cli-manifest.json regenerates 17 entries with the new key — counts match.
- Zero func-body changes, zero test changes — purely declarative metadata.

Scope split is the right call:
- Write adapters (post/reply/quote/like/retweet/bookmark/follow/list-add/list-remove/delete/hide-reply/block/accept/follow) excluded — `reuse:'site'` for write paths bumps into the path-state-pollution hazard from #1383 follow-up (e.g. `twitter follow xxx` leaves tab on `/xxx`, hostname-only pre-nav match would skip the home-reset that next read assumes).
- `followers` (Strategy.UI) is the only UI-strategy adapter in the set — not a concern because the func runs its own `goto(/${user}/followers)` regardless of incoming tab URL.
- INTERCEPT (notifications/search) — same pattern; both navigate to `/notifications` or `/search` themselves before listening.

Two non-blocking observations:

(1) **idle-timeout coupling.** `INTERACTIVE_BROWSER_IDLE_TIMEOUT_SECONDS = 600` (PR #1383). The perf gain only realizes when calls land within that window — first call after >10min still pays cold start. Worth knowing for live verify framing: the curve is 35s (cold) → 9s (warm tab + goto) → 3.4s (warm tab, no goto) per WAWQAQ msg=fa209a2c, but the 3.4s case is the steady-state, not the average across mixed cadence. No change requested — just makes the perf claim more precise in the live-verify report.

(2) **download.js omits `browser: true,`.** Diff places `browserSession` after `strategy: Strategy.COOKIE` because there's no `browser:` line. Pre-existing condition (not yours), and `Strategy.COOKIE` defaults to `browser:true` in framework, but worth noting if the convention-audit ever tightens this convention later — would catch a 0-edit cleanup PR. Out of scope for #1454.

Tests: 218/218 vitest twitter + 8/8 convention-audit + 189/189 typed-error + 103/103 silent-column-drop + 11/11 CI green. Approving with comment.

Refs: #OpenCLI:3889b5cf, msg=fa209a2c (perf curve), msg=73bbfc49 (PR #1383 follow-up framing).

PR #1452, opened 2026-05-10
perf(claude): replace fixed-sleep waits with selector-based readiness (D2)

jackwener:
## Summary
- D2 in the LLM-adapter wait→event sweep started by deepseek (D1, #1449)
- Convert 9 of 18 \`page.wait(N)\` calls in \`clis/claude/\` from fixed-duration sleeps to event-based readiness checks (\`page.wait({ selector, timeout })\`)
- Drop 6 redundant settle waits whose function is already covered by the next CDP roundtrip or the new event waits
- Keep 3 in-loop polling / React-debounce slots out of scope

## Wait site classification (18 sites)

**Converted (9):**
| File | Site | New selector | Why |
|------|------|--------------|-----|
| utils.js:29 | \`ensureOnClaude\` post-goto | \`COMPOSER_SELECTOR\` | composer is the canonical "page is ready" signal |
| utils.js:114 | \`getConversationList\` post-goto-/recents | \`a[href*="/chat/"]\` | recents links mount when history loads |
| utils.js:150 | \`selectModel\` post-trigger.click() | \`div[role="menuitemradio"]\` | dropdown menuitems mount when open |
| utils.js:178 | \`setAdaptiveThinking\` post-trigger.click() | \`div[role="menuitem"]\` | same |
| ask.js:41 | \`--new\` path post-goto | \`COMPOSER_SELECTOR\` | matches \`ensureOnClaude\` semantics |
| ask.js:51 | resume-conversation post-link.click() | \`MESSAGE_SELECTOR\` | resumed conversation renders bubbles |
| new.js:19 | post-goto | \`COMPOSER_SELECTOR\` | downstream \`ensureClaudeComposer\` already throws if missing |
| detail.js:24 | post-goto-/chat/{id} | \`MESSAGE_SELECTOR\` | first assistant bubble |
| send.js:26 | \`--new\` path post-goto | \`COMPOSER_SELECTOR\` | same as ask.js:41 |

All wrapped in \`try { ... } catch { /* fall through */ }\` so login redirects, empty conversations, and missing access still produce the right typed error downstream (\`AuthRequiredError\` / \`CommandExecutionError\` / \`EmptyResultError\`).

**Deleted (6):**
| File | Site | Why dropping is safe |
|------|------|----------------------|
| ask.js:55 | standalone settle | \`ensureClaudeComposer\` reads composer presence directly via \`getPageState\` |
[...]

PR #1449, opened 2026-05-10
perf(deepseek): replace fixed-sleep waits with selector-based readiness

jackwener:
## Summary

Proof template for the **D-LLM** sweep (#OpenCLI thread `3889b5cf`): convert deepseek's 18 `page.wait(N)` fixed-duration sleeps to event-based selector waits where the wait is genuinely tied to a DOM marker. Once this lands cleanly, PRs D2–D8 will copy the same pattern across claude / chatgpt / doubao / yuanbao / qwen / grok / gemini.

10 of 18 sites are touched here. The other 8 (in-loop polling ticks + send.js native-input flush) stay — they are already polling patterns and out of scope for D1.

### Changes

**Page-ready waits** — replaced `page.wait(N)` with `page.wait({ selector, timeout })` (MutationObserver, resolves as soon as the selector matches):

| Site | Old | New |
|------|-----|-----|
| `utils.js:46` (`ensureOnDeepSeek`) | `wait(3)` | wait for `TEXTAREA_SELECTOR`, timeout 8s, swallow on login page |
| `ask.js:38` (`--new` branch) | `wait(3)` | wait for `TEXTAREA_SELECTOR`, timeout 8s |
| `ask.js:52` (resume branch) | `wait(2)` | wait for `TEXTAREA_SELECTOR`, timeout 5s |
| `detail.js:28` | `wait(5)` | wait for `MESSAGE_SELECTOR`, timeout 10s |
| `new.js:19` | `wait(2)` | wait for `TEXTAREA_SELECTOR`, timeout 8s; throw `CommandExecutionError` on miss |

**Redundant settles deleted** (5 sites):

| Site | Reason |
|------|--------|
| `ask.js:56` standalone `wait(2)` | covered by the new selector waits in the if/else above |
| `ask.js:79` post-`selectModel` 0.5s settle | next CDP eval already gives React time to flush `aria-checked` |
| `ask.js:105` post-`setFeature` 0.5s settle | same as above |
| `ask.js:118` pre-`waitForResponse` 3s settle | `waitForResponse`'s first 3s tick already covers this |
| `read.js:19` post-`ensureOnDeepSeek` 5s settle | `ensureOnDeepSeek` now waits for the textarea selector itself |

**Behavior tightening**: `deepseek new` previously returned `Status: 'New chat started'` even when the composer never mounted (e.g., logged-out, network error). It now throws `CommandExecutionError` on timeout with a hint about login.

### What is **not** changed

[...]

PR #1456, opened 2026-05-10
perf(chatgpt): replace fixed-sleep waits with selector-based readiness (D3)

jackwener:
## Summary

Continues the wait→event sweep started in #1449 (deepseek) and #1452 (claude). Same 3-bucket classification across 5 chatgpt adapter files.

### CONVERT (5) — fixed sleep → `page.wait({selector, timeout})`
| Site | Before | After |
|---|---|---|
| `utils.js` `ensureOnChatGPT` | 2s settle after `goto(/)` | wait composer 8s |
| `utils.js` `startNewChat` | 2s settle after `goto(/new)` | wait composer 8s |
| `utils.js` `getConversationList` openSidebar | 1.5s | wait `a[href*="/c/"]` 3s |
| `utils.js` `getConversationList` fallback goto | 2.5s | wait composer 8s |
| `detail.js` post-`/c/<id>` goto | 2s | wait `[data-message-author-role]` 10s |

### DELETE (6) — redundant settles after helpers that now wait internally
- `ask.js` / `send.js` / `read.js`: standalone 2s settle after `ensureOnChatGPT` / `startNewChat`
- `utils.js sendChatGPTMessage`: 0.5s post-closeBtn wait + 1.5s pre-composer-focus wait (composer is already focused; the focus call itself is the readiness signal)
- `utils.js getConversationList`: 2s settle after `ensureOnChatGPT`

### KEEP (6) — legitimate polling cadence
- `sendChatGPTMessage` ProseMirror React debounce ticks
- `waitForChatGPTResponse` streaming response polling

## Risk

Low. Per-call cost: warm path (composer already mounted) drops from fixed 2s → ≤16ms (one selector poll). Cold path (composer mounting) caps at the same 8s the implicit-mount window already needed; if composer never mounts, downstream `ensureChatGPTComposer` / `ensureChatGPTLogin` still surface the typed error — try/catch around the `wait()` is intentional and matches D1/D2 precedent.

No typed-error harmonization needed in this PR — chatgpt's existing helpers already gate through `ensureChatGPTLogin` (AuthRequiredError) and `ensureChatGPTComposer` (CommandExecutionError) correctly. The deferred D1/D2 backport remains its own follow-up.

## Verification

| Check | Result |
|---|---|
| `npx vitest run clis/chatgpt` | 20/20 (4 files) |
| Full vitest | 3379 pass / 1 skip / 2 errors (`background.test.ts` + `daemon.test.ts` EADDRINUSE — unrelated, also seen on #1449/#1452/#1454/#1455) |
| `tsc --noEmit` | clean |
| `npm run check:typed-error-lint` | 189/189 unchanged |
| `npm run check:silent-column-drop` | 103/103 unchanged |
| `npm run build` | 802 manifest entries |

## Diff stat
+50 / -13 across 5 files (`ask.js`, `detail.js`, `read.js`, `send.js`, `utils.js`).

## Test plan
- [ ] CI green
- [ ] Cross-review by @opencli-user (per WAWQAQ msg=af5b5367 互 review 授权)

### DIFF
--- a/clis/chatgpt/ask.js
+++ b/clis/chatgpt/ask.js
@@ -43,7 +43,8 @@ export const askCommand = cli({
         } else {
             await ensureOnChatGPT(page);
         }
-        await page.wait(2);
+        // startNewChat / ensureOnChatGPT now wait for the composer selector
+        // after navigating, so the previous standalone 2 s settle is redundant.
         await ensureChatGPTComposer(page, 'ChatGPT ask requires a logged-in ChatGPT session with a visible composer.');
 
         const baseline = await getBubbleCount(page);
--- a/clis/chatgpt/detail.js
+++ b/clis/chatgpt/detail.js
@@ -3,6 +3,7 @@ import { EmptyResultError } from '@jackwener/opencli/errors';
 import {
     CHATGPT_DOMAIN,
     CHATGPT_URL,
+    CONVERSATION_MESSAGE_SELECTOR,
     ensureChatGPTLogin,
     getVisibleMessages,
     messageHtmlToMarkdown,
@@ -29,7 +30,11 @@ export const detailCommand = cli({
         const id = parseChatGPTConversationId(kwargs.id);
         const wantMarkdown = normalizeBooleanFlag(kwargs.markdown, false);
         await page.goto(`${CHATGPT_URL}/c/${id}`, { settleMs: 2000 });
-        await page.wait(4);
+        try {
+            await page.wait({ selector: CONVERSATION_MESSAGE_SELECTOR, timeout: 10 });
+        } catch {
+            // Empty conversation, missing access, or login redirect — handled by ensureChatGPTLogin / EmptyResultError below.
+        }
         await ensureChatGPTLogin(page, 'ChatGPT detail requires a logged-in ChatGPT session.');
         const messages = await getVisibleMessages(page);
         if (!messages.length) {
--- a/clis/chatgpt/read.js
+++ b/clis/chatgpt/read.js
@@ -25,8 +25,9 @@ export const readCommand = cli({
     columns: ['Index', 'Role', 'Text'],
     func: async (page, kwargs) => {
         const wantMarkdown = normalizeBooleanFlag(kwargs.markdown, false);
+        // ensureOnChatGPT now waits for the composer selector after navigating,
+        // so the previous standalone 2 s settle is redundant.
         await ensureOnChatGPT(page);
-        await page.wait(2);
         await ensureChatGPTLogin(page, 'ChatGPT read requires a logged-in ChatGPT session.');
         const messages = await getVisibleMessages(page);
         if (!messages.length) {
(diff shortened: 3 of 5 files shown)


jackwener (commented):
LGTM (comment-mode, same as #1451/#1454/#1449/#1452 cross-review pattern).

Audited 161-line diff in full vs D1/D2 precedent — three-bucket classification (CONVERT 5 / DELETE 6 / KEEP 6) is internally consistent, helper extraction (`COMPOSER_WAIT_SELECTOR` + `CONVERSATION_LINK_SELECTOR` + exported `CONVERSATION_MESSAGE_SELECTOR`) keeps selector source-of-truth in one place.

**Verified**:
- `CONVERSATION_MESSAGE_SELECTOR` (utils.js:138) matches the `querySelectorAll` at L310 inside `getVisibleMessages` byte-for-byte → no selector divergence between detail.js wait and downstream extraction.
- `try { await page.wait({selector, timeout: N}) } catch {}` pattern matches D1/D2 — silent-on-timeout is intentional because downstream `ensureChatGPTLogin` / `ensureChatGPTComposer` / `EmptyResultError` carries the typed-error responsibility. Validated by reading send.js:42-44 → `if (!sent) throw new CommandExecutionError(...)`.

**Three non-blocking observations**:

1. **`sendChatGPTMessage` pre-composer settle deletion (utils.js:212-223)** — both `wait(0.5)` + `wait(1.5)` removed; comment claims "next page.evaluate roundtrip flushes React update". This is *probably* true (CDP marshalling latency ≥ React microtask commit), but not strictly guaranteed under fast local Chrome. **Failure mode is graceful** though: race → `findComposer()` returns null → `typeResult=false` → caller throws `CommandExecutionError('Failed to send message...')`. So worst case = transient flaky typed error, not silent wrong data. Non-blocker; if it shows up in field flake reports a cheap belt-and-suspenders is `try { await page.wait({selector: COMPOSER_WAIT_SELECTOR, timeout: 2}) } catch {}` after the close-sidebar click.

2. **`COMPOSER_WAIT_SELECTOR` is a subset of `COMPOSER_SELECTORS`** (utils.js:131-133 comment correctly admits this). Drift risk if a future variant is added to `COMPOSER_SELECTORS` and *only* that one mounts on some user — wait would timeout while composer is actually present, then `findComposer()` would still locate it, `ensureChatGPTComposer` passes. So drift = wasted timeout, not wrong-data. Same risk class as deepseek/claude. Non-blocker.

3. **typed-error harmonization**: agree with your call — chatgpt's existing `ensureChatGPTLogin` (AuthRequiredError) + `ensureChatGPTComposer` (CommandExecutionError) split is already correct. D1/D2 backport of `new.js` mount-timeout → `AuthRequiredError` is its own follow-up PR, doesn't block here.

`docs/adapters/browser/chatgpt.md` doesn't reference any of the deleted `page.wait(N)` numbers (pre-checked), so no doc drift.

D-track 第三连续 silent-failure-free pattern locked in. 🟢

Issue #1043, opened 2026-04-15
[Feature]: 实现类似qoderwork的浏览器打开效果，新建网页分组打开，不要自动关闭窗口


lww2626:
### Feature Description

<img width="240" height="77" alt="Image" src="https://github.com/user-attachments/assets/a0792a90-8c49-4574-b351-eb132875d418" />

qoderwork就比较好， 和opencli很类似，也是插件方式， 会在当前的浏览器窗口自动创建分组打开网页，不会自动关闭网页

### Use Case

我爬取5i5j和贝壳房源的时候，会不停地打开不同houseid的tab，获取网页元素解析，当出现验证码或着扫码的时候，脚本会异常终端，我当时不在电脑前，过一段时间也可以在当前扫码网页认证或着重新登录，然后可以重跑脚本。最近1.7.3版本，不知道支持这种操作吗？OPENCLI_WINDOW_FOCUSED=1 opencli baidu bottom-layer 类似这样的操作我也看不到浏览器， 是自动关闭了吗？

### Proposed Solution

希望能在当前的浏览器窗口打开网页，增加选项不自动关闭

Astro-Han:
I think this issue is mixing a few different requests:

- keep the automation window open
- reuse the current browser window or tab
- open new pages as grouped tabs in the current window

Those are related, but they are not the same change.

Also, `OPENCLI_WINDOW_FOCUSED=1` only controls whether the automation window opens in the foreground. It does not keep the window alive after the command finishes.

If your main use case is captcha, QR login, or other cases where you want the page to stay open until you come back, I would focus this issue on one thing first: do not auto-close the browser window after the command completes.

lww2626:
> I think this issue is mixing a few different requests:
> 
> * keep the automation window open
> * reuse the current browser window or tab
> * open new pages as grouped tabs in the current window
> 
> Those are related, but they are not the same change.
> 
> Also, `OPENCLI_WINDOW_FOCUSED=1` only controls whether the automation window opens in the foreground. It does not keep the window alive after the command finishes.
> 
> If your main use case is captcha, QR login, or other cases where you want the page to stay open until you come back, I would focus this issue on one thing first: do not auto-close the browser window after the command completes.

是的，目前最需要的是主动显示的窗口不要自动关闭

jackwener:
The main use case described here is now covered on current main.

[...]

PR #1463, opened 2026-05-10
feat(extension): add opt-out flag to disable automation tab grouping

keen0206:
## Summary

Adds a `chrome.storage.local` boolean (`opencli_disable_automation_tab_group`) that, when set to `true`, makes the extension skip grouping owned tabs into a named tab group (`OpenCLI Browser` / `OpenCLI Adapter`). Default behavior is unchanged.

## Motivation

Chrome 119+ ships a "Saved Tab Groups" feature that auto-saves any named/colored tab group to the user's sidebar and keeps it there even after the underlying window is closed. For workflows that repeatedly spin up short-lived owned windows (e.g. periodic adapter calls from a scheduled job), entries accumulate in the saved-groups list with no API surface to remove them — `chrome.tabGroups.*` doesn't expose deletion of saved entries, and the three related Chrome flags (`tab group sync` / `tab groups save` / `saved tab groups`) have all been removed.

Closing tabs / windows / ungrouping via `chrome.tabs.remove`, `chrome.windows.remove`, or `chrome.tabs.ungroup` all leave the saved entry behind. The only effective mitigation is **never to name/color the group in the first place**.

Rather than remove the existing grouping behavior — which other users explicitly requested (see #1043) for visual organisation — this PR adds an opt-out flag so users running automation-heavy workflows can suppress it without affecting anyone else.

## What changes

- `extension/src/background.ts`:
  - New constant `DISABLE_TAB_GROUP_KEY = 'opencli_disable_automation_tab_group'`
  - New helper `isAutomationTabGroupDisabled()` reading `chrome.storage.local`
  - Early-return added to `ensureOwnedContainerTabGroup` when the flag is set
- `extension/src/background.test.ts`: new test covering the opt-out path (asserts neither `chrome.tabs.group` nor `chrome.tabGroups.update` is called when the flag is set; tab id is still resolvable)
- `extension/README.md`: new "Configuration" section documenting the flag, the rationale (Chrome 119+ saved tab groups), and how to enable/revert from the service worker DevTools console
- `extension/dist/background.js`: rebuilt via `npm run build`

## How tabs stay tracked without grouping

The extension already tracks owned tabs by id via the `automationSessions` map plus persisted state in `chrome.storage.local[REGISTRY_KEY]`. The tab group is purely a visual marker — disabling it does not affect adapter functionality, lease lifecycle, or tab reuse logic. The flag only skips `chrome.tabs.group` + `chrome.tabGroups.update`; everything else is unchanged.

## Test plan

- [x] `cd extension && npx tsc --noEmit` — passes
- [x] `npx tsc --noEmit` (repo root) — passes
- [x] `npx vitest run extension/src/background.test.ts` — 51/51 pass (50 existing + 1 new)
- [x] `cd extension && npm run build` — produces updated `dist/background.js`
- [ ] Manual smoke: load patched extension, set the flag, run an adapter command, confirm no new entry appears in Chrome's saved tab groups sidebar

### DIFF
--- a/extension/README.md
+++ b/extension/README.md
@@ -24,3 +24,29 @@ Suggested Chrome Web Store justification for `downloads`:
 > so agents can wait for downloads triggered during an automation workflow. The
 > command filters by a user-provided filename or URL pattern and timeout. We do
 > not modify, redirect, or persist user download history.
+
+## Configuration
+
+### `opencli_disable_automation_tab_group` (chrome.storage.local, boolean)
+
+When set to `true`, the extension stops placing owned tabs into a named tab
+group (`OpenCLI Browser` / `OpenCLI Adapter`). Tabs are still tracked
+internally by id, so adapters keep working — only the visual grouping is
+suppressed.
+
+Why this exists: Chrome 119+ ships a "Saved Tab Groups" feature that
+auto-saves any named/colored tab group to the user's sidebar and keeps it
+there even after the underlying window is closed. For automation-heavy
+workflows that spin up many short-lived owned windows, this causes
+`OpenCLI Adapter` (or `OpenCLI Browser`) entries to accumulate. The flag
+lets users opt out.
+
+To enable, open the extension's service worker DevTools console
+(`chrome://extensions` → OpenCLI → "Inspect views: service worker") and run:
+
+```js
+chrome.storage.local.set({ opencli_disable_automation_tab_group: true });
+```
+
+To revert, set the same key to `false` (or remove it) and reload the
+extension. The default behavior (tab grouping enabled) is unchanged.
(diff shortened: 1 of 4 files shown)


GreyC:
Related follow-up issue: #1535. #1463 adds an opt-out for automation tab grouping; #1535 is narrower and keeps grouping enabled but reuses an existing OpenCLI Adapter/legacy OpenCLI group before creating a duplicate.

jackwener:
Extension 1.0.17（[#1794](https://github.com/jackwener/OpenCLI/pull/1794) / v1.8.1）确立了 OpenCLI Browser/Adapter tab group 的 **strict ownership** 设计：

- canonical title 强制 normalize（用户 rename 会被改回）
- 同 profile 跨 window dup 自动收敛进 canonical group
- create-fail rollback
- docs 明确写明用户不应在 OpenCLI tab group 中放入自己的 tab 或重命名（automation container）

引入 opt-out flag 让用户禁用 grouping 会让 extension 同时维护两套并行 tab 管理 mode（with/without），增加 self-healing surface 和分支复杂度，与 1.0.17 收敛到 single source of truth 的方向相反。

关闭本 PR。如果未来出现明确的用户场景需要 disable grouping，可以重新讨论；目前 single mode 更利于 invariant 维护。

PR #1512, opened 2026-05-12
fix(twitter): repair search and tweets readback

darthjaja6:
## Summary\n- switch twitter/search to the current SearchTimeline POST GraphQL path instead of brittle SPA navigation\n- resolve current X operation metadata from loaded client bundles and normalize Browser Bridge GraphQL payload envelopes\n- repair twitter/tweets parsing for current UserTweets payloads and add post id/url readback columns\n\n## Verification\n- npx vitest run --project adapter clis/twitter/search.test.js clis/twitter/tweets.test.js clis/twitter/post.test.js clis/twitter/shared.test.js\n- npm run typecheck\n- npm run build\n- npm run check:silent-column-drop\n- live: node dist/src/main.js --profile 86v9bx9k twitter search "VibeShare agent publishing" --limit 1 -f json --trace retain-on-failure\n- live: node dist/src/main.js --profile 86v9bx9k twitter tweets darthjajaj6z --limit 1 -f json --trace retain-on-failure

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -23244,7 +23244,9 @@
     "columns": [
       "status",
       "message",
-      "text"
+      "text",
+      "id",
+      "url"
     ],
     "type": "js",
     "modulePath": "twitter/post.js",
@@ -23460,7 +23462,7 @@
     "description": "Search Twitter/X for tweets, with optional --from / --has / --exclude / --product filters mapped to X's search operators",
     "access": "read",
     "domain": "x.com",
-    "strategy": "intercept",
+    "strategy": "cookie",
     "browser": true,
     "args": [
       {
@@ -23553,7 +23555,7 @@
     "type": "js",
     "modulePath": "twitter/search.js",
     "sourceFile": "twitter/search.js",
-    "navigateBefore": true,
+    "navigateBefore": "https://x.com",
     "siteSession": "persistent"
   },
   {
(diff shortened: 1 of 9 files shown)


jackwener:
Maintainer-edit `05f26048` — fix a silent-fallback bug in `sanitizeTwitterOperationMetadata`.

**The bug**

`resolveTwitterOperationMetadata`'s bundle parser locates `queryId:"..."` and then independently extracts `featureSwitches:[...]` / `fieldToggles:[...]` via separate regexes. When minification rearranges the snippet (or the 2500-char snippet window truncates before the array), either regex can miss while queryId still resolves. `keysToFlags(undefined)` then returns `{}`.

`sanitizeTwitterOperationMetadata` previously accepted any object as `features` / `fieldToggles`, including `{}`. Twitter's GraphQL endpoint rejects SearchTimeline / UserTweets requests with empty features (HTTP 400), surfacing the misleading "queryId may have expired" error — the queryId is actually fresh; only the feature flags are missing.

**The fix**

Defer to the baked fallback whenever the resolved map is empty:

```js
features: value?.features
    && typeof value.features === 'object'
    && Object.keys(value.features).length > 0
    ? value.features
    : normalizedFallback.features,
```

**Reverse-validation**

I stashed the production change and ran just the new test against unfixed code:

```
- "features": { "fallback_feature": true },
+ "features": {},
```

Confirms the test would have caught the bug.

**Tests + audits**

- `clis/twitter/shared.test.js`: 20 → 22 (+ 2 new — empty-map fallback + null fallback)
- Full twitter suite: 222/222
- `check-silent-column-drop.mjs`: 103/103 (0 new)
- `check-typed-error-lint.mjs`: 189/189 (0 new)

PR #1515, opened 2026-05-13
[codex] fix xiaohongshu download image order

### LINKED ISSUE #1514: [Bug] xiaohongshu download may save carousel images in DOM order instead of display order
## Problem

`opencli xiaohongshu download` can save carousel images in a different order from the order shown on Xiaohongshu/Rednote.

Observed on an image note where the platform-visible cover is the second downloaded file:

- note id: `69f9716c000000003601f90e`
- note title: `找到一个可以全网检测情报的 Skill`
- command shape: `opencli xiaohongshu download <signed-note-url-or-xhslink> --output ./xhs-order-test`
- observed: `<noteId>_1.jpg` is not the visible cover; `<noteId>_2.jpg` is the visible cover
- expected: downloaded image order should match the platform carousel/display order, so `<noteId>_1.jpg` should be the displayed first image/cover

I am not including the signed `xsec_token` URL in this public issue; the problem is about extraction order, not authentication.

## Likely root cause

Current `clis/xiaohongshu/download.js` collects image URLs from DOM selectors such as `.swiper-slide img`, `.carousel-image img`, `.note-slider img`, etc., stores them in a `Set`, and later appends that Set order into `result.media`.

That means the saved order is DOM discovery order, not necessarily the platform's canonical carousel order. DOM order can be affected by hidden slides, preloaded slides, duplicated carousel nodes, thumbnails, or responsive/lazy rendering.

The shared `downloadMedia` helper then names files by the array index it receives (`${filenamePrefix}_${i + 1}.jpg`), so once the adapter passes a non-display-ordered media array, the filenames and returned `index` become misleading for downstream consumers.

## Suggested fix

Prefer canonical note media data from the page state before falling back to DOM scraping:

1. In `xiaohongshu/download`, inspect `window.__INITIAL_STATE__` / note detail map for the note's image list.
2. Preserve the order of the platform-provided image array.
3. Use DOM selector extraction only as a fallback when structured note media is unavailable.
4. Optionally include extra metadata in media items/results (`sourceIndex`, `url`, `filename`, maybe `role: cover/body`) so downstream tools can audit or recover order.
5. Add a regression test where DOM discovery order differs from the canonical note image list order, and assert `downloadMedia` receives the canonical order.

This matters for workflows that analyze Xiaohongshu visual storytelling, because cover/body/back-page classification depends on stable carousel order.

Scofy0123:
## Summary

- Fixes #1514
- Prefer Xiaohongshu page-state image arrays when extracting note media for `xiaohongshu download`
- Keep the existing DOM selector extraction as a fallback when structured note media is unavailable
- Add a regression test that proves canonical state image order wins over mismatched DOM discovery order

## Root Cause

`xiaohongshu/download` previously collected images from carousel-related DOM selectors and stored them in a `Set`. That preserved DOM discovery order, not necessarily the order shown in the platform carousel. Hidden slides, preloaded slides, duplicated carousel nodes, or responsive rendering can make the visible cover appear after another DOM image, causing filenames like `<noteId>_1.jpg` to be misleading.

The shared `downloadMedia` helper simply names files by the media array index it receives, so the adapter needs to pass media in platform display order.

## Validation

- `npm run test:adapter -- clis/xiaohongshu/download.test.js`
- `npm run test:adapter -- clis/xiaohongshu/*.test.js`
- `npm run typecheck`

### DIFF
--- a/clis/xiaohongshu/download.js
+++ b/clis/xiaohongshu/download.js
@@ -39,6 +39,81 @@ export function buildDownloadExtractJs(noteId) {
           seenMedia.add(key);
           result.media.push({ type, url });
         };
+        const cleanImageUrl = (value) => {
+          if (typeof value !== 'string') return '';
+          let src = value.trim();
+          if (!src) return '';
+          if (!(src.includes('xhscdn') || src.includes('xiaohongshu') || src.includes('rednote'))) return '';
+          src = src.split('?')[0];
+          return src.replace(/\\/imageView\\d+\\/\\d+\\/w\\/\\d+/, '');
+        };
+        const imageUrls = new Set();
+        const addImageUrl = (value) => {
+          const src = cleanImageUrl(value);
+          if (src) imageUrls.add(src);
+        };
+        const getImageUrl = (value) => {
+          if (!value) return '';
+          const direct = cleanImageUrl(value);
+          if (direct) return direct;
+          if (typeof value !== 'object') return '';
+          const candidates = [
+            value.urlDefault, value.url_default,
+            value.urlPre, value.url_pre,
+            value.url, value.src,
+            value.originUrl, value.origin_url,
+            value.masterUrl, value.master_url,
+          ];
+          for (const candidate of candidates) {
+            const src = cleanImageUrl(candidate);
+            if (src) return src;
+          }
+          const nestedLists = [value.infoList, value.info_list, value.urlList, value.url_list].filter(Array.isArray);
+          for (const list of nestedLists) {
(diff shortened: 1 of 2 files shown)


jackwener:
这个 PR 的目标已经通过 [#1687](https://github.com/jackwener/OpenCLI/pull/1687)（已合入）实现 —— 方案是从 `__INITIAL_STATE__.imageList` 中提取 carousel 顺序保持轮播图序。当前 `main` 已覆盖此功能。

关闭此 draft PR。

PR #1518, opened 2026-05-13
fix(browser): drop session injection from extension exec results

hansnow:
## Summary

Fixes browser-backed searches that can visually render results in Chrome but return empty / `NOT_FOUND` from adapters because `pageScopedResult()` in the extension was injecting the lease's `session` into the result `data`. For the `exec` action this contaminated arbitrary user-JavaScript returns:

- Array / primitive returns came back as `{ session, data: <value> }` envelopes — `Array.isArray(result)` was false, adapters extracted nothing.
- Plain-object returns had an extra `session` key spliced in (and any user `session` field was silently overwritten by the lease).

Originally proposed (this PR's first commit) as a client-side unwrap in `Page.evaluate()`. After upstream review the scope changed to fix the root cause in the extension and drop the client-side workaround.

This PR now:

- Reverts `pageScopedResult()` to its pre-#1461 form (`{ id, ok, data, page }`). No session injection. `Result.session` is not added to the protocol — no consumers exist and `Result.page` already exposes routing identity. Bumps extension to **1.0.14**.
- Waits for `#rso a h3` (5s) before extracting Google SERP, falling back to the existing `wait 2`. Avoids empty extraction on Chrome 148 / Linux Wayland when DOM settles before SERP anchors populate.
- Extracts visible Xiaohongshu cards before scrolling, then merges post-scroll rows by URL. Xiaohongshu's virtualized masonry can evict initial note cards from the DOM after scroll.

Original commit attribution kept on @hansnow; scope rewritten with maintainer push.

## Repro environment (from original report)

- OpenCLI: `1.7.18`
- Browser Bridge extension: `1.0.12` → `1.0.14`
- Google Chrome: `148.0.7778.96`
- OS/session: Linux `7.0.0-15-generic` x86_64, Wayland
- Node.js: `22.22.1`, npm: `9.2.0`

Observed failure mode:

- `opencli google search "美食" --limit 5 --lang zh -f yaml`
- `opencli xiaohongshu search "美食" --limit 5 -f yaml`

Chrome showed normal public search results, but adapters received non-array values and treated the extraction as empty.

## Validation

- `npm run typecheck`
- `npm run build`
- `npx vitest run --project unit` — 1083 passed (one unrelated `EADDRINUSE` from daemon.test.ts port contention on local machine)
- `npx vitest run --project extension` — 64 passed
- `npx vitest run --project adapter clis/xiaohongshu/ clis/google/` — 113 passed
- Extension `vite build` — 75.49 kB

## Root cause history

The session injection in `pageScopedResult` was added by #1461 (`refactor(browser): replace workspaces with sessions`). The same refactor also caused the doctor connectivity regression fixed in 1.7.18 — both were silent breakages from a partial refactor that wasn't end-to-end smoke-tested. Lesson recorded for future large refactors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+* **browser** — `page.evaluate()` / `evaluateInFrame()` now return the user JavaScript value directly. Browser Bridge `exec` previously routed through a shared `pageScopedResult` helper that spread / wrapped the lease's `session` into the result `data`, contaminating arbitrary user returns: array / primitive returns came back as `{ session, data }` envelopes, and plain-object returns had an extra `session` key injected (overwriting any user `session` field). `google search` and `xiaohongshu search` were the visible repro — Chrome rendered results correctly but adapters extracted an empty array. Fixed in extension 1.0.14 by reverting `pageScopedResult` to its pre-1461 form (`{ id, ok, data, page }`); no client-side unwrap is needed.
+* **google/search** — wait for `#rso a h3` before extracting, falling back to the existing fixed wait. On Chrome 148 + Linux Wayland the DOM can settle before SERP anchors are populated, making extraction return empty even with the envelope bug fixed.
+* **xiaohongshu/search** — extract initially visible cards before scrolling, then merge post-scroll rows by URL. Xiaohongshu's virtualized masonry layout can evict the initial cards from the DOM after scroll, so the previous always-scroll-then-extract flow could lose the top results.
+
 ### Features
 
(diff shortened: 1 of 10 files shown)


jackwener (approved):
LGTM. Verified the extension-side fix keeps exec/evaluate data raw by reverting pageScopedResult to { id, ok, data, page }, with no Page-side unwrap residue. Extension 1.0.14 version files and dist are synced; Google/Xiaohongshu adapter fixes are scoped and reasonable. Local checks: extension background 52/52, Page unit 21/21, Xiaohongshu adapter 19/19, root typecheck/build, extension typecheck/build, diff-check.

PR #1524, opened 2026-05-13
chore: drop util.styleText to support Node v20+

jackwener:
## Summary

- Drop `util.styleText` from logger / output / doctor / tui / update-check / cli / download/progress / commands/daemon
- `engines.node`: `>=21.0.0` → `>=20.0.0`
- CHANGELOG + troubleshooting doc updated

## Why

A user reported on Node `v21.2.0`:
```
import { styleText } from 'node:util';
         ^^^^^^^^^
SyntaxError: The requested module 'node:util' does not provide an export named 'styleText'
```

`util.styleText` was added in **Node v21.7.0 / v20.12.0**. Anything in `v20.0–v20.11` or `v21.0–v21.6` throws at module load before any user code runs.

`npm engines` can't express "v20.12+ OR v21.7+" with a single semver, so just bumping `>=20.12.0` would still let `v21.0–v21.6` install and crash. A polyfill module would solve runtime, but the deeper question: **why do we have `styleText` at all?**

OpenCLI is primarily an agent-facing CLI. Agents see ANSI escapes as noise. The `[OK]` / `[WARN]` / `[FAIL]` / `ℹ` / `⚠` / `✖` markers already carry the semantic info — colors were redundant emphasis for humans on TTY. No user has ever reported missing colors as a problem.

So the fix is delete, not polyfill. ~150 `styleText(...)` calls → just the second arg; 8 imports gone; one AST-based codemod followed by manual cleanup of resulting `${'literal'}` and `${`...`}` artifacts.

## Validation

- `npm run typecheck` clean
- `npm run build` clean (815 manifest entries unchanged)
- Targeted unit tests: 199/199 pass (`output.test.ts` `doctor.test.ts` `cli.test.ts` `commands/daemon.test.ts` `help.test.ts`)
- Smoke: `opencli --help`, `opencli list`, `opencli doctor`, `opencli daemon status` all render correctly without colors
- `grep -rn "styleText\|node:util" src/` returns 0 results

## Test plan

- [x] CI: typecheck / unit / build / extension build / audit / doc
- [x] Local smoke (doctor, list, daemon status, --help)
- [ ] Reviewer: verify removed colors don't degrade any user-visible flow they care about (TUI checkbox picker, doctor output, plugin install/uninstall flow)


jackwener:
Blocking review finding: this PR does not actually make OpenCLI start on the reported Node range yet.

1. The runtime guard still rejects Node 20. After building this PR, `npx -y node@20.0.0 dist/src/main.js --help` exits with `OpenCLI requires Node.js >= 21.0.0.` So `package.json` says `>=20.0.0`, but the executable still enforces the old floor.

2. Node 21.2 still crashes at startup after `styleText` is removed, now through `undici`. Repro on this PR after `npm run build`: `npx -y node@21.2.0 dist/src/main.js --help` throws `TypeError: webidl.util.markAsUncloneable is not a function` from `node_modules/undici/lib/web/cache/cachestorage.js`. Current installed `undici@8.1.0` declares `engines.node >=22.19.0`, and `src/main.ts` imports `./node-network.js` during normal startup, so this is still a startup crash for the exact user class.

3. `package-lock.json` root metadata still says `engines.node: >=21.0.0`, so the engine change is incomplete even before dependency compatibility.

Suggested fix: either keep the Node floor aligned with the real runtime dependency floor, or actually support the lower range by changing the runtime guard and pinning/downgrading/removing `undici` to a version that supports it. `undici@6.25.0` declares `node >=18.17`; if you choose that route, verify the `EnvHttpProxyAgent` path still works. Then smoke both `node@20.0.0 dist/src/main.js --help` and `node@21.2.0 dist/src/main.js --help`.

jackwener:
Phase 5 re-review LGTM.

Previous blockers are resolved:
- `MIN_SUPPORTED_NODE_MAJOR` is now 20 and `runtime-detect.test.ts` covers `v20.0.0` true / `v19.9.0` false.
- `package.json` uses `engines.node: >=20.0.0` and `undici: ^6.25.0`; `package-lock.json` root metadata matches `>=20.0.0` and resolves runtime `node_modules/undici` to `6.25.0`.
- `undici@6.25.0` exports the APIs `node-network.ts` uses: `Agent`, `EnvHttpProxyAgent`, `fetch`, and `Dispatcher` type surface is still accepted by typecheck.
- `styleText` / runtime `node:util` style dependency is gone from `src/`.

Local verification:
- `src/runtime-detect.test.ts src/node-network.test.ts` 14/14.
- broader targeted set `output/doctor/cli/daemon/help/runtime-detect/node-network` 213/213.
- `npm run typecheck` + `npm run build` pass.
- clean-HOME smoke: `npx -y node@20.0.0 dist/src/main.js --help` exits 0; `npx -y node@21.2.0 dist/src/main.js --help` exits 0; Node 21.2 `doctor` exits 0. The remaining `UNDICI-EHPA` warning is from `EnvHttpProxyAgent` and does not block startup.

No further blockers from me.

PR #1593, opened 2026-05-15
fix(daemon): differentiate multi-profile state in status output (#1575)

### LINKED ISSUE #1575: daemon status: `Extension: disconnected` is misleading when multiple profiles are connected without a default
## Summary

When two or more Chrome/Chromium profiles each have the OpenCLI Browser Bridge extension installed and are running concurrently with no default selected, `opencli daemon status` prints `Extension: disconnected` even though both extensions are actively connected. This is a common setup (work + personal profile, or testing across browser channels), and the "disconnected" wording sends users down the wrong debug path — the actual fix is `opencli profile use <name>`, not reinstalling the extension.

## Reproduce

Connect two browser profiles with the OpenCLI Browser Bridge extension installed and active in both. Don't set `OPENCLI_PROFILE` and don't `opencli profile use <name>`:

```
$ opencli daemon status
Daemon: running (PID ...)
Version: v1.7.19
Extension: disconnected           # ← misleading
Profiles: <id-a> v1.0.x, <id-b> v1.0.x   # both are actually connected
```

## Source trace

`daemon.js:55-79` `resolveExtensionConnection()`:

- 0 profiles → `errorCode: 'extension_not_connected'` → status display `disconnected` ✓ accurate
- 1 profile → returns connection ✓ accurate
- **2+ profiles + no default → `errorCode: 'profile_required'` → `route.connection` is undefined → `extensionConnected: !!route.connection` is `false` → status display `disconnected`** ⚠️ misleading

The status endpoint at `daemon.js:200-220` already carries the richer state in JSON (`profileRequired`, `profileDisconnected`, `profiles[]`), but the human-readable CLI formatter collapses all three "no route" states into one `disconnected` label.

## Suggested fix

Differentiate the three states in the CLI formatter, e.g.:

```
Extension: 2 profiles connected, none selected — run `opencli profile use <name>`
Profiles: <id-a> v1.0.x, <id-b> v1.0.x
```

or split into explicit states `connected / disconnected / ambiguous (n profiles)`.

The JSON payload doesn't need to change.

## Why this matters

A user with two browser channels reads "Extension: disconnected" and starts reinstalling extensions, restarting Chrome, restarting the daemon. The one-command fix (`opencli profile use <name>`) isn't surfaced anywhere in the misleading status output. Adding a route-aware hint here is a low-risk papercut fix.

Happy to draft the formatter patch if useful.

## Versions

- opencli v1.7.19
- Browser Bridge extension v1.0.14
- macOS 26.x

Chen17-sq:
## Linked issue

Closes #1575.

## Summary

`opencli daemon status` rendered the line `Extension: disconnected` for three structurally different states:

1. zero profiles connected — accurate "disconnected"
2. **2+ profiles connected, no default selected** — misleading
3. requested profile vanished mid-session — misleading

Users hitting case 2 (a common setup: work + personal Chrome profiles both running with the OpenCLI extension installed) read "disconnected" and start reinstalling extensions / restarting Chrome / restarting the daemon — when the actual one-command fix is `opencli profile use <name>`.

The daemon JSON has carried `profileRequired` / `profileDisconnected` flags for a while (see `daemon.ts:259-260`). This PR teaches the CLI formatter to mirror that distinction.

## Implementation

`src/commands/daemon.ts`:

- Extracted `formatExtensionLabel(status: DaemonStatus): string` (exported for direct unit testing).
- Renders one of four states:
  - `connected (vX.Y.Z)` — single profile selected, extension OK (unchanged).
  - `N profile(s) connected, none selected — run \`opencli profile use <name>\`` — covers the 2+ profile no-default case AND the single-orphan-profile case where the user hasn't picked one yet. Grammar adjusts on count.
  - `requested profile not connected — open that Chrome profile, or pick another with \`opencli profile use <name>\`` — covers `profileDisconnected`.
  - `disconnected` — true zero-profile case (unchanged).

Imported `DaemonStatus` type from `daemon-client.js` to type the helper signature.

The JSON payload does not change. The CLI text is the only user-visible surface that needed the three-state distinction.

## Why this matters

A new user with a work + personal Chrome profile setup runs `opencli daemon status`, sees "Extension: disconnected", and goes down the reinstall-everything debug path. With this change they instead see:

```
Extension: 2 profiles connected, none selected — run `opencli profile use <name>`
Profiles: work v1.0.14, personal v1.0.14
```

— which is both diagnostic AND actionable in a single line.

## Test plan

Added two `describe` blocks to `src/commands/daemon.test.ts`:

- `formatExtensionLabel — multi-profile state differentiation` — covers all four label branches plus singular/plural grammar at the 1-profile-no-default edge.
- `daemonStatus output for issue #1575 fix` — asserts the end-to-end formatter never emits the bare `Extension: disconnected` line in the 2-profiles-no-default case, and that the `Profiles:` follow-up line still lists both profile names.

### Local test run

```
npx vitest run src/commands/daemon.test.ts
 Test Files  1 passed (1)
      Tests  16 passed (16)
```

5 new tests; 11 pre-existing tests still pass.

## Checklist

- [x] CLI text only — no breaking changes to JSON payload or to `getDaemonHealth()` state machine.
- [x] Singular/plural grammar handled.
- [x] Existing tests pass.
- [x] New regression tests added.

### DIFF
--- a/src/commands/daemon.test.ts
+++ b/src/commands/daemon.test.ts
@@ -104,6 +104,79 @@ describe('daemonStatus', () => {
   });
 });
 
+// ────────────────────────────────────────────────────────────────────
+// GH #1575: differentiate the three "no route" states. The pre-fix
+// behaviour collapsed multi-profile-no-default + profile-disconnected
+// + zero-profile all to "Extension: disconnected", sending users on
+// reinstall-everything debug paths when the actual fix was
+// `opencli profile use <name>`.
+// ────────────────────────────────────────────────────────────────────
+
+describe('daemonStatus extension label states (#1575)', () => {
+  let stdoutSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    stdoutSpy = vi.spyOn(console, 'log').mockImplementation(() => undefined);
+    fetchDaemonStatusMock.mockReset();
+  });
+  afterEach(() => vi.restoreAllMocks());
+
+  async function extensionLineFor(extra: Record<string, unknown>): Promise<string | undefined> {
+    fetchDaemonStatusMock.mockResolvedValue({
+      ok: true,
+      pid: 99,
+      uptime: 60,
+      daemonVersion: PKG_VERSION,
+      pending: 0,
+      memoryMB: 32,
+      port: 19825,
+      ...extra,
+    });
+    await daemonStatus();
+    return stdoutSpy.mock.calls
+      .map((c: unknown[]) => c[0])
+      .find((line: unknown): line is string =>
+        typeof line === 'string' && line.startsWith('Extension:'));
+  }
+
+  it('prints a route hint for 2+ profiles with no default (not bare "disconnected")', async () => {
+    const line = await extensionLineFor({
+      extensionConnected: false,
+      profileRequired: true,
+      profiles: [
+        { contextId: 'work', extensionConnected: true, pending: 0 },
+        { contextId: 'personal', extensionConnected: true, pending: 0 },
+      ],
+    });
+    expect(line).not.toBe('Extension: disconnected');
+    expect(line).toContain('2 profiles connected');
+    expect(line).toContain('opencli profile use');
+  });
+
+  it('defensively uses singular grammar for a one-profile profile-required payload', async () => {
+    const line = await extensionLineFor({
+      extensionConnected: false,
+      profileRequired: true,
+      profiles: [{ contextId: 'work', extensionConnected: true, pending: 0 }],
+    });
+    expect(line).toContain('1 profile connected');
+  });
+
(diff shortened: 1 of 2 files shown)


Chen17-sq:
Self-review: tightened the patch.

- Inlined the label logic back into `daemonStatus()` — no more exported `formatExtensionLabel` helper. The function was 8 lines; pulling it out as a named API was over-engineering for what's effectively a single 4-branch `if/else`.
- Dropped one of the two test `describe` blocks. All four state branches are still covered, but only via the public `daemonStatus()` output (which is what users actually see).
- Simplified the `profileDisconnected` message to match the `profileRequired` voice (`run \`opencli profile use <name>\`` in both cases).

Diff went from +148/-7 to +91/-3. Test count went from 16 to 14 (4 state-coverage tests + 10 pre-existing).

PR #1653, opened 2026-05-18
fix(extension): cleanup empty owned tab groups

### LINKED ISSUE #1627: feat: auto-remove empty tab groups after --site-session ephemeral cleanup
## Problem

When using `--site-session ephemeral`, tabs are correctly auto-closed after the command completes. However, the Chrome tab group (named 'OpenCLI Browser') persists as an empty group.

This clutters the tab strip with residual empty groups that require manual cleanup (right-click → Ungroup).

## Steps to Reproduce

```bash
opencli xiaohongshu search test --window background --site-session ephemeral -f json
```

- Tab opens ✓
- Tab auto-closes after command ✓
- Empty 'OpenCLI Browser' tab group remains ✗

## Expected Behavior

After ephemeral tab cleanup, any empty tab groups created by OpenCLI should also be removed automatically.

## Environment

- OpenCLI: 1.7.22
- Chrome: 148.0.7778.168 (arm64, macOS)
- Extension: 1.0.15

## Suggestion

After closing ephemeral tabs, check if the parent tab group is empty and remove it via `chrome.tabGroups` API from the extension context.

jun0315:
## Description

  Fixes #1627.

  When an owned OpenCLI tab is removed, Chrome can leave behind an empty named tab group such as `OpenCLI
  Adapter` / `OpenCLI Browser`. This is visible after ephemeral adapter sessions clean up their temporary tabs.

  This PR adds a small owned-tab removal helper that checks whether the tab being removed is the last tab in an
  OpenCLI-managed tab group. If so, it calls `chrome.tabs.ungroup(tabId)` before `chrome.tabs.remove(tabId)`,
  preventing Chrome from leaving an empty group behind.

  The default grouping behavior is unchanged. OpenCLI still groups owned browser/adapter tabs as before; this only
  cleans up the final tab before removal.

  Related issue: #1627

  ## Type of Change

  - [x] 🐛 Bug fix
  - [ ] ✨ New feature
  - [ ] 🌐 New site adapter
  - [ ] 📝 Documentation
  - [ ] ♻️ Refactor
  - [ ] 🔧 CI / build / tooling

  ## Checklist

  ## Screenshots / Output

  ```bash
  npx vitest run extension/src/background.test.ts
  # Test Files  1 passed (1)
  # Tests  61 passed (61)

  npx vitest run --project extension
  # Test Files  2 passed (2)
  # Tests  73 passed (73)

  cd extension && npm run typecheck
  # passed

  npm run typecheck
  # passed

  cd extension && npm run build
  # passed

  git diff --check
  # passed

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -1113,6 +1113,26 @@ async function ensureOwnedContainerTabGroup(role, windowId, tabIds) {
     console.warn(`[opencli] Failed to mark ${role} tab group: ${err instanceof Error ? err.message : String(err)}`);
   }
 }
+async function ungroupOwnedTabIfLastInGroup(role, windowId, tabId) {
+  try {
+    const tab = await chrome.tabs.get(tabId);
+    const groupId = tab.groupId;
+    if (groupId === void 0 || groupId === chrome.tabGroups.TAB_GROUP_ID_NONE) return;
+    const group = await chrome.tabGroups.get(groupId);
+    if (group.windowId !== windowId) return;
+    if (!getOwnedContainerGroupTitles(role).includes(group.title ?? "")) return;
+    const tabs = await chrome.tabs.query({ windowId });
+    const groupedTabs = tabs.filter((entry) => entry.groupId === groupId);
+    if (groupedTabs.length > 1) return;
+    await chrome.tabs.ungroup(tabId);
+    if (ownedContainers[role].groupId === groupId) ownedContainers[role].groupId = null;
+  } catch {
+  }
+}
+async function removeOwnedContainerTab(role, windowId, tabId) {
+  await ungroupOwnedTabIfLastInGroup(role, windowId, tabId);
+  await chrome.tabs.remove(tabId);
+}
 async function ensureOwnedContainerWindow(role, initialUrl, mode = "background") {
   const container = ownedContainers[role];
   if (container.promise) return container.promise;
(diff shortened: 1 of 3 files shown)


jackwener:
Thanks for digging into this @jun0315 — closing this PR after a deeper design review.

The change works as written (and the test additions are solid), but the direction conflicts with how OpenCLI's owned tab groups are designed to work:

**The empty `OpenCLI Browser` / `OpenCLI Adapter` group is a feature, not a leak.**

OpenCLI uses these named groups as *persistent automation container markers*. When a tab is removed, the group is intentionally left behind as a stable anchor for the next session. The existing reuse path (PR #1541's `discoverOwnedContainerFromTabGroup` + `ensureOwnedContainerTabGroup`) finds the existing empty group on the next `--site-session ephemeral` run and places the new tab back into it — verified just now via a temporary unit test on `current main`: with an empty `OpenCLI Adapter` group present and no reusable tab, the next `resolveTabId(...)` reuses the group and adds the new tab to it (no second group created).

So users see exactly **one** empty OpenCLI group between sessions, never accumulation. Auto-ungrouping the last tab would actively break that reuse path — Chrome would then create a fresh group with a new ID each session.

If you'd like to take another swing at issue #1627, the right direction would be to verify whether there's a scenario where reuse *doesn't* trigger (e.g. cross-window edge case, race during teardown), and fix that path — rather than removing the group anchor. Happy to look at a fresh PR going that direction.

Related design: PR #1541 (cross-window owned-container discovery), PR #1566 (regression test for groupId tiebreaker).

Issue #847, opened 2026-04-07
代码安全审核的代码结果


jglwiz:
High 级别发现

  1. operate eval - 任意 JavaScript 执行

  - 文件: src/cli.ts:498-503
  - 描述: 接受用户输入的任意 JS 并在浏览器页面上下文中执行, 无任何过滤或沙箱. 可窃取 cookies、读取敏感 DOM 内容.
  - 建议: 虽然是设计功能, 但应添加 --allow-eval 安全开关并在文档中明确风险.

  2. Plugin 安装 - 不可信代码执行

  - 文件: src/plugin.ts:570-601
  - 描述: 从 GitHub 克隆插件后立即执行 npm install + 动态加载. 插件的 package.json 可包含 postinstall 脚本实现任意代码执行.
  - 建议: 使用 npm install --ignore-scripts, 并显示安全警告.

  ---
  Medium 级别发现

  3 & 4. operate get text/value/html/attributes 及 operate type - DOM 注入

  - 文件: src/cli.ts:381-402, src/cli.ts:420-429
  - 描述: index 参数直接拼接进 JS 模板字符串, 无转义:
  document.querySelector('[data-opencli-ref="${index}"]')
  - 恶意 index 如 "]') || eval(...)// 可注入执行.
  - 建议: 使用 JSON.stringify() 转义 (同文件中 operate select 在 line 446 已正确使用了此方式).

  5. 网络拦截器 - 敏感数据捕获

  - 文件: src/cli.ts:311, src/interceptor.ts
  - 描述: 拦截所有 fetch/XHR 响应体 (最大 200 条, 每条 50KB), 可能包含 auth tokens、PII、会话密钥.
  - 建议: 默认过滤 auth/login/token 相关端点, 添加敏感数据警告.

  ---
  Low 级别发现

  ┌─────┬───────────────────────┬───────────────────────────────┬────────────────────────────────────────────────────────────────┐
  │  #  │         问题          │             文件              │                              说明                              │
  ├─────┼───────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────────────┤
  │ 6   │ 命令注入 (有部分防护) │ src/external.ts:100-120       │ shell 操作符正则未覆盖所有情况, 但 execFileSync 提供了额外保护 │
  ├─────┼───────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────────────┤
[...]

PR #1754, opened 2026-05-26
fix(plugin): pass --ignore-scripts to npm install for cloned plugin repos

### LINKED ISSUE #1753: [Bug]: plugin install runs untrusted npm lifecycle scripts (postinstall) without --ignore-scripts
### Description

`installDependencies` in `src/plugin.ts:567` runs `execFileSync('npm', ['install', '--omit=dev'], ...)` against a plugin repo that was just cloned from a third-party Git URL via `opencli adapter install`. Without `--ignore-scripts`, npm executes `preinstall`, `install`, and `postinstall` lifecycle scripts declared in the plugin's `package.json` (and in every transitive dep), so a malicious plugin or a compromised dep can run arbitrary code with the user's privileges the moment the user installs the plugin. The actual adapter code is loaded later by the plugin discovery path, which is the expected attack surface; the install-time script execution is an extra vector that does not need to exist for adapter plugins to work.

This is Finding 2 in #847. @Astro-Han ranked it the top-priority unfixed item in [the triage comment](https://github.com/jackwener/OpenCLI/issues/847#issuecomment-4197297296):

> `plugin install` is a real issue, and one we should treat seriously. [...] Running `npm install` on an untrusted repo is the extra risk here because of lifecycle scripts. I think this one is valid. Using `--ignore-scripts` plus a clear warning would be a good first step.

### Steps to Reproduce

1. Create a plugin repo whose `package.json` contains a `postinstall` script that writes a marker, e.g. `"scripts": { "postinstall": "node -e \"require('fs').writeFileSync('/tmp/opencli-postinstall-marker', '1')\"" }`
2. Push it to a Git host
3. Run `opencli adapter install <repo-url>`
4. Check `ls -l /tmp/opencli-postinstall-marker`

### Expected Behavior

Install completes without running the plugin's `postinstall` script. The marker file is not created. Equivalent to the npm invocation passing `--ignore-scripts`.

The fix is a one-flag change at the install call:

```ts
execFileSync('npm', ['install', '--omit=dev', '--ignore-scripts'], { ... });
```

### OpenCLI Version

1.8.0

### Node.js Version

Other

### Operating System

macOS

### Logs / Screenshots

Constructed illustration of the repro, not a captured run:

```
$ cat malicious-plugin/package.json
{ "name": "evil", "scripts": { "postinstall": "node -e \"require('fs').writeFileSync('/tmp/opencli-postinstall-marker', '1')\"" } }

$ opencli adapter install https://github.com/me/malicious-plugin
$ ls -l /tmp/opencli-postinstall-marker
-rw-r--r--  1 me  staff  1 May 26 ...
```

Benjamin-eecs:
## Description

`installDependencies` in `src/plugin.ts` runs `execFileSync('npm', ['install', '--omit=dev'], ...)` against a directory just cloned from a third-party Git URL via `opencli adapter install`. Without `--ignore-scripts`, npm executes preinstall / install / postinstall hooks declared in the plugin's `package.json` (and in every transitive dep), so a malicious plugin or a compromised dep can run arbitrary code with the user's privileges at install time. Adapter code is loaded later by the plugin discovery path; the install-time script execution is an extra vector that adapter plugins do not need.

This is Finding 2 in the #847 security audit. @Astro-Han ranked it the top-priority unfixed item in his triage: *"`plugin install` is a real issue, and one we should treat seriously … Using `--ignore-scripts` plus a clear warning would be a good first step."* This PR is the `--ignore-scripts` half; a follow-up could add a one-time advisory if useful.

Plugins that depend on install scripts to compile native modules will silently skip those hooks. If a real case appears, an opt-in such as a manifest field or env var can be added in a follow-up; opencli adapter plugins observed in the wild depend on small pure-JS packages and do not exercise lifecycle scripts.

Related issue: fixes #1753.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Existing test `throws when npm install fails` still passes (mock-based regression check at the install error path). New test `passes --ignore-scripts to npm install` asserts the flag reaches the `execFileSync` call via the existing `mockExecFileSync` infrastructure. Full `src/plugin*.test.ts` suite: 105/105 pass on this branch.

### DIFF
--- a/src/plugin.test.ts
+++ b/src/plugin.test.ts
@@ -613,6 +613,21 @@ describe('installDependencies', () => {
 
     fs.rmSync(tmpDir, { recursive: true, force: true });
   });
+
+  it('passes --ignore-scripts to npm install', () => {
+    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'opencli-plugin-ok-'));
+    fs.writeFileSync(path.join(tmpDir, 'package.json'), JSON.stringify({ name: 'plugin-ok' }));
+
+    _installDependencies(tmpDir);
+
+    expect(mockExecFileSync).toHaveBeenCalledWith(
+      'npm',
+      ['install', '--omit=dev', '--ignore-scripts'],
+      expect.objectContaining({ cwd: tmpDir }),
+    );
+
+    fs.rmSync(tmpDir, { recursive: true, force: true });
+  });
 });
 
 describe('postInstallMonorepoLifecycle', () => {
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -564,7 +564,9 @@ function installDependencies(dir: string): void {
   if (!fs.existsSync(pkgJsonPath)) return;
 
   try {
-    execFileSync('npm', ['install', '--omit=dev'], {
+    // --ignore-scripts: dir was just cloned from a third-party Git URL, so
+    // running its npm lifecycle hooks would execute arbitrary code as the user.
+    execFileSync('npm', ['install', '--omit=dev', '--ignore-scripts'], {
       cwd: dir,
       encoding: 'utf-8',
       stdio: ['pipe', 'pipe', 'pipe'],


Benjamin-eecs:
Superseded: `--ignore-scripts` landed in #2125 (654019ee), so this branch no longer carries a change. Closing.

PR #1762, opened 2026-05-26
fix(extension): skip user windows in tab-group discovery

### LINKED ISSUE #1760: [Bug] Empty tab group residue causes resolveTab to reuse user's existing tabs instead of creating new automation window
### Description

On Windows 11, after running a browser adapter command, an "OpenCLI Adapter" tab group is left behind in the user's Chrome window. On subsequent runs, this residual tab group causes the extension to adopt the user's window as an automation container, leading to user tabs being reused/replaced.

### Root Cause

**`discoverOwnedContainerFromTabGroup()`** (~line 582 in `background.ts`) searches for tab groups named "OpenCLI Adapter" across **all windows**, not just owned automation windows. When it finds a residual tab group in the user's window, it adopts that window as the automation container.

Then **`resolveTab()`** (~line 1132) falls through to **`findReusableOwnedContainerTab()`**, which searches the adopted window for any debuggable tab — including the user's active `https://` tabs. This replaces the user's tab content instead of creating a new tab.

### Reproduction

1. Run any browser adapter command (e.g., a command that uses `browser: true`)
2. After completion, notice an "OpenCLI Adapter" tab group remains in your Chrome window
3. Make sure your active tab is the rightmost tab
4. Run another browser adapter command
5. Observe: your active tab is replaced by the automation tab instead of a new tab being created

### Expected Behavior

- Tab groups should only exist in owned automation windows, not user windows
- `discoverOwnedContainerFromTabGroup()` should only look in owned windows, or at minimum verify the window is an automation window before adopting it
- `findReusableOwnedContainerTab()` should never reuse user tabs (non-blank, non-extension tabs)

### Environment

- Windows 11 Home (10.0.26200)
- Chrome latest
- OpenCLI latest

### Suggested Fix

In `discoverOwnedContainerFromTabGroup()`, after finding a tab group, verify the parent window is an owned automation window (check `session.owned` or similar flag) before adopting it. If it's a user window, clean up the tab group instead.

---

### 问题描述

在 Windows 11 上，运行浏览器型适配器命令后，用户 Chrome 窗口中会残留一个名为 "OpenCLI Adapter" 的标签页分组。再次运行命令时，该残留的标签页分组会导致扩展将用户窗口当作自动化容器窗口，从而复用/替换用户的标签页。

### 根本原因

**`discoverOwnedContainerFromTabGroup()`**（`background.ts` 约第 582 行）在**所有窗口**中搜索名为 "OpenCLI Adapter" 的标签页分组，而不仅仅在已拥有的自动化窗口中搜索。当它在用户窗口中发现残留的标签页分组时，会将该窗口采纳为自动化容器。

随后 **`resolveTab()`**（约第 1132 行）会进入 **`findReusableOwnedContainerTab()`**，该函数在被采纳的窗口中搜索任何可调试的标签页——包括用户正在使用的 `https://` 标签页。这导致用户标签页的内容被替换，而非创建新标签页。

### 复现步骤

1. 运行任意浏览器型适配器命令（使用 `browser: true` 的命令）
2. 命令完成后，观察 Chrome 窗口中残留的 "OpenCLI Adapter" 标签页分组
3. 确保当前激活的标签页是最右侧的标签页
4. 再次运行浏览器型适配器命令
5. 观察：当前激活的标签页被自动化标签页替换，而非创建新标签页

### 期望行为

- 标签页分组应仅存在于已拥有的自动化窗口中，而非用户窗口
- `discoverOwnedContainerFromTabGroup()` 应只在已拥有的窗口中查找，或至少在采纳前验证该窗口是否为自动化窗口
- `findReusableOwnedContainerTab()` 不应复用用户标签页（非空白、非扩展标签页）

### 建议修复

在 `discoverOwnedContainerFromTabGroup()` 中，找到标签页分组后，验证其父窗口是否为已拥有的自动化窗口（检查 `session.owned` 或类似标志），而非用户窗口。如果是用户窗口，应清理该标签页分组而非采纳它。

Benjamin-eecs:
## Description

`discoverOwnedContainerFromTabGroup` queries Chrome for any tab group titled "OpenCLI Adapter" across all windows and adopts the first match as the automation container. When Chrome moves our tab group into the user's main window (via tab drag or "Merge all windows"), the residual group survives there after the original automation window closes. The next service-worker discovery cycle then adopts the user's window, and `findReusableOwnedContainerTab` happily returns the user's real http(s) tab as a reusable automation target. `resolveTab` proceeds to overwrite that tab.

Adapter tabs created by the extension are always added to our group via `ensureOwnedContainerTabGroup`, so any http(s) tab in the candidate window that is outside our group must be a user tab. Use that as the residue signal in `toOwnedContainerDiscoveryCandidate`: when present, return null so `selectOwnedContainerDiscoveryCandidate` does not adopt the window and a fresh automation window is created instead.

Related issue: fixes #1760.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Unit test `skips a residual OpenCLI Adapter group when its window holds user http tabs outside the group` covers the regression: a window with a user `https://` tab outside the matched group and an "OpenCLI Adapter" group inside it now triggers a fresh automation window via `chrome.windows.create` and the user's tab url is preserved.

Live test on Chrome for Testing 149 with the patched extension loaded: a probe driven from the service-worker debug WS prepared the residue state (user window with one `https://example.com/users-real-tab` tab outside the group and one `about:blank` tab inside an "OpenCLI Adapter" group) and the patched discovery returned `adopted: null` while the pre-patch logic would have selected the user's window. Full extension suite: 75/75 pass.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -1045,6 +1045,11 @@ async function focusOwnedWindowIfRequested(windowId, mode) {
 async function toOwnedContainerDiscoveryCandidate(group) {
   try {
     const chromeWindow = await chrome.windows.get(group.windowId);
+    const windowTabs = await chrome.tabs.query({ windowId: group.windowId });
+    const hasUserTabsOutsideGroup = windowTabs.some(
+      (tab) => tab.groupId !== group.id && !!tab.url && isSafeNavigationUrl(tab.url)
+    );
+    if (hasUserTabsOutsideGroup) return null;
     const reusableTabId = await findReusableOwnedContainerTab(group.windowId);
     return {
       windowId: group.windowId,
--- a/extension/src/background.test.ts
+++ b/extension/src/background.test.ts
@@ -1290,6 +1290,35 @@ describe('background tab isolation', () => {
     expect(chrome.tabGroups.update).not.toHaveBeenCalled();
   });
 
+  it('skips a residual OpenCLI Adapter group when its window holds user http tabs outside the group', async () => {
+    const { chrome, tabs, groups } = createChromeMock();
+    tabs.push({
+      id: 77,
+      windowId: 7,
+      url: 'https://example.com/users-real-tab',
+      title: 'users real tab',
+      active: true,
+      status: 'complete',
+      groupId: -1,
+    });
+    groups.push({
+      id: 99,
+      windowId: 7,
+      title: 'OpenCLI Adapter',
+      color: 'orange',
+      collapsed: true,
+    });
+    vi.stubGlobal('chrome', chrome);
+
+    const mod = await import('./background');
+    const tabId = await mod.__test__.resolveTabId(undefined, adapterKey('twitter'));
+
+    expect(chrome.windows.create).toHaveBeenCalledTimes(1);
+    expect(mod.__test__.getAutomationWindowId(adapterKey('twitter'))).not.toBe(7);
+    expect(tabs.find((tab) => tab.id === 77)?.url).toBe('https://example.com/users-real-tab');
+    expect(tabs.find((tab) => tab.id === tabId)?.windowId).not.toBe(7);
+  });
+
   it('prefers a focused OpenCLI Adapter group when multiple matching groups exist', async () => {
     const { chrome, tabs, groups, setLastFocusedWindowId } = createChromeMock();
     setLastFocusedWindowId(8);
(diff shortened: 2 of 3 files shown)


jackwener:
Closing this as obsolete after the v1.0.17 owned-tab-group rewrite (#1794).\n\nThe direction here is the wrong abstraction for current main: skipping any window that has user http(s) tabs outside an OpenCLI group would make OpenCLI abandon an existing owned group and create another one, which conflicts with the new invariant that Chrome group state is the source of truth and duplicates should converge into one canonical OpenCLI Adapter/Browser group.\n\nThe real remaining risk from #1760 should be handled against the new model by scoping reusable-tab selection to owned/group tabs (or another explicit ownership signal), not by refusing to adopt the canonical group because the same Chrome window also contains normal user tabs.

PR #1797, opened 2026-05-31
feat: conversation mgmt across 4 AI surfaces (grok/codex/antigravity/trae-solo) + grok i18n + Chromium 142 launcher

LeoLin990405:
Three changes bundled (kept together because the codex commits build on patterns introduced for grok):

## 1. fix(grok): locale-independent selectors

Two selectors in `clis/grok/utils.js` were keyed on English `aria-label`:
- `button[aria-label="Submit"]`         (composer send)
- `button[aria-label="Model select"]`   (model picker)

These miss in non-English locales (verified zh-CN: `aria-label="提交"` / `"模型选择"`).

Fix: prefer locale-independent anchors with localized aria-label fallbacks.
- Submit: `button[data-testid="chat-submit"]` → `"Submit"` / `"提交"` / `"送信"`
- Model:  `#model-select-trigger` → `"Model select"` / `"模型选择"` / `"モデル選択"`

## 2. feat(grok): add delete/pin/unpin

`opencli grok delete <id> --yes`, `opencli grok pin <id>`, `opencli grok unpin <id>`.

Right-clicking any `/c/<id>` sidebar link surfaces a radix context menu with 4 items (`Open in new tab` / `Rename` / `Pin` or `Unpin` / `Delete`). 3 of these are wired up. Implementation note: must dispatch the full pointer chain (`pointerdown` → `mousedown` → `contextmenu` → `pointerup` → `mouseup` with `button=2`); the bare `MouseEvent('contextmenu')` alone is ignored by radix.

Grok deletes WITHOUT a confirmation dialog, so `--yes` is required.

Known: `rename` left as a stub-with-error because the radix `onSelect` for the Rename menu item doesn't respond to programmatic clicks (synthetic chains all report `clicked:true` but the inline rename input never materializes). Needs CDP `Input.dispatchMouseEvent` at viewport coordinates — separate followup.

## 3. feat(codex): add pin/unpin/archive/rename

`opencli codex pin/unpin/archive/rename` for sidebar conversations.

All four use the SAME path: select the target chat (`--project / --conversation / --index / --thread-id`), then open the `button[aria-label="Chat actions"]` header dropdown (full pointer chain again — same radix issue), then click the matching menu item (`Pin chat` / `Unpin chat` / `Archive chat` / `Rename chat`).

Why the header menu and not per-row sidebar buttons? Codex's per-row Pin/Archive buttons are React-lazily-mounted only when the row is hovered AND `document.visibilityState === 'visible'`. When the user has the Codex window minimized, even programmatic `mouseenter` won't surface them. The Chat actions menu mounts on click, independent of visibility.

`archive` is dry-run by default (Codex applies it without UI confirmation); pass `--yes` to actually archive.

Known partial: `rename` returns `status:'renamed'` and the rename contenteditable DOES receive the new title, but the commit-on-Enter sometimes doesn't persist. Pin/Unpin/Archive are fully verified end-to-end.

## Verified live

Grok (zh-CN, free tier):
```
$ opencli grok ask 'reply with: working'
- response: 思考了 3s working

$ opencli grok status
- Model: Fast    # was null before

$ opencli grok pin/unpin/delete --yes <uuid>
- status: pinned / unpinned / deleted    # sidebar still_present=false after delete
```

Codex (macOS desktop, app://-/index.html):
```
$ opencli codex pin --project leo --index 2
- status: pinned

$ opencli codex archive --project leo --index 2 --yes
- status: archived
```

Tests pass: 29/29 grok, 14/14 codex.

## Related
- Pairs with #1789 (`--remote-allow-origins=*` for Chromium 142+ launchers).
- Addresses the conversation-management gap across all opencli AI surfaces.

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -1572,22 +1572,30 @@
   {
     "site": "antigravity",
     "name": "model",
-    "description": "Switch the active LLM model in Antigravity",
-    "access": "read",
-    "domain": "localhost",
+    "description": "Read or switch the active model in Antigravity. Without arguments, reports the current model. With <name> (substring, case-insensitive), switches.",
+    "access": "write",
+    "domain": "127.0.0.1",
     "strategy": "ui",
     "browser": true,
     "args": [
       {
         "name": "name",
         "type": "str",
-        "required": true,
+        "required": false,
         "positional": true,
-        "help": "Target model name (e.g. claude, gemini, o1)"
+        "help": "Substring (case-insensitive) of target model name. Omit to read current."
+      },
+      {
+        "name": "list",
+        "type": "boolean",
+        "default": false,
+        "required": false,
+        "help": "List models in the picker (does not switch)"
       }
     ],
     "columns": [
-      "Status"
+      "Status",
+      "Model"
     ],
     "type": "js",
     "modulePath": "antigravity/model.js",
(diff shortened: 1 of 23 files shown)


LeoLin990405:
Restructuring per-surface — splitting into 4 surface-specific PRs (grok / codex / antigravity / trae-solo) so each surface can review independently. Will link the 4 follow-up PRs here.

PR #1798, opened 2026-05-31
feat(grok): conversation mgmt (delete/pin/unpin) + locale-independent selectors

LeoLin990405:
Split from #1797. This PR contains all grok-adapter changes. Other surfaces (codex / antigravity / trae-solo) are in separate PRs.

See commit message for full details — bundled because both changes touch `clis/grok/utils.js`:

1. **fix**: composer Submit + Model picker selectors now use `data-testid` / stable id with i18n aria-label fallback (was broken in zh-CN locale).
2. **feat**: new `opencli grok delete/pin/unpin` commands using the radix sidebar context menu, with the full pointer chain needed to trigger radix events.

Verified live on zh-CN Chrome 148 / grok.com free tier. `rename` is stubbed (radix React onSelect intercept — separate followup).


LeoLin990405:
**Grok exhaustive coverage summary** (commits `05e2a5b5` + `0fd66ac6`)

Did the same deep-audit pass on Grok that landed in PR #1801 for Trae SOLO. Grok is now exhaustively wrapped at every layer reachable from outside its server-side API.

**Total: 24 commands across 3 layers**

| Layer | Commands |
|---|---|
| UI / CDP (chat + sidebar) | `ask` `send` `read` `detail` `history` `new` `delete` `pin` `unpin` `rename` `image` `status` `model` `copy-message` `regenerate` `react` `incognito` `account` `search-conversations` `share` (19, +1 missed: image-2) |
| Browser-side storage (analog to Trae's FS state) | `storage-keys` `storage-get` `user-settings` `cookies` `idb-list` (5) |

**What's NOT wrapped — and why:**

- **Server-side `/api/...` calls** — would use the user's session cookie to RPC into Grok's backend bypassing the UI. That's a TOS gray-area at minimum and breaks on any backend change. **Intentionally skipped.**
- **httpOnly auth cookies** — `cookies` shows JS-visible cookies only; session tokens are correctly httpOnly and we don't try to extract them from CDP. That's a security feature, not a coverage gap.
- **AiTopia IndexedDB contents** — the only `idb-list` returns one suspicious `AiTopia` DB whose only row holds a 5KB base64 blob (keyed by a 16-byte hash). It looks like third-party extension data, not Grok-managed state. `idb-list` enumerates DBs but deliberately does not read contents.
- **Voice mode + Dictate (Ctrl+D)** — UI exists ([93] [94] in the earlier mapping) but audio input is low-value for a CLI agent.
- **Grammarly extension** — third-party, not Grok.
- **Projects panel** — empty on the test account; could add `project-list` / `project-new` later if someone with active projects requests it.

**Safety patterns:**

- `delete <id>` requires `--yes` (irreversible, no confirmation dialog from Grok)
- `share` is write-class because it creates a public URL; the implementation is verified by static review only — I deliberately did NOT run a live E2E because that would publish one of the user's real conversations. Caller should opt-in explicitly.
- `cookies` truncates each value preview to 40 chars to avoid leaking session tokens into logs even if Grok ever marks an auth cookie as non-httpOnly.
- All `storage-*` and `cookies` and `idb-list` are READ-ONLY by design (writing to Grok's localStorage from outside would race React state).

**E2E verified live on grok.com:** account name+email extraction, search-conversations with multi-language queries (`AI新闻` / `Taipei`), incognito navigation, model list (Fast/Auto/Expert/Heavy), copy-message returning actual assistant text, storage-keys with filter, idb-list, user-settings decomposition.

**Known limitation (documented in code):** `model --set` clicks the dropdown item successfully, but the selection doesn't persist across separate opencli invocations because each command opens a fresh tab. Useful inside a single-tab workflow.

Happy to split into smaller PRs if reviewer prefers — currently 3 commits, each a coherent unit.

PR #1815, opened 2026-06-01
fix(twitter): ignore timeline media after post submit

### LINKED ISSUE #1781: twitter post --images reports timeout failure but tweet is actually posted successfully
## Bug Description

`opencli twitter post --images <path> "<text>"` reports a failure with message **"Tweet submission did not complete before timeout."** even though the tweet (with image) is actually posted successfully.

## Steps to Reproduce

```bash
opencli twitter post "test tweet" --images /path/to/image.png --trace on
```

## Expected Behavior

The command should return `status: success` with the tweet URL and ID.

## Actual Behavior

Returns `status: failed` with message `"Tweet submission did not complete before timeout."` — but the tweet **is** live on the timeline with the image attached.

Verified by running `opencli twitter thread <id>` afterwards — the tweet exists with `has_media: true`.

## Root Cause Analysis

In `clis/twitter/post.js`, the `submitTweet()` function (line 145-196) polls after clicking the tweet button:

1. After a successful post, X navigates away from `/compose/post` back to the home timeline
2. The **composer textarea** (`tweetTextarea_0`) no longer contains the original text → `composerStillHasText` is `false`
3. However, `hasMedia` on line 188-189 matches `tweetPhoto` elements **in the timeline** (not just the composer), so `hasMedia` becomes `true`
4. Since `!composerStillHasText && !hasMedia` is never true (hasMedia is true from timeline), the loop keeps running until timeout

The fix should scope the `hasMedia` check to the composer context only, not the entire page. For example, query within a composer ancestor element, or check if we are still on `/compose/post` before evaluating media.

Alternatively, detecting the page navigation away from `/compose/post` to home (or finding the new tweet in the timeline matching the posted text) could serve as a success signal.

## Environment

- OpenCLI: v1.8.0
- OS: macOS 26.5
- Node.js: latest bundled with opencli
- Browser: Chromium via opencli built-in

gauravsaxena1997:
## Summary
- fixes image posts being reported as timed out after X navigates away from /compose/post
- only treats media as blocking submit completion while still on the compose route
- adds a regression check for timeline media not being counted as composer media

Fixes #1781.

## Test plan
- npm test -- clis/twitter/post.test.js
- npm run build

### DIFF
--- a/clis/twitter/post.js
+++ b/clis/twitter/post.js
@@ -235,8 +235,11 @@ async function submitTweet(page, text) {
 
             const boxes = Array.from(document.querySelectorAll('[data-testid="tweetTextarea_0"]')).filter(visible);
             const composerStillHasText = boxes.some((box) => normalize(box.innerText || box.textContent || '').includes(expectedText));
-            const hasMedia = !!document.querySelector('[data-testid="attachments"], [data-testid="tweetPhoto"]')
-                || document.querySelectorAll('img[src^="blob:"], video[src^="blob:"]').length > 0;
+            const isComposeRoute = /^\\/compose\\/post\\/?$/.test(window.location.pathname);
+            const hasMedia = isComposeRoute && (
+                !!document.querySelector('[data-testid="attachments"], [data-testid="tweetPhoto"]')
+                || document.querySelectorAll('img[src^="blob:"], video[src^="blob:"]').length > 0
+            );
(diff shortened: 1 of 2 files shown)


gauravsaxena1997:
Closing as superseded by current main. The equivalent image-timeout fix line is already present on upstream main, so it is not being republished as a new PR.

PR #1817, opened 2026-06-01
fix(linkedin): preserve multiline safe-send messages

gauravsaxena1997:
## Summary

This fixes LinkedIn `safe-send` flattening multiline messages before they reach the LinkedIn composer.

The command previously validated `--message` through `normalizeWhitespace(...)` and then reused that normalized value for sending. That made messages with paragraphs, numbered answers, or bullet lists arrive at the composer as a single line. The composer then had no chance to preserve user-intended formatting.

## Why this matters

`linkedin safe-send` is commonly used for careful human-approved outreach and replies. Those messages often need readable structure:

- short paragraphs
- numbered answers
- bullet lists
- screening-question responses

Flattening everything into one paragraph is hard to read and makes approved messages look less professional, especially for recruiter or lead replies.

## What changed

- Preserve the raw `--message` string after validating that it is non-empty.
- Keep existing normalized comparisons for safety checks.
- Add a LinkedIn-specific multiline composer fill path for messages containing `\n` or `\r`.
- For multiline content, try paste-style contenteditable insertion first, then fall back to DOM multiline insertion with `<br>` nodes and input events.
- Keep the previous `page.insertText(...)` path for single-line messages.
- Keep all fail-closed recipient/thread/latest-message checks unchanged.

## Safety model

This does not relax the existing write safety boundaries:

- dry-run remains the default
- exact thread URL is still required
- expected recipient is still verified
- optional latest-message text/hash guards still run before filling
- post-fill normalized composer text must still match the intended message
- send still requires explicit `--send true`

The main change is that validation no longer destroys formatting before insertion.

## Validation

- `npm run test:adapter -- clis/linkedin/safe-send.test.js`
- `npm run typecheck`
- `npm run test:adapter`

Full adapter suite result: 379 files passed, 3717 tests passed.

### DIFF
--- a/clis/linkedin/safe-send.js
+++ b/clis/linkedin/safe-send.js
@@ -53,6 +53,10 @@ function textContainsNormalized(haystack, needle) {
   return !n || h.includes(n);
 }
 
+function hasLineBreaks(value) {
+  return /\r|\n/.test(String(value ?? ''));
+}
+
 function selectBestHeaderName(headerNames, expectedName) {
   const expected = normalizeName(expectedName);
   const names = (Array.isArray(headerNames) ? headerNames : [])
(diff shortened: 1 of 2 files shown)


gauravsaxena1997:
Superseded by #1930, which republishes this LinkedIn safe-send fix on a clean branch from current main.

PR #1818, opened 2026-06-01
fix(twitter): fall back when quote composer card is missing

gauravsaxena1997:
## Summary
- keep the existing dedicated X quote composer path
- when the quoted card does not render from /compose/post?url=..., append the target tweet URL into the composer before posting
- add unit coverage for fallback quote text and the guarded fallback path

## Why
The X UI can load the quote composer text area without rendering the quoted tweet card. The current adapter fails in that state with "Quote target did not render in the composer", which blocks daily quote workflows. Appending the target tweet URL preserves quote behavior on X instead of failing or posting orphaned text.

## Tests
- npm run build
- npx vitest run --project adapter clis/twitter/quote.test.js clis/twitter/shared.test.js

### DIFF
--- a/clis/twitter/quote.js
+++ b/clis/twitter/quote.js
@@ -17,39 +17,87 @@ function buildQuoteComposerUrl(url) {
     return `https://x.com/compose/post?url=${encodeURIComponent(parsed.url)}`;
 }
 
-async function submitQuote(page, text, tweetId) {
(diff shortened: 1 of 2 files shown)


gauravsaxena1997:
Superseded by #1931, which consolidates the still-relevant Twitter quote fallback work on a clean branch from current main.

PR #1861, opened 2026-06-05
fix(daemon): SIGKILL fallback when stale daemon refuses graceful shutdown

jackwener:
## Summary

When the CLI detects a stale daemon (`daemonVersion !== PKG_VERSION` after `npm install -g @jackwener/opencli@latest`), it currently asks the daemon to exit via `POST /shutdown` and waits up to 3 seconds for the port to release. If the old daemon hangs, refuses `/shutdown`, or the endpoint is missing entirely (pre-shutdown-endpoint version), the port stays held and the user sees `Stale daemon could not be replaced`. They then have to run `opencli daemon stop && opencli doctor` manually.

99% of "I just upgraded and have to run `opencli doctor` every time" reports land here.

This patch reads the stale daemon's pid from its existing `/status` response (`src/daemon.ts:252` already exposes `pid: process.pid`) and falls back to `process.kill(pid, 'SIGKILL')` after graceful shutdown fails. Cross-platform: Node's `process.kill(_, 'SIGKILL')` maps to `TerminateProcess` on Windows, so no `taskkill` shell-out is needed.

The user-visible `Stale daemon could not be replaced` error now only fires when **both** graceful shutdown AND SIGKILL fail (cross-user owner / cross-machine PID, neither of which is reachable from a normal CLI invocation).

## Reviewer

@coding-claude-opus (independent cross-check on the deep-investigation thread).

## Test plan

- [x] Added 2 unit tests in `src/browser.test.ts`:
  - SIGKILL succeeds → stale path is passed; subsequent no-extension wait surfaces (proves the stale branch is no longer terminal).
  - SIGKILL throws EPERM AND `waitForDaemonStop` keeps returning false → falls through to the original stale-daemon error.
- [x] All 5 stale-daemon tests pass locally (`npx vitest run src/browser.test.ts -t "stale"`).
- [x] `npm run docs:build` clean.
- [x] `npx tsc --noEmit` clean.
- [ ] CI green.

## Out of scope

- Extension version mismatch (Chrome Web Store / unpacked dev install lag) — separate concern, opt-out of this PR per @coding-claude-opus's split proposal. Can be a follow-up with `status` adding `extensionVersion` if WAWQAQ wants.

### DIFF
--- a/docs/guide/troubleshooting.md
+++ b/docs/guide/troubleshooting.md
@@ -42,6 +42,8 @@ opencli doctor
 
 > The daemon is persistent and stays alive until explicitly stopped (`opencli daemon stop`) or the package is uninstalled.
 
+> When the CLI detects a stale daemon (version mismatch after `npm install -g @jackwener/opencli@latest`), it first asks the daemon to shut down via `/shutdown`, then falls back to `SIGKILL` if the daemon does not release the port within 3 seconds. Manual `opencli daemon stop` is only needed if SIGKILL itself is rejected (cross-user owner / cross-machine PID file).
+
 ### Desktop adapter connection issues
 
 For Electron/CDP-based adapters (Cursor, Codex, etc.):
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -151,7 +151,7 @@ describe('BrowserBridge state', () => {
       state: 'no-extension',
       status: {
         ok: true,
-        pid: 1,
+        pid: 999999,
         uptime: 0,
         daemonVersion: PKG_VERSION,
         extensionConnected: false,
@@ -171,7 +171,7 @@ describe('BrowserBridge state', () => {
       state: 'no-extension',
       status: {
         ok: true,
-        pid: 1,
+        pid: 999999,
         uptime: 0,
         extensionConnected: false,
         pending: 0,
@@ -180,6 +180,9 @@ describe('BrowserBridge state', () => {
       },
     });
     vi.spyOn(daemonClient, 'requestDaemonShutdown').mockResolvedValue(false);
+    // Keep the SIGKILL fallback's poll short — neither pid 999999 nor the test
+    // daemon exists, so we'd otherwise spin for 2s on every test in the block.
+    vi.spyOn(daemonLifecycle, 'waitForDaemonStop').mockResolvedValue(false);
 
     const bridge = new BrowserBridge();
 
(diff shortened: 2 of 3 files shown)


jackwener (commented):
LGTM — 22/22 tests pass, typecheck clean. Core fix是干净的 minimal patch（~13 行），用现有 `health.status.pid` 走 SIGKILL，跨平台正确（Node 在 Windows 把 `process.kill(_, 'SIGKILL')` 映射成 `TerminateProcess`）。

两条 **非阻塞** observation，是否处理你拍：

### Nit 1: ESRCH race — kill 抛异常时漏掉了第二次 poll

`bridge.ts:91-98`：
```ts
try {
  process.kill(stalePid, 'SIGKILL');
  portReleased = await waitForDaemonStop(2000);
} catch {
  // EPERM / ESRCH: fall through
}
```

ESRCH 场景：daemon 在 status 查询和 SIGKILL 之间窗口里自己挂了（很窄但存在）→ `process.kill` 抛 ESRCH → catch 吞掉 → **不 poll** → `portReleased` 留 false → 报错 "Stale daemon could not be replaced"。但此时进程已死、端口大概率已释放，下条命令一定 OK。多一次假错误。

更紧的写法：把 poll 拉到 try/catch 外面，无论 throw 与否都问 OS：

```ts
try {
  process.kill(stalePid, 'SIGKILL');
} catch {
  // EPERM (cross-user) / ESRCH (already dead): port may still be free
}
portReleased = await waitForDaemonStop(2000);
```

三种情况都对：成功→端口释放→poll true；ESRCH→端口本来就释放了→poll true；EPERM→端口仍占→poll false→报错。

### Nit 2: 既有 stale tests `pid: 1` 现在有副作用

PR 前，`browser.test.ts:189/210` 那两个 stale 测试 mock 了 `getDaemonHealth.status.pid = 1` 但根本不走 SIGKILL 路径，pid 数字无所谓。PR 后这俩测试会**实际**执行 `process.kill(1, 'SIGKILL')` —— 对 init/launchd 发信号，EPERM 兜住所以测试还过。但本地/CI 多了个对 PID 1 的 syscall side effect。

建议这俩既有 test 顺手改成：要么 `vi.spyOn(process, 'kill').mockImplementation(() => true)`，要么 `pid: -1`（Node 会直接 EINVAL）。

---

两个都不阻塞 merge。Nit 1 顺手改利落一些；Nit 2 是 hygiene。你拍要不要现在改还是 follow-up。

PR #1906, opened 2026-06-09
fix(plugin): reject unsafe plugin names on install to prevent path traversal

Chen17-sq:
## Problem

`opencli plugin install` derives the on-disk plugin directory name from the (attacker-controllable) plugin manifest `name` field — or a monorepo `plugins` key — and joins it straight into `PLUGINS_DIR` with no sanitization. A manifest whose `name` is `../evil`, `evil/sub`, or an absolute path escapes `~/.opencli/plugins`, letting an install write or symlink files outside the plugins directory (e.g. `path.join(PLUGINS_DIR, '../evil')` → a sibling of the plugins dir).

This is a filesystem-write / symlink-placement primitive triggered simply by installing a hostile (or compromised) plugin repo.

## Root cause

`src/plugin.ts` joins a name into `PLUGINS_DIR` at three install/link sites without validating it:

- `installSinglePlugin` — `src/plugin.ts:770` (`const pluginName = manifest?.name ?? name; path.join(PLUGINS_DIR, pluginName)`)
- `installLocalPlugin` — `src/plugin.ts:817` (same shape)
- `publishMonorepoPlugins` — `src/plugin.ts:699` (`path.join(pluginsDir, plugin.name)`, where `plugin.name` comes from the monorepo manifest's `plugins` keys)

`readPluginManifest` (`src/plugin-manifest.ts`) does no name validation, and the existing scaffold regex in `src/plugin-scaffold.ts` only guards `plugin create`, never the install path. (Note: sub-plugin *paths* are already contained by `resolveRepoContainedPath`; the gap is the *name*, which becomes the link's basename.)

## Fix

Add `assertSafePluginName(name)` (`src/plugin.ts`) and call it immediately before every `path.join(PLUGINS_DIR, name)` on the install/link paths. It throws `PluginError` for names that are empty, absolute, contain a path separator (`/` or `\`), contain `..`, start with `.`, or fail the conservative pattern `/^[a-z0-9_][a-z0-9._-]*$/i`. The pattern aligns with the scaffold's naming rules but stays case-insensitive and permits a leading underscore so it does not reject any pre-existing valid install name or monorepo sub-plugin key.

`publishMonorepoPlugins` is the shared chokepoint for both install and update link creation, so validating there covers monorepo installs and is defense-in-depth for updates.

Behavior unchanged for currently-valid plugin names: all 96 pre-existing plugin tests still pass, including installs of names like `github-trending`, `hot-digest`, monorepo keys (`alpha`, `beta`), and the `__test-…__` fixtures.

## Tests

`src/plugin.test.ts`:
- `assertSafePluginName` — asserts conservative names (`github-trending`, `my_tool`, `Plugin.v2`, `a`, `__test-mono-sub__`) are accepted and that `../evil`, `a/b`, `a\b`, `..`, `.hidden`, ``, `/abs/path` throw `PluginError`.
- `installPlugin name sanitization` — stubs `git clone` to emit a manifest whose `name` is `../evil` (and one with `/`); asserts `installPlugin` throws `PluginError` and that nothing is created at `path.join(PLUGINS_DIR, '..', 'evil')`.

Gate: `npx tsc --noEmit` (exit 0) and `npx vitest run src/plugin.test.ts src/plugin-manifest.test.ts` (101 passed).

---

This change is scoped strictly to plugin-name sanitization. It is complementary to and independent of the open #1754 (`--ignore-scripts`), which mitigates a different install-time risk (lifecycle-script execution); neither subsumes the other.

### DIFF
--- a/src/plugin.test.ts
+++ b/src/plugin.test.ts
@@ -8,6 +8,7 @@ import * as os from 'node:os';
 import * as path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { PLUGINS_DIR } from './discovery.js';
+import { PluginError } from './errors.js';
 import type { LockEntry } from './plugin.js';
 import * as pluginModule from './plugin.js';
 
@@ -31,6 +32,7 @@ const {
   _parseSource,
   _updateAllPlugins,
   _validatePluginStructure,
+  _assertSafePluginName,
   _writeLockFile,
   _writeLockFileWithFs,
   _isSymlinkSync,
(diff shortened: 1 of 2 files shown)


Chen17-sq:
Closing this. Re-reading the install path (and the prior reasoning in #847): installing a plugin already runs `npm install` and loads/executes the plugin's JS, so a name-based write/symlink-placement primitive doesn't cross a new trust boundary, and the cleanly-scoped script-execution hardening is already being handled in #1754. The 'every name→path sink' framing here was also overstated (`repoName` from `parseSource` stays unguarded). If there's appetite for name validation as explicit defense-in-depth — ideally via post-join containment using the existing `resolveRepoContainedPath` helper, before the repo's npm install runs — I'm glad to revisit it as that, rather than as a security fix. Thanks!

PR #1926, opened 2026-06-12
fix(cli): accept browser `--window` placed after the subcommand (#1850)

### LINKED ISSUE #1850: [Bug] opencli browser --window only works before the leaf subcommand

LeoLin990405:
## What

`opencli browser <session> <subcommand> ... --window <mode>` fails with `error: unknown option '--window'` (#1850).

`--window` is registered on the `browser` **parent** command, so commander only accepts it *before* the leaf subcommand:

- ✅ `opencli browser work --window background open https://x.com`
- ❌ `opencli browser work open https://x.com --window background` ← the obvious placement

## Fix

`rewriteBrowserArgv` (the existing `browser <session> <subcommand>` argv preprocessor) now hoists a trailing `--window <mode>` / `--window=<mode>` to the parent option slot, so both placements parse identically. argv without `--window` is unchanged; only the first occurrence is hoisted; the scan stops at `--`.

## Tests / verification

- 4 new `rewriteBrowserArgv` cases: trailing form, `=` form, already-in-slot no-op, and with a leading `--profile`. `vitest run src/cli-argv-preprocess.test.ts` → **32 passed**.
- Live: `opencli browser <session> open https://example.com --window background` now opens the page (previously errored with `unknown option '--window'`).

Fixes #1850

### DIFF
--- a/src/cli-argv-preprocess.test.ts
+++ b/src/cli-argv-preprocess.test.ts
@@ -116,6 +116,35 @@ describe('rewriteBrowserArgv', () => {
     ]);
   });
 
+  it('hoists a trailing parent `--window` option to before the subcommand (#1850)', () => {
+    // The natural placement (after the subcommand + its args) must work the same
+    // as the parent-slot placement.
+    expect(rewriteBrowserArgv(['browser', 'work', 'open', 'https://x.com', '--window', 'background'])).toEqual([
+      'browser', '--session', 'work', '--window', 'background', 'open', 'https://x.com',
+    ]);
+    expect(rewriteBrowserArgv(['browser', 'work', 'state', '--window', 'foreground'])).toEqual([
+      'browser', '--session', 'work', '--window', 'foreground', 'state',
+    ]);
+  });
+
+  it('hoists the `--window=<mode>` long form too (#1850)', () => {
+    expect(rewriteBrowserArgv(['browser', 'work', 'open', 'https://x.com', '--window=background'])).toEqual([
+      'browser', '--session', 'work', '--window=background', 'open', 'https://x.com',
+    ]);
+  });
+
+  it('leaves an already-parent-slot `--window` untouched (#1850)', () => {
+    expect(rewriteBrowserArgv(['browser', 'work', '--window', 'background', 'open', 'https://x.com'])).toEqual([
+      'browser', '--session', 'work', '--window', 'background', 'open', 'https://x.com',
+    ]);
+  });
+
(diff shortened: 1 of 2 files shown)


jackwener:
Thanks for the focused fix. I compared this with #1899 because both PRs address #1850.

I opened a replacement PR that uses your lower-blast approach as the core design: keep `--window` as a browser namespace option and hoist the natural trailing form in `rewriteBrowserArgv`, instead of changing the browser command tree.

Replacement PR: #1963

I also credited you in the replacement commit as a co-author:

`Co-authored-by: Zhongyue Lin <zhongyuelin990405@gmail.com>`

Main differences in #1963:
- adds nested command coverage like `browser work get url --window background` and `browser work tab close ... --window background`
- avoids rewriting anything after literal `--`
- avoids hoisting a bare `--window` without a value, so Commander cannot accidentally consume the subcommand token as the option value
- adds a browser help example for the trailing placement

LeoLin990405:
Makes sense — #1963 is the better version (nested-command coverage + the `--` / bare-`--window` edge cases). Closing this in favor of it. Thanks for folding in the approach and the co-author credit! 🙏

PR #1935, opened 2026-06-13
fix(launcher): don't kill the user's app on non-darwin platforms

Chen17-sq:
## Problem

`resolveElectronEndpoint()` gates its auto-launch flow with:

```ts
if (process.platform !== 'darwin' && process.platform !== 'linux') { ...graceful... }
```

so **Linux falls through** into the process-detect → confirm → **kill** → discover → relaunch path. But `discoverAppPath()` resolves an app path on **macOS only** — it hard-returns `null` on every other platform.

The result on Linux, when an Electron app (Cursor, Codex, ChatGPT-app, …) is running *without* CDP:

1. `detectProcess()` finds it (pgrep works on Linux),
2. the user confirms "restart with debug port?",
3. `killProcess()` **pkills the running app** (pkill works on Linux),
4. `discoverAppPath()` returns `null` (non-darwin),
5. the function throws **`Could not find <app> on this machine. Install <app>…`**

So the user's app is killed and never relaunched, behind a misleading "not installed" error — strictly worse than not running the command. The kill branch is reachable on a platform where the subsequent launch path can never succeed.

## Fix

Gate the kill/relaunch path on `darwin` alone:

```ts
if (process.platform !== 'darwin') { ...graceful... }
```

Every non-macOS platform now takes the existing graceful branch (manual-launch hint + `OPENCLI_CDP_ENDPOINT`) **before any kill** — the same degradation already applied to Windows. This drops Linux *auto-launch* (which never worked, since `discoverAppPath` is macOS-only) rather than killing the app and failing; a real Linux `discoverAppPath` implementation can re-enable it later.

## Test

Adds a `resolveElectronEndpoint` regression test: on a non-darwin platform with an app that isn't reachable on CDP, it asserts the call rejects with the graceful "not reachable / auto-launch not supported" error (not the destructive "Could not find …" path) **and that `pkill` is never invoked**. Verified red→green — it fails on `main` (the app is pkilled, wrong error) and passes with the fix.

## Checks

- `npx tsc --noEmit` — clean
- `npm test` — full gate green (514 files, 5345 passed, 1 skipped)

### DIFF
--- a/src/launcher.test.ts
+++ b/src/launcher.test.ts
@@ -1,6 +1,6 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type { ElectronAppEntry } from './electron-apps.js';
-import { detectProcess, discoverAppPath, electronLaunchArgs, launchDetachedApp, launchElectronApp, probeCDP, resolveExecutableCandidates } from './launcher.js';
+import { detectProcess, discoverAppPath, electronLaunchArgs, launchDetachedApp, launchElectronApp, probeCDP, resolveElectronEndpoint, resolveExecutableCandidates } from './launcher.js';
 
 interface MockChildProcess {
   once: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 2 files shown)


jackwener:
关闭为 current main 已消除该破坏路径。现在 resolveElectronEndpoint 会先通过 discoverAppPathForEntry 解析可执行文件；该函数在非 darwin 平台直接返回 null，因此 Linux 不会进入 detect/kill/relaunch 分支，更不会先 pkill 用户应用。当前行为是安全降级为找不到本机 app，原 PR 要防止的 kill 已不可达。

PR #1938, opened 2026-06-13
fix(daemon-client): don't retry a profile_required 409 as a duplicate id

Chen17-sq:
## Problem

`sendCommand()` treats **any** HTTP 409 as a retryable duplicate-command-id collision:

```ts
const isDuplicateCommandId = res.status === 409
  || (result.error ?? '').includes('Duplicate command id');
if (isDuplicateCommandId && attempt < maxRetries) { continue; }
```

But the daemon emits `409` for **two** distinct conditions:

1. genuine duplicate id — `"Duplicate command id already pending; retry"` (retryable with a fresh id), and
2. `profile_required` — `"Multiple Browser Bridge profiles are connected; choose one with --profile."` with `errorCode: 'profile_required'` (**not** retryable; the user must pick a profile).

When multiple Browser Bridge profiles are connected and none is selected, the bare `res.status === 409` makes the client silently re-POST up to `maxRetries` times — every attempt returns the same 409 — then throw the opaque **`sendCommand: max retries exhausted`**, discarding the `BrowserCommandError` `code` (`profile_required`) and `hint` that tell the user exactly what to do (`--profile`).

## Fix

Gate the retry on the duplicate-id **message** alone — the duplicate-id 409 always carries that exact text (`daemon.ts`):

```ts
const isDuplicateCommandId = (result.error ?? '').includes('Duplicate command id');
```

The `profile_required` 409 (no such text, `classifyBrowserError` → non-retryable) now falls through to `throw new BrowserCommandError(result.error, result.errorCode, result.errorHint)`, surfacing the actionable message + code on the **first** attempt.

## Test

Adds a test asserting a `profile_required` 409 rejects with a `BrowserCommandError` (`code: 'profile_required'`, hint preserved) and that `fetch` is called **once** (no retry loop). The existing "retries with a new id when daemon reports a duplicate pending id" test — which sends the duplicate-id message — stays green. Verified red→green.

## Checks

- `npx tsc --noEmit` — clean
- `npm test` — full gate green

### DIFF
--- a/src/browser/daemon-client.test.ts
+++ b/src/browser/daemon-client.test.ts
@@ -222,6 +222,31 @@ describe('daemon-client', () => {
     expect(ids[0]).not.toBe(ids[1]);
   });
 
+  it('sendCommand surfaces a profile_required 409 immediately instead of retrying', async () => {
+    const fetchMock = vi.mocked(fetch);
+    fetchMock.mockResolvedValue({
+      ok: false,
+      status: 409,
+      json: () => Promise.resolve({
+        id: 'server',
+        ok: false,
+        errorCode: 'profile_required',
+        error: 'Multiple Browser Bridge profiles are connected; choose one with --profile.',
+        errorHint: 'Run opencli profile list, then use opencli --profile <name> ...',
+      }),
+    } as Response);
+
+    // The profile_required 409 is not a duplicate-id collision: it must throw
+    // the actionable error on the first attempt, not loop until "max retries
+    // exhausted".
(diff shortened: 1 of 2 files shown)


jackwener:
关闭为已由 current main 的 transport retry 重构覆盖。sendCommandRaw 现在按 typed error 分类：profile_required 不是 pre-dispatch retry code，直接抛出 BrowserCommandError；现有测试 sendCommand does NOT wait when the bridge reports profile_required 明确断言 ensure 未调用且 fetch 仅一次。旧的仅修改 409 判定的补丁已无剩余目标。

PR #1940, opened 2026-06-13
chore(build): regenerate cli-manifest.json to sync siteSession (#1462)

Chen17-sq:
## Problem

`main`'s CI is red on the `build (ubuntu-latest)` job:

```
::error::cli-manifest.json is out of sync with the source. Run 'npm run build' and commit the result.
```

## Root cause

#1462 (`467fdd0b`, "rename site browser reuse to persistent sessions") added the `siteSession` field to the manifest generator — `src/build-manifest.ts:129`:

```ts
siteSession: cmd.siteSession,
```

…but `cli-manifest.json` was never regenerated, so the committed manifest lacks `siteSession` on the affected adapter commands. The `build (ubuntu-latest)` job runs `npm run build` and fails if the regenerated `cli-manifest.json` differs from the committed one — which it now does on `main` and therefore on every PR branched off it.

## Fix

Ran `npm run build-manifest`. The result is a **deterministic** 116/58-line diff that **only** adds the missing `"siteSession": "persistent"` fields to the affected commands — no reordering, no other field changes. Regenerating twice produces an identical file.

This restores `main` (and any PR based on it) to a green manifest-sync check.

Happy to close if you'd rather regenerate it yourself — just flagging that the gate is currently red for everyone branching off `main`.

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -343,7 +343,8 @@
     "type": "js",
     "modulePath": "12306/auth.js",
     "sourceFile": "12306/auth.js",
-    "navigateBefore": false
+    "navigateBefore": false,
+    "siteSession": "persistent"
   },
   {
     "site": "1688",
@@ -557,7 +558,8 @@
     "type": "js",
     "modulePath": "1688/auth.js",
     "sourceFile": "1688/auth.js",
-    "navigateBefore": false
+    "navigateBefore": false,
+    "siteSession": "persistent"
   },
   {
     "site": "1point3acres",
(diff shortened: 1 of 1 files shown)


Chen17-sq:
Closing as obsolete. `cli-manifest.json` was regenerated on `main` as part of the recent adapter additions (e.g. twitter/xiaohongshu/huodongxing/slock, up to `8ed8ca26`), so the missing `siteSession` fields are already in sync upstream and `main`'s CI is green again. This PR now only conflicts against the rewritten manifest, so there's nothing left for it to do. Thanks!

Issue #1777, opened 2026-05-29
[Bug]: Browser connectivity test failed: This operation was aborted


iodone:
### Description

Using Arc browser

### Steps to Reproduce

1. Run `opencli doctor`

```
pencli v1.8.0 doctor (node v25.9.0)

[OK] Daemon: running on port 19825 (v1.8.0)
[OK] Extension: connected (v1.0.15)

Profiles:
  • 3fu9qprx: connected v1.0.15
[FAIL] Connectivity: failed (This operation was aborted)

Issues:
  • Browser connectivity test failed: This operation was aborted
```

### Expected Behavior

None

### OpenCLI Version

1.8.0

### Node.js Version

Other

### Operating System

macOS

### Logs / Screenshots

```shell

```

LeoLin990405:
Looked into this — sharing what the code shows in case it helps narrow things down.

## Where the abort comes from

`doctor` → `checkConnectivity()` in `src/commands/doctor.ts` runs `page.evaluate('1 + 1')` over the bridge → `requestDaemon('/command')` in `src/browser/daemon-client.ts:113-114` wraps the fetch in an `AbortController` with a 30 s timeout. On timeout the fetch throws a DOMException whose `.message` is exactly `"This operation was aborted"`. The `getErrorMessage` helper strips `err.name`/`err.cause`, so the user only sees the abort string — the root cause (daemon never responded within 30 s) is hidden.

## Why Arc specifically might be hitting this

Nothing in OpenCLI is Arc-specific (grepped `src/` — no `Arc` references), but Arc is more aggressive than vanilla Chrome about suspending idle extension service workers. A plausible sequence:

1. `getDaemonHealth` → daemon pings extension → service worker wakes → reports `connected` → check passes
2. `bridge.connect()` succeeds (instant snapshot)
3. `page.evaluate('1+1')` → daemon sends `exec` over WS to extension
4. Arc has re-suspended the worker between steps 1 and 3
5. Daemon's 30 s timeout fires → `AbortError: "This operation was aborted"`

This would explain the issue body's "daemon/extension/profile all OK, only connectivity fails" pattern — a race between the snapshot health checks and the actual command round-trip.

## Other things worth ruling out

- **Multiple Arc Spaces / profiles** — if the extension is installed in Space A but `doctor` binds to a tab in Space B, the WS round-trip lands in the wrong process.
[...]

PR #2067, opened 2026-07-02
fix(browser): end-to-end command deadlines, safe transport retries, CDP timeouts

jackwener:
## Summary

Three P0 connectivity/stability fixes from a deep audit of the CLI → daemon → extension bridge. They share one root cause: the timeout and retry contracts between the three layers were disconnected.

### 1. One command deadline plumbed through all three layers

- `daemon-client` hardcoded a **30s** HTTP timeout while the daemon default was **120s**, and no caller ever set `body.timeout` — the daemon timer and its 408 branch were dead code. Every command slower than 30s (slow navigate, long eval, big screenshot, `wait --for download`) died with an opaque client-side `AbortError` while still running in the browser. Likely the root cause of #1777.
- Now the transport computes an effective timeout and sends it as `body.timeout` (seconds):
  - default 120s;
  - `--timeout N` reaches the transport via `setDaemonCommandTimeoutSeconds()` (wired in `execution.ts`);
  - extension-side waits (`timeoutMs`, e.g. wait-download) get a 15s margin on top.
- Layered so failures surface innermost-first with a real error: **extension CDP deadline < daemon timer < client HTTP abort** (+10s margin).
- The daemon per-command timer now rejects with the structured `command_result_unknown` contract (408) via `buildCommandTimeoutFailure()` instead of a bare `Error` the client cannot classify.

### 2. No more replaying possibly-dispatched commands on fetch `TypeError`

Any `TypeError: fetch failed` used to trigger ensure + resend **with a fresh id**, bypassing the daemon's duplicate-id guard. undici raises the same `TypeError` for `ECONNREFUSED` (request never sent — safe to retry) and `ECONNRESET`/socket hang-up (daemon may have already dispatched the command — a daemon crash mid-`click` could double-submit a form). The retry now checks `err.cause` and only pre-connect failures (`ECONNREFUSED`, `UND_ERR_CONNECT_TIMEOUT`, …) go through the ensure+resend path; post-connect drops surface as `command_result_unknown`, consistent with the existing write-safety contract.

### 3. Real deadlines for `chrome.debugger` commands

The extension's CDP calls had no timeout at all — `sendCommandInFrameTarget` declared `_timeoutMs` and never used it. A page-blocking native dialog (`alert`/`confirm`/`beforeunload`) hung `Runtime.evaluate` forever, and since `ensureAttached`'s health probe is also a `Runtime.evaluate`, **every subsequent command on the tab wedged**. Now:

- all `chrome.debugger.sendCommand` calls go through `sendDebuggerCommand()` which races a timer (default 60s, health probe 2s);
- `exec` / `cdp` commands derive their deadline from the transport's `body.timeout`, undercut by 5s so the more specific extension error beats the daemon's generic timer;
- the timeout error message points at the likely cause (native dialog).

## Testing

- `npm test` — full suite green (544 files, 5901 tests).
- `tsc --noEmit` clean in both packages (extension has pre-existing unrelated test-file strictness errors, unchanged).
- `vite build` for the extension bundle succeeds; `extension/dist/background.js` rebuilt and committed per repo convention.
- New regression tests:
  - `daemon-client`: default/`timeoutMs`/user-timeout plumbing into `body.timeout`; ECONNREFUSED retried, ECONNRESET and bare `TypeError` → `command_result_unknown` without resend.
  - `daemon`: `buildCommandTimeoutFailure` contract (408 + `command_result_unknown`).
  - `extension/cdp`: hung `Runtime.evaluate` now rejects at the default and caller-supplied deadlines instead of hanging forever.

Fixes the "connected but `This operation was aborted`" failure mode reported in #1777.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -11,6 +11,24 @@ let frameTargetCleanupRegistered = false;
 const CDP_RESPONSE_BODY_CAPTURE_LIMIT = 8 * 1024 * 1024;
 const CDP_REQUEST_BODY_CAPTURE_LIMIT = 1 * 1024 * 1024;
 const networkCaptures = /* @__PURE__ */ new Map();
+const CDP_COMMAND_TIMEOUT_MS = 6e4;
+const CDP_PROBE_TIMEOUT_MS = 2e3;
+async function sendDebuggerCommand(target, method, params, timeoutMs = CDP_COMMAND_TIMEOUT_MS) {
+  let timer;
+  const commandPromise = params === void 0 ? chrome.debugger.sendCommand(target, method) : chrome.debugger.sendCommand(target, method, params);
+  try {
+    return await Promise.race([
+      commandPromise,
+      new Promise((_, reject) => {
+        timer = setTimeout(() => reject(new Error(
+          `CDP command ${method} timed out after ${Math.round(timeoutMs / 1e3)}s — the page may be blocked by a native dialog (alert/confirm/print)`
+        )), timeoutMs);
+      })
+    ]);
+  } finally {
+    if (timer !== void 0) clearTimeout(timer);
+  }
+}
 function isDebuggableUrl$1(url) {
   if (!url) return true;
   return url.startsWith("http://") || url.startsWith("https://") || url === "about:blank" || url.startsWith("data:");
@@ -29,10 +47,10 @@ async function ensureAttached(tabId, aggressiveRetry = false) {
   }
   if (attached.has(tabId)) {
     try {
-      await chrome.debugger.sendCommand({ tabId }, "Runtime.evaluate", {
+      await sendDebuggerCommand({ tabId }, "Runtime.evaluate", {
         expression: "1",
         returnByValue: true
-      });
+      }, CDP_PROBE_TIMEOUT_MS);
       return;
     } catch {
       attached.delete(tabId);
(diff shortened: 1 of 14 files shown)


jackwener:
Self + adversarial review of the deadline design landed three follow-up commits:

- `2cd610f8` — swallow post-timeout CDP rejections (unhandled rejection in the SW when the losing race promise rejected later, e.g. detach on tab close); doctor's connectivity probe now shrinks the transport deadline to its 8s budget instead of inheriting the 120s default (a hung-but-accepting daemon made `doctor` hang 2 minutes).
- `c3382c5c` — `evaluateInFrame`'s `Runtime.enable` warm-up on the frame-target fallback path dropped the derived deadline (fell back to the 60s default), so on the cross-frame path the daemon's generic 408 always beat the extension's specific error. Also pinned the two previously untested links of the deadline chain: client HTTP abort at exactly `timeout*1000 + 10s`, and `handleExec` deriving `115s` from a `120s` transport timeout (10s floor).

Known design limits, deliberately out of scope here (follow-ups): duration-based propagation doesn't absorb SW wake / queueing latency (an absolute deadline timestamp would — same machine, no clock skew); retries still mint a fresh id per attempt, so true retry safety needs stable idempotency ids + a daemon result cache; deadlines bound the wait but don't cancel the work (`Runtime.terminateExecution` + auto `Page.handleJavaScriptDialog` would remove the dialog-hang root cause entirely).

PR #2070, opened 2026-07-03
refactor(transport): exactly-once command transport — journal, waiters, absolute deadlines

jackwener:
## Why

The transport layer had accumulated overlapping compensation at every hop: 6 retry owners, 7 timeout owners with 5 hand-tuned margin constants, 3 parallel error-classification mechanisms, and 4 sibling session-override Maps. Each layer distrusted the one below it. This PR rebuilds the contract around one principle:

> **exactly-once = at-least-once retry + an idempotent executor.**

Once the executor dedupes command ids, re-sending is harmless — so most of the per-layer compensation becomes deletable, and `command_result_unknown` shrinks from "any connection blip" to "the browser itself exited".

## The three primitives

**1. Command journal** (`extension/src/journal.ts`, ~130 lines, `chrome.storage.session`)
Every command id executes exactly once: concurrent duplicates attach to the in-flight promise; completed ids replay the recorded result; ids whose service worker died mid-execution report `command_lost` honestly (write-ahead `started` marker). `storage.session` survives SW restarts and clears on browser exit — precisely the lifetime a retry cares about. Oversized results (>64KB) are not recorded; their replay returns `result_evicted` instead of silently re-executing.

**2. Stable ids + daemon waiters**
Transport retries keep the **same** id; the daemon attaches duplicate ids to the pending command (waiter list) instead of rejecting with 409. Retry coverage matrix:
- WS flap / SW keepalive jitter → in-memory dedup hits
- SW killed and restarted → `storage.session` journal hits
- daemon crash/restart → new daemon forwards, extension journal hits
- browser exit → honestly unknown (irreducible)

Gated on extension ≥ 1.0.22 (learned from the ensure path's health probe); legacy extensions keep the old conservative pre-connect-only retry. Semantic retries (`attach_failed`/`tab_gone` — failures **before** any page code ran) are the only place a new id is minted, once.

**3. Absolute deadline** (`deadlineAt`, epoch ms)
Same machine, one clock: every layer computes `remaining = deadlineAt - now`, so daemon queueing and SW wake latency no longer silently shrink the innermost budget or invert the layering. The legacy `timeout` (seconds) field is still sent for older daemons.

## Error contract: classify once, at the source

The extension tags failures with machine-readable codes (`attach_failed`, `tab_gone`, `target_navigated`, `detached_mid_command`, `cdp_timeout`); `errors.ts` prefers codes and keeps the message-pattern tables only as a legacy fallback. `detached_mid_command`/`cdp_timeout` are now correctly **non-retryable** — they die mid-execution, so a blind re-run could double-apply a write (previously "Detached while handling command" was blindly retried).

## Deleted / simplified

- extension `evaluate()`'s inner retry loop — the client owns semantic retries now (one owner per retry class: transport=same id, semantic=new id, target-navigation=page layer)
- the fast/slow phased reconnect window + `notifyDaemonReachable` rescheduling → plain exponential backoff with jitter (1s→…→15s cap, reset on success)
- the three parallel session-override Maps → one record per lease
- the three one-shot retry flags in `sendCommandRaw` → two explicit retry classes

## Stability fixes riding the same contract

- **WS application-level keepalive** (`{type:'ping'}` every 20s): Chrome 116+ only extends the SW lifetime on WS *activity*; an idle socket lived on a knife-edge between the 30s idle kill and the 30s keepalive alarm
- **idle-lease release deferred while a command executes** (refcount) — a 30s idle timer can no longer tear the tab down mid-command; the idle window now measures from command completion
- **daemon shutdown flushes structured 503s** (`daemon_shutting_down`) before exiting, instead of `process.exit()` killing the queued responses; clients re-ensure and retry the same id
- **results delivered on the freshest open socket** after a reconnect instead of being dropped when the executing socket was superseded

## Compatibility

- new CLI + old daemon: sends `timeout` alongside `deadlineAt`; old daemon's 409 path is only reachable from old clients (new clients never reuse an id against a still-pending old daemon without the ensure probe running first)
- new CLI + old extension: capability gate falls back to the previous conservative retry semantics
- old CLI + new daemon/extension: never reuses ids; journal and waiters are inert

## Testing

- Full suite: **545 files / 5913 tests green**; `tsc --noEmit` clean in both packages; extension `vite build` OK
- New: `journal.test.ts` (replay, in-flight attach, `command_lost` after WAL, `result_evicted`, storage-less degradation), same-id transport retry semantics, capability gating, executor-transient new-id retry, `detached_mid_command` non-retry, backoff shape tests
- Extension version bumped to **1.0.22** (= `MIN_JOURNAL_EXTENSION_VERSION`)

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -115,34 +115,25 @@ async function ensureAttached(tabId, aggressiveRetry = false) {
   }
 }
 async function evaluate(tabId, expression, aggressiveRetry = false, timeoutMs = CDP_COMMAND_TIMEOUT_MS) {
-  const MAX_EVAL_RETRIES = aggressiveRetry ? 3 : 2;
-  for (let attempt = 1; attempt <= MAX_EVAL_RETRIES; attempt++) {
-    try {
-      await ensureAttached(tabId, aggressiveRetry);
-      const result = await sendDebuggerCommand({ tabId }, "Runtime.evaluate", {
-        expression,
-        returnByValue: true,
-        awaitPromise: true
-      }, timeoutMs);
-      if (result.exceptionDetails) {
-        const errMsg = result.exceptionDetails.exception?.description || result.exceptionDetails.text || "Eval error";
-        throw new Error(errMsg);
-      }
-      return result.result?.value;
-    } catch (e) {
-      const msg = e instanceof Error ? e.message : String(e);
-      const isNavigateError = msg.includes("Inspected target navigated") || msg.includes("Target closed");
-      const isAttachError = isNavigateError || msg.includes("attach failed") || msg.includes("Debugger is not attached") || msg.includes("chrome-extension://");
-      if (isAttachError && attempt < MAX_EVAL_RETRIES) {
-        attached.delete(tabId);
-        const retryMs = isNavigateError ? 200 : 500;
-        await new Promise((resolve) => setTimeout(resolve, retryMs));
-        continue;
-      }
-      throw e;
+  try {
+    await ensureAttached(tabId, aggressiveRetry);
+    const result = await sendDebuggerCommand({ tabId }, "Runtime.evaluate", {
+      expression,
+      returnByValue: true,
+      awaitPromise: true
+    }, timeoutMs);
+    if (result.exceptionDetails) {
+      const errMsg = result.exceptionDetails.exception?.description || result.exceptionDetails.text || "Eval error";
+      throw new Error(errMsg);
     }
+    return result.result?.value;
+  } catch (e) {
+    const msg = e instanceof Error ? e.message : String(e);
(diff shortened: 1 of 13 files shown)


jackwener:
Adversarial review landed one follow-up commit (`322784b5`): **(1)** `daemon_shutting_down` was resent same-id without the journal capability gate — on a pre-1.0.22 extension that re-executes a dispatched write. The daemon now splits the shutdown contract by `dispatched` (pre-dispatch → safe-resend contract; dispatched → `daemon_shutting_down`), and the client resends the latter only when the extension journals ids. **(2)** `ensureBridge`'s 45s connect wait is now clamped to the command's remaining deadline so repeated daemon failures can't stretch a 30s `--timeout` past two minutes. Both pinned by new regression tests; full suite 5915 green. Known follow-up: daemon-level integration tests for the waiter/shutdown paths (currently covered only from the client side).

PR #2073, opened 2026-07-03
fix(browser): stale default profile must not veto live connections

jackwener:
## The bug (user report)

`browser-profiles.json` has `defaultContextId: zvypsyje` from an old `opencli profile use`; the extension was since reinstalled and the only connected profile is `pavmrekj`. Every command — including `doctor` — fails with `profile_disconnected`, violating the README promise: *"With only one connected profile, OpenCLI uses it automatically."*

Introduced by #1235 (v1.7.10): the multi-profile feature folded three inputs of very different nature — `--profile` argument, `OPENCLI_PROFILE` env, and the persisted config default — into one hard `contextId` on every wire command. The daemon's single-profile auto-use branch is only reachable when NO contextId is requested, so a stale default silently converts a *preference* into a *requirement*. Worse, the ensure path then waits for the dead profile until the connect timeout, so `doctor` hangs before failing.

## First-principles fix

**A persisted default's lifetime routinely exceeds the extension instance it names** (reinstalls regenerate the contextId) — so a default must never have veto power over live state. The fix separates REQUIREMENT from PREFERENCE end to end, and moves arbitration to the only component that knows live connections:

- `profile.ts` resolves a `ProfileSelection` tagged `explicit` (`--profile` arg / `OPENCLI_PROFILE` env → strict, fail loud) or `preferred` (config default); `profileRouteParams()` maps it onto the wire.
- New wire field `preferredContextId` (kept mutually exclusive with the strict `contextId`).
- Daemon arbitrates via a pure `resolveProfileRoute()` (daemon-utils, fully unit-tested):
  - requested → strict match, unchanged;
  - preferred → use when connected; **fall back to the only connected profile when stale** (warned once per stale id in the daemon log); `profile_required` with a stale-default hint when multiple are connected;
  - none → existing behavior.
- `bridge.connect`/ensure pin readiness to a profile **only for explicit requirements** — a stale preference no longer makes `connect()`/`doctor` wait for a dead profile.
- `doctor` now surfaces the stale default explicitly: which alias/id is stale, what commands currently fall back to, and the recovery command (`opencli profile use`).

## Compatibility

- **New CLI + old daemon**: old daemon ignores `preferredContextId` → equivalent to sending no contextId → single-profile auto-use, which is exactly the documented behavior (better than today). Multi-profile setups get `profile_required` until the daemon is replaced (daemons are version-matched and auto-replaced).
- **Old CLI + new daemon**: unchanged (old clients only send the strict `contextId`).
- Extension ignores the new field entirely; no version bump needed.

## Testing

- Full suite green: **546 files / 5928 tests**; both packages `tsc --noEmit` clean; extension `vite build` OK.
- New: `profile.test.ts` (selection source-tagging matrix, alias resolution, route param mapping), `resolveProfileRoute` arbitration matrix (strict/fallback/multi/none), wire-level split (`OPENCLI_PROFILE` → `contextId`, config default → `preferredContextId`), doctor stale-default report with fallback note.

Reproduces the reporter's exact scenario in tests: `preferred=zvypsyje, connected=[pavmrekj]` → routes to `pavmrekj` with `fallbackFrom: zvypsyje`.

### DIFF
--- a/extension/src/protocol.ts
+++ b/extension/src/protocol.ts
@@ -75,8 +75,14 @@ export interface Command {
   idleTimeout?: number;
   /** Frame index for cross-frame operations (0-based, from 'frames' action) */
   frameIndex?: number;
-  /** Browser profile/context selected by the CLI. Used by the daemon for routing. */
+  /** Browser profile/context REQUIRED by the CLI (--profile / env). Used by the daemon for strict routing. */
   contextId?: string;
+  /**
+   * Browser profile/context PREFERRED by the CLI (persisted config default).
+   * Daemon-only routing hint: used when connected, otherwise the daemon falls
+   * back to the only connected profile. The extension ignores this field.
+   */
+  preferredContextId?: string;
   /**
    * Daemon-side command timeout in seconds, set by the CLI transport. The
    * extension derives its CDP deadline from this so it fails just before the
--- a/src/browser/bridge.ts
+++ b/src/browser/bridge.ts
@@ -6,7 +6,7 @@ import type { ChildProcess } from 'node:child_process';
 import type { IPage } from '../types.js';
 import type { IBrowserFactory } from '../runtime.js';
 import { Page } from './page.js';
-import { resolveProfileContextId } from './profile.js';
+import { profileRouteParams, resolveProfileSelection } from './profile.js';
 import { ensureBrowserBridgeReady } from './daemon-lifecycle.js';
 
 const DAEMON_SPAWN_TIMEOUT = 10000; // 10s to wait for daemon + extension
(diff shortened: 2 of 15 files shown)


jackwener:
Adversarial review landed one follow-up (`e4a1e117`): **saved-tab scope key mismatch** — `getBrowserPage` keyed the remembered-tab scope by the profile *selection* (explicit or preferred) while `getPageScope` read only the explicit `Page.contextId`, so with a config-default profile `browser tab select` saved under `<session>` but every lookup used `<contextId>:<session>` — the selection was silently forgotten on each command. Both sites now key by the selected profile.

Known observability gap, deliberately out of scope: `/status` and `getDaemonHealth` don't see `preferredContextId`, so a live fallback isn't visible in health state — doctor covers it by independently comparing the config default against connected profiles, and the daemon logs `fallbackFrom` once per stale id on the command path. Full suite 5928 green.

PR #2098, opened 2026-07-06
fix(extension): stop SW wake events from wiping the lease registry, add owned-group ledger

### LINKED ISSUE #2097: [Bug]: Multiple OpenCLI Browser tab groups can accumulate instead of reusing one group
## Description

Chrome can accumulate multiple `OpenCLI Browser` tab groups instead of reusing a single group.

This is separate from the feature request to disable tab groups entirely. One reusable empty `OpenCLI Browser` group may be by design, but multiple same-name groups should not keep accumulating.

## Environment

- OpenCLI: 1.8.6
- Extension: 1.0.22
- Chrome: 149.0.7827.201
- macOS: 26.5.2

## Observed behavior

In one Chrome window, we observed 4 separate `OpenCLI Browser` tab groups.

Some groups were empty or left at `about:blank`. The Chrome window also showed the OpenCLI debug banner, so these were OpenCLI-owned browser groups rather than user-created groups.

The issue appeared while multiple agents were using browser-backed OpenCLI automation around the same time.

## Expected behavior

OpenCLI should reuse one existing `OpenCLI Browser` group for the browser role, or otherwise clean up stale groups, instead of leaving multiple same-name groups in the tab strip.

## Scope

This issue is only about duplicate / accumulating `OpenCLI Browser` groups.

It is not asking to disable tab groups entirely. That is covered by #2069.

## Related history

- #1692 reported duplicate `OpenCLI Browser` / `OpenCLI Adapter` groups, but it is closed and was reported against an older version.
- #1627 clarified that one empty reusable group can be by design, but multiple accumulating groups should be treated as a bug.

Astro-Han:
## Description

Chrome accumulates multiple `OpenCLI Browser` tab groups instead of reusing one (#2097).

**Root cause** — two defects chain together:

1. **SW wake events wipe the persisted registry.** An MV3 service worker woken by an event can run `windows.onRemoved` / `tabs.onRemoved` / the lease idle alarm before `initialize()`'s recovery chain has rehydrated in-memory state from storage. Each of those handlers ends in `persistRuntimeState()`, so the empty pre-recovery snapshot overwrites the registry — destroying the `groupId` self-heal pointer (#1862) and every lease record. `tabs.onRemoved` fires for every tab the user closes anywhere in the profile, so this happens routinely.
2. **Untitled orphans then become unrecoverable.** A worker crash between `chrome.tabs.group` and the `tabGroups.update` that sets the title leaves an untitled group whose recovery depends entirely on the persisted pointer or a registered lease tab — exactly the two signals the wipe destroys. Invisible to all discovery layers, the orphan lingers while the next command creates a fresh titled group. Concurrent agents amplify wake and ensure frequency, matching the report.

**Fix:**

- **Startup readiness gate** — `workerReady` resolves once contextId + registry recovery complete. The three event entry points and `connect()` gate on it, so an empty snapshot can never overwrite the registry and daemon commands can't run against un-rehydrated state. The gate always resolves, and `connect()` keeps synchronous `connectInFlight` coalescing via a settled-state mirror, leaving the steady-state path unchanged.
- **Owned-group ledger** — every interactive group id created or adopted in the current browser session is recorded (as `groupIds` on the registry's interactive container) and used as an extra discovery layer in `collectOwnedGroupCandidates`, so an untitled orphan stays findable with no lease and no title across MV3 worker restarts. Ids whose `chrome.tabGroups.get` fails are pruned. The id is persisted between `chrome.tabs.group` and the `tabGroups.update` title call, so a crash in that window self-heals.
- **Session-scoped registry** — the lease registry lives in `chrome.storage.session`, not `chrome.storage.local`. Every meaningful field in it (window ids, tab ids, group ids) is a Chrome runtime number valid only within one browser session; persisting them across a browser restart never enabled real recovery, but a recycled id could collide with a user-created window/tab/group and let the restore path claim, group, navigate, or close it (regression-tested for all three id kinds). `storage.session` survives MV3 worker restarts — the case recovery exists for — and is cleared exactly when the ids die. `initialize()` best-effort removes the legacy `storage.local` key so old data can never be trusted again. Documented boundary: `storage.session` is also cleared on extension disable/reload/update, so recovery is only promised across service-worker restarts within one browser session.
- **Startup convergence** — `reconcileTargetLeaseRegistry` runs one interactive group ensure at the end, so existing orphans (including ones left by older versions) are adopted, retitled, and merged on wake.

Related issue: fixes #2097

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

- [x] I ran the checks relevant to this PR (`tsc --noEmit`; full unit suite: 548 files / 5961 tests pass; `npm run test:e2e`: fixed-port transport suites pass — 3 remaining failures are live-site content assertions (hackernews/tieba/ruanyifeng) that vary between runs and are unrelated)
- [x] I updated tests or docs if needed (9 regression tests, each fails on pre-fix code)
- [x] I included output or screenshots when useful (live end-to-end verification below)

Not an adapter PR — extension-only change; `extension/dist/background.js` rebuilt and committed per repo convention.

## Screenshots / Output

Live verification against a real browser (Chrome for Testing 148, fresh profile, this branch's unpacked extension + built CLI, driving the extension SW over CDP):

**1. Four concurrent agent sessions → exactly one group:**

```
$ for s in agent1..agent4: opencli --profile <test> browser $s open https://example.com   (parallel)
chrome.tabGroups.query({}) →
[{"id":1895863660,"title":"OpenCLI Browser","color":"orange","windowId":149875950}]
```

**2. Kill the SW, wake it via `tabs.onRemoved` (the pre-fix wipe path) → registry survives:**

```
before: {"groupId":1946898140,"groupIds":[1946898140]}
SW stopped; victim tab closed (wake event fired)
after:  {"groupId":1946898140,"groupIds":[1946898140]}   // pre-fix: groupId wiped to null
```

**3. Inject the #2097 crash artifact (untitled group + ledger entry), kill + wake the SW → orphan adopted, not duplicated:**

```
injected: [{"id":1946898140,"title":"OpenCLI Browser"},{"id":227016254,"title":""}]  // 2 groups, 1 window
after wake:
[{"id":227016254,"title":"OpenCLI Browser","color":"orange"}]                        // converged to 1
```

**4. Two more concurrent sessions → still one group, dead ledger id pruned:**

```
[{"id":227016254,"title":"OpenCLI Browser"}]
ledger: {"groupId":227016254,"groupIds":[227016254]}
```

**5. Seed a legacy registry whose persisted `groupId` collides with a live user group, kill + wake the SW → user group untouched:**

```
seeded: local registry ownedContainers.interactive = {"windowId":null,"groupId":1766190816}   // id 1766190816 = user "Vacation" group
after wake: [{"id":1766190816,"title":"Vacation","color":"blue"}]                             // no retitle, no merge, tab kept
```

**6. Simulate post-browser-restart state (stale `storage.local` registry claiming a live user tab + window, `storage.session` empty), kill + wake the SW → user tab untouched, legacy key removed:**

```
seeded: local registry with ownedContainers.*.windowId and a lease preferredTabId all pointing at a real user tab (https://example.com)
after wake: userTab {"url":"https://example.com/","groupId":-1}   // not grouped, not navigated, not closed
            localRegistryAfter: null                              // legacy storage.local key removed
            sessionRegistry: {"leases":[],"containers":{...windowId:null}}   // live registry now in storage.session, no false leases
```

**7. Same-session SW kill with a lease in the session registry → lease survives the `tabs.onRemoved` wake:**

```
after wake: leaseSurvived {"preferredTabId":215961409,"windowId":215961404}, leaseCount 1
```

Note: `tests/e2e/browser-ax-chrome.test.ts` soft-passes on macOS when the bridge is unavailable (branded Chrome ≥137 ignores `--load-extension`), so the live run above was done with Chrome for Testing to exercise the real MV3 service worker.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -944,29 +944,7 @@ const ownedContainers = {
   interactive: { windowId: null, groupId: null, promise: null, groupPromise: null },
   automation: { windowId: null, groupId: null, promise: null, groupPromise: null }
 };
-const INTERACTIVE_GROUP_LEDGER_KEY = "opencli_interactive_group_ledger_v1";
 const interactiveGroupLedger = /* @__PURE__ */ new Set();
-async function restoreInteractiveGroupLedger() {
-  try {
-    const session = chrome.storage?.session;
-    if (!session) return;
-    const raw = await session.get(INTERACTIVE_GROUP_LEDGER_KEY);
-    const stored = raw[INTERACTIVE_GROUP_LEDGER_KEY];
-    interactiveGroupLedger.clear();
-    if (Array.isArray(stored)) {
-      for (const id of stored) {
-        if (typeof id === "number") interactiveGroupLedger.add(id);
-      }
-    }
-  } catch {
-  }
-}
-async function persistInteractiveGroupLedger() {
-  try {
-    await chrome.storage?.session?.set({ [INTERACTIVE_GROUP_LEDGER_KEY]: [...interactiveGroupLedger] });
-  } catch {
-  }
-}
 class CommandFailure extends Error {
   constructor(code, message, hint) {
     super(message);
@@ -1063,7 +1041,10 @@ function emptyRegistry() {
     version: 2,
     contextId: currentContextId,
     ownedContainers: {
-      interactive: { windowId: ownedContainers.interactive.windowId },
+      interactive: {
+        windowId: ownedContainers.interactive.windowId,
+        groupIds: [...interactiveGroupLedger]
+      },
       automation: { windowId: ownedContainers.automation.windowId }
     },
     leases: {}
@@ -1082,7 +1063,8 @@ async function readRegistry() {
       contextId: currentContextId,
       ownedContainers: {
         interactive: {
-          windowId: typeof storedContainers.interactive?.windowId === "number" ? storedContainers.interactive.windowId : null
+          windowId: typeof storedContainers.interactive?.windowId === "number" ? storedContainers.interactive.windowId : null,
+          groupIds: Array.isArray(storedContainers.interactive?.groupIds) ? storedContainers.interactive.groupIds.filter((id) => typeof id === "number") : []
         },
         automation: {
           windowId: typeof storedContainers.automation?.windowId === "number" ? storedContainers.automation.windowId : null
@@ -1122,7 +1104,10 @@ async function persistRuntimeState() {
     version: 2,
     contextId: currentContextId,
     ownedContainers: {
-      interactive: { windowId: ownedContainers.interactive.windowId },
+      interactive: {
+        windowId: ownedContainers.interactive.windowId,
+        groupIds: [...interactiveGroupLedger]
+      },
       automation: { windowId: ownedContainers.automation.windowId }
     },
     leases
@@ -1233,7 +1218,7 @@ async function collectOwnedGroupCandidates(role) {
       ledgerPruned = true;
     }
   }
-  if (ledgerPruned) await persistInteractiveGroupLedger();
+  if (ledgerPruned) await persistRuntimeState();
   for (const title of getOwnedContainerGroupTitles(role)) {
     const groups = await chrome.tabGroups.query({ title });
     for (const group of groups) groupsById.set(group.id, group);
@@ -1330,10 +1315,7 @@ async function createOwnedGroup(role, windowId, ids) {
   const groupId = await chrome.tabs.group({ tabIds: ids, createProperties: { windowId } });
   ownedContainers[role].groupId = groupId;
   ownedContainers[role].windowId = windowId;
-  if (role === "interactive") {
-    interactiveGroupLedger.add(groupId);
-    await persistInteractiveGroupLedger();
-  }
+  if (role === "interactive") interactiveGroupLedger.add(groupId);
   await persistRuntimeState();
   const group = await chrome.tabGroups.update(groupId, {
     color: OWNED_TAB_GROUP_COLOR,
@@ -1372,7 +1354,7 @@ async function ensureOwnedContainerGroupUnlocked(role, fallbackWindowId, ids) {
       ownedContainers[role].groupId = canonical.id;
       if (!interactiveGroupLedger.has(canonical.id)) {
         interactiveGroupLedger.add(canonical.id);
-        await persistInteractiveGroupLedger();
+        await persistRuntimeState();
       }
     } else {
       ownedContainers[role].groupId = null;
@@ -2338,7 +2320,8 @@ async function releaseLease(leaseKey, reason = "released") {
 }
 async function reconcileTargetLeaseRegistry() {
   const registry = await readRegistry();
-  await restoreInteractiveGroupLedger();
+  interactiveGroupLedger.clear();
+  for (const id of registry.ownedContainers.interactive.groupIds) interactiveGroupLedger.add(id);
   for (const role of Object.keys(ownedContainers)) {
     ownedContainers[role].windowId = registry.ownedContainers[role]?.windowId ?? null;
     const windowId = ownedContainers[role].windowId;
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -742,6 +742,8 @@ const CONTEXT_ID_KEY = "opencli_context_id_v1";
 let currentContextId = "default";
 let contextIdPromise = null;
 let connectInFlight = null;
+let workerReady = Promise.resolve();
+let workerRecovered = true;
 async function getCurrentContextId() {
   if (contextIdPromise) return contextIdPromise;
   contextIdPromise = (async () => {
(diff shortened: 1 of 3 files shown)


Astro-Han:
Addressed external review feedback in c2361998:

- **P1** — the orphan-group ledger no longer touches `chrome.storage.local`. Tab group ids are session-scoped, so a stale persisted id could collide with a recycled id on a user-created group after a browser restart and let the ledger retitle/merge it. The ledger now lives in `chrome.storage.session` (survives MV3 worker restarts, cleared with the browser session); `StoredRegistry` is back to its pure v2 shape and legacy `groupIds` left in `storage.local` are ignored — covered by a new regression test (a live user group named "Vacation" referenced by a legacy ledger entry stays untouched after reconcile).
- **P3** — the ledger is now interactive-only module state; the dead `groupIds` field on the automation container is gone.

Re-verified live with Chrome for Testing: injected an untitled orphan + session-ledger entry, killed the SW via CDP, woke it with a tab close — the orphan is adopted and converged (1 group), and `storage.session` survives the worker kill. `tsc --noEmit` and the full suite (5957 tests) pass.

Astro-Han:
Follow-up on the review: the singular `ownedContainers.interactive.groupId` had the same browser-session-scoped hazard as the plural `groupIds` and is now also gone from the durable registry — `chrome.storage.local` stores `windowId` only, and `readRegistry()` ignores any legacy `groupId`/`groupIds` fields. Fixed in 2cb5fb17.

- Within one browser session, recovery after an MV3 worker restart is covered by the `chrome.storage.session` ledger + title layer + lease layer, so the local pointer was redundant (the ledger records the id before the `tabGroups.update` title call lands).
- New regression test: local registry seeds `interactive.groupId = 400` colliding with a live user-created "Vacation" group (session ledger empty) → after `reconcileTargetLeaseRegistry()`, no `tabGroups.update`/`tabs.group` calls, the user group keeps its title and tabs, and the in-memory container never adopts 400. The test fails on the previous commit with `expected 'OpenCLI Browser' to be 'Vacation'` — i.e. it reproduces the hijack.
- Also verified live (Chrome for Testing 148, CDP): seeded the legacy registry against a real "Vacation" group, killed the SW, woke it via `tabs.onRemoved` → group untouched, registry rewritten without any group id (output in the updated PR description, item 5).

Astro-Han:
Follow-up on the second review round: rather than patching `windowId`/`preferredTabId` field by field, the whole lease registry now lives in `chrome.storage.session` — fixed in 292339c6.

Rationale: every meaningful field in the registry (window ids, tab ids, group ids) is a Chrome runtime number valid only within one browser session, and the registry's sole purpose is surviving MV3 service-worker restarts. There is no stable cross-restart state in it (contextId is recomputed at startup; lifecycle/deadlines hang off the leases). `storage.session`'s lifetime matches exactly — survives worker restarts, cleared when the ids die — so this removes the recycled-id hazard as a category instead of one field at a time. `storage.local` no longer holds the key at all; `initialize()` best-effort removes the legacy copy older versions wrote.

- New regression tests (red on the previous commit, green now): a legacy local registry claiming a live user window is ignored (the next lease creates its own window instead of opening tabs into the user's); a legacy lease pointing at a recycled tab id is not resurrected (no group/navigate/remove on the user tab); the legacy local key is removed on startup.
- Existing recovery tests were re-seeded to `storage.session` (they model same-session SW restarts); the three earlier hijack tests keep seeding `storage.local` and now jointly pin down that nothing in local is ever trusted.
- Verified live (Chrome for Testing 148, CDP): simulated post-browser-restart state (stale local registry claiming a real user tab, session empty) → user tab untouched, legacy key removed, live registry in session; and a same-session SW kill with a session-registry lease → lease survives the `tabs.onRemoved` wake (outputs in the updated PR description, items 6-7).

Full deterministic suite: 5961 tests pass.

Astro-Han:
Both P3 follow-ups from the review are addressed in 305de76d:

- **Ledger folded into the session registry.** The separate `opencli_interactive_group_ledger_v1` key and its restore/persist pair are deleted; the ledger is now a `groupIds` array on the registry's interactive container (in-memory Set and pruning/adoption logic unchanged). The crash-self-heal persist between `chrome.tabs.group` and `tabGroups.update` stays at the same point and is now one storage write instead of two. Net −37 lines of production code.
- **Recovery boundary documented.** The `StoredRegistry` comment now states that `storage.session` is also cleared on extension disable/reload/update, so recovery is only promised across service-worker restarts within one browser session — no durable-id logic should be added for the reload/update case.

Verified: full deterministic suite (5961 tests) green; live check with Chrome for Testing — injected an untitled orphan group recorded only in the merged registry `groupIds`, killed the SW, woke it via `tabs.onRemoved` → orphan adopted (titled/colored), `groupIds` tracks it, and session storage now holds the single registry key.

PR #2119, opened 2026-07-10
fix(skills): quote sitemap author frontmatter

TopTomCruise:
## Why
`opencli-sitemap-author` has invalid YAML frontmatter: unquoted colons in both `description` and `allowed-tools` make strict YAML parsers fail. Hermes currently falls back to a line parser, but other consumers can reject the skill.

## Change
Quote both scalar values without changing their content.

## Verification
Parsed the complete frontmatter with PyYAML and asserted the preserved `name`, `description`, and `allowed-tools` values.


jackwener:
Reviewed against current main after #2122. I would not merge this branch now. The breaking part of the YAML frontmatter issue, the unquoted `description` containing `: ` sequences, is already fixed on main by e2e9a126 / #2122. I also parsed the current frontmatter with `js-yaml`; `allowed-tools: Bash(opencli:*), Read, Edit, Write, Grep` is accepted as the intended scalar because the colon is not followed by whitespace.\n\nThe remaining diff here only quotes `allowed-tools`, which is harmless but no longer fixes a demonstrated current-main failure. Recommend closing as superseded to avoid carrying a dirty/stale branch.

Issue #1481, opened 2026-05-11
[Feature]: 携程(ctrip) adapter — add flight and hotel-search commands


CubeEdge:
## Feature Request

**Adapter:** `clis/ctrip/` (already exists with `search` and `hotel-suggest`)

**Proposed commands:**

### 1. `ctrip flight` — Flight search and results
Search real-time domestic and international flights on Ctrip.

- **Args:** `from` (departure city/airport), `to` (arrival city/airport), `date` (YYYY-MM-DD), optional `returnDate` for round-trip, optional `adults` (default 1)
- **Output columns:** airline, flightNo, departureTime, arrivalTime, duration, price, stops, cabin class, aircraft type
- **Strategy:** Browser-based (browser: true) — Ctrip flight results are dynamically rendered and present anti-scraping measures
- **Endpoint investigated:** `https://flights.ctrip.com/itinerary/api/` (has structure but likely needs browser session)

### 2. `ctrip hotel-search` — Hotel listing search
Search hotel listings by city/dates, beyond the existing suggest-only `hotel-suggest`.

- **Args:** `city` (city name or ID), `checkin` (YYYY-MM-DD), `checkout` (YYYY-MM-DD), optional keyword filter, optional star rating filter, limit (default 20)
- **Output columns:** hotelId, name, starRating, reviewScore, reviewCount, priceMin, address, lat, lon, url
- **Strategy:** Browser-based (browser: true) — hotel listing pages require session context

### Why now
The existing ctrip adapter proves the maintainer is interested in this site. Adding flight and hotel-search would make it a fully useful travel CLI that AI agents can use for real booking research.

**References:**
- Existing ctrip adapter: `clis/ctrip/` with search.js, hotel-suggest.js, utils.js
- Ctrip flight search: https://flights.ctrip.com/
- Ctrip hotel search: https://hotels.ctrip.com/

PR #2142, opened 2026-07-16
fix(ctrip): support current flight card DOM

dxxbb:
## Summary

- support Ctrip's current `[data-testid^="flight-item-"]` flight cards
- preserve compatibility with the legacy `.flight-list > span > div` DOM
- parse airline, flight number, airports, terminal, price, currency, and cabin from current stable field IDs
- add a JSDOM regression case based on the current rendered card structure

## Root cause

`ctrip/flight` waited for the legacy card selector, so the command timed out even when the current page had rendered flight results. The new DOM also places the cross-day marker between arrival time and airport and renders the price amount before the currency symbol, so changing only the outer selector would silently produce incorrect fields.

## Impact

The command now recognizes both DOM variants and keeps the existing output schema and typed error behavior unchanged.

## Validation

- `npx vitest run --project adapter clis/ctrip/ctrip.test.js` — 70 passed
- `npm run typecheck` — passed
- `npm run check:typed-error-lint` — no new violations
- `npm run check:silent-column-drop` — no new violations
- `node dist/src/main.js validate ctrip/flight` — passed
- live source run: `BJS → SHA`, `2026-07-18`, returned complete flight rows with prices

Refs #1481

### DIFF
--- a/clis/ctrip/ctrip.test.js
+++ b/clis/ctrip/ctrip.test.js
@@ -682,6 +682,39 @@ describe('ctrip buildFlightExtractJs (JSDOM)', () => {
         }]);
     });
 
+    it('extracts current data-testid cards via stable field ids', () => {
+        const html = `
+          <div data-testid="flight-item-1">
+            <span id="airlineName3U8282_1784385000000-0">四川航空</span>
+            <div id="comfort-3U8282_1784385000000-0">
+              <span>3U8282<span>空客320(中)</span></span>
+            </div>
+            <div>22:30</div>
+            <span id="departureFlightTrain3U8282_1784385000000-0">首都国际机场</span>
+            <span id="departureTerminal3U8282_1784385000000-0">T2</span>
+            <div>01:55<span id="crossDays3U8282_1784385000000-0">+1天</span></div>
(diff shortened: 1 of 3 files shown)


jackwener:
Reviewed against current main (5256711a+ / after #2156). I would not merge this PR as-is now: the branch is dirty because Ctrip was substantially expanded in #2156, and the root fix from this PR is already represented in current main. `clis/ctrip/flight.js` now uses `.flight-item`, relaxes required `flightNo` for the current cards, waits for rendered cards with time+price, and the expanded `clis/ctrip/ctrip.test.js` covers the modern flight-card extraction path.\n\nSo the original issue this PR targeted has been superseded by newer Ctrip work. Keeping this open risks reintroducing stale assumptions from the older branch; recommend closing rather than rebasing/merging.

jackwener:
关闭为已被后续实现取代。current main 的 Ctrip flight 已由 merged PR #2393／commit 64e6f0e3 改为捕获 batchSearch 结构化响应，包含 schema 校验、认证/验证码/超时分类、完整分页完成信号和稳定排序，不再依赖此 PR 修补的旧 flight-card DOM selector。继续合并旧 DOM 路径会恢复脆弱的双重实现。

PR #2148, opened 2026-07-18
fix(runtime): honor OPENCLI_CDP_ENDPOINT for web adapters

axxop:
## Problem

`OPENCLI_CDP_ENDPOINT` is documented as the manual CDP endpoint override, but regular website adapters ignore it. `getBrowserFactory()` selects `CDPBridge` only for registered Electron apps and always selects `BrowserBridge` for sites such as `xianyu`.

As a result, a command such as:

```bash
OPENCLI_CDP_ENDPOINT=http://127.0.0.1:9333 opencli xianyu whoami
```

still tries to connect through the Browser Bridge daemon. In a setup with multiple connected extension profiles, it fails with `BROWSER_CONNECT: Multiple Browser Bridge profiles are connected`, even though the explicitly configured headless Chrome CDP endpoint is reachable and authenticated.

## Fix

Give an explicitly configured `OPENCLI_CDP_ENDPOINT` precedence in `getBrowserFactory()` and select `CDPBridge` for any adapter when it is present. Existing behavior is unchanged when the variable is absent: registered Electron apps still use CDP and regular sites still use Browser Bridge.

Add regression coverage for both branches:

- a regular site uses `BrowserBridge` by default
- a regular site uses `CDPBridge` when `OPENCLI_CDP_ENDPOINT` is configured

## Verification

- `npx vitest run --project unit src/runtime.test.ts` — 2 passed
- `npm run typecheck` — clean
- Live verification with a dedicated authenticated Chrome profile running under `--headless=new --remote-debugging-port=9333`: `opencli xianyu whoami -f json` returned `logged_in: true` through the manual CDP endpoint.

### DIFF
--- a/src/runtime.test.ts
+++ b/src/runtime.test.ts
@@ -1,0 +1,19 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { BrowserBridge, CDPBridge } from './browser/index.js';
+import { getBrowserFactory } from './runtime.js';
+
+describe('getBrowserFactory', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  it('uses BrowserBridge for regular sites by default', () => {
+    expect(getBrowserFactory('xianyu')).toBe(BrowserBridge);
+  });
+
+  it('uses CDPBridge when OPENCLI_CDP_ENDPOINT is configured', () => {
+    vi.stubEnv('OPENCLI_CDP_ENDPOINT', 'http://127.0.0.1:9333');
+
+    expect(getBrowserFactory('xianyu')).toBe(CDPBridge);
+  });
+});
--- a/src/runtime.ts
+++ b/src/runtime.ts
@@ -7,10 +7,12 @@ import { DEFAULT_BROWSER_COMMAND_TIMEOUT, DEFAULT_BROWSER_CONNECT_TIMEOUT } from
 export { DEFAULT_BROWSER_COMMAND_TIMEOUT, DEFAULT_BROWSER_CONNECT_TIMEOUT };
 
 /**
- * Returns the appropriate browser factory based on site type.
- * Uses CDPBridge for registered Electron apps, otherwise BrowserBridge.
+ * Returns the appropriate browser factory based on explicit configuration and site type.
+ * A manual CDP endpoint takes precedence, registered Electron apps use CDPBridge,
+ * and all other sites use BrowserBridge.
  */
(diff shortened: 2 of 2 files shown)


jackwener:
Reviewed for the low-risk PR sweep. I am not merging this one in the batch.

The reported behavior is plausible: current `getBrowserFactory()` only selects `CDPBridge` for registered Electron apps, so a regular web adapter does not globally switch to CDP just because `OPENCLI_CDP_ENDPOINT` is set. But making that environment variable override all web adapters changes the core browser-routing contract, not just one adapter. It needs a separate decision on whether a globally exported CDP endpoint should intentionally bypass the Browser Bridge for every browser-backed site, plus broader regression coverage around browser commands, Electron apps, profile routing, and accidental inherited env vars.

Recommended disposition: defer as a core/runtime behavior change until that contract is explicitly accepted and tested, rather than treating it as a low-risk fix.

jackwener:
Merged as `4e8109b6`. This restores behaviour the docs already promised, which is what made it a clear fix rather than new surface.

`docs/advanced/android-chrome.md:157` ships exactly this:
```
OPENCLI_CDP_ENDPOINT=http://localhost:9222 opencli twitter trending
```
`twitter` is a website CLI, not a registered Electron app — so `getBrowserFactory()` routed it to `BrowserBridge` and the documented command could not work. `docs/advanced/remote-chrome.md` makes the same promise for headless/CI setups.

**Verified:**
- Change is one line plus a comment; `src/runtime.test.ts` is new and covers both branches (default → `BrowserBridge`, endpoint set → `CDPBridge`).
- Mutation-checked: removing `if (process.env.OPENCLI_CDP_ENDPOINT) return CDPBridge;` fails exactly the `uses CDPBridge when OPENCLI_CDP_ENDPOINT is configured` case.
- Typecheck clean and 2/2 green **against the merge result** — worth noting because this branch forked at `b0f84c99`, so checking out the branch tree wholesale reverts unrelated files and produces a misleading failure. The gates above were run with only the two changed files applied on top of current `main`.
- Fresh merge-tree clean.

**On CI:** this PR showed `no checks reported`, but that did **not** mean checks were unavailable — no workflow run had ever been recorded for the branch (opened 2026-07-18). I closed/reopened to retrigger, approved the runs, and merged only after all three went green. Flagging that because "no checks reported" is easy to read as "this PR can't get CI".

**Note for #2286:** it adds this same line. Its remaining content (new `dedicatedTarget`/`keepTab` connect options, e2e infra, `vitest.config.ts` changes) is separate from this fix and will need rebasing.

PR #2150, opened 2026-07-19
fix: Add bearer token authentication for daemon HTTP endpoints

nobugpal:
## Summary

Adds bearer-token authentication to the daemon HTTP API (`127.0.0.1:19825`) as a defense-in-depth security measure. The `/command` and `/shutdown` endpoints now require an `Authorization: Bearer <token>` header. The `/ping` endpoint remains open for health-check compatibility.

## Background

The daemon currently relies on an `X-OpenCLI` custom header and an origin check to gate access to the command API. While these prevent browser-originated CSRF attacks, any local process on the same machine can trivially set these headers. This change adds a cryptographic bearer token that is:

- **Randomly generated** (64 hex chars = 256 bits of entropy) on daemon startup
- **Persisted** to `~/.opencli/daemon.token` with owner-only permissions (0600)
- **Reused** across daemon restarts so CLI sessions are uninterrupted

### `src/daemon.ts`
- Import `crypto.randomBytes`, `fs`, `path`, `os` for token management
- Add `loadOrGenerateToken()` — reads existing token or creates a new one
- Add `verifyAuth(req)` — validates `Authorization: Bearer <token>` header
- Apply token check to `POST /command` and `POST /shutdown` handlers
- Update doc comment to reflect the new security layer

### `src/browser/daemon-transport.ts`
- Import `fs`, `path`, `os` for reading the token file
- Add `getDaemonToken()` — loads token from `~/.opencli/daemon.token` on first request
- Automatically includes `Authorization: Bearer <token>` in all daemon HTTP requests

## Backward Compatibility

- **CLI daemon-client**: Automatically reads and sends the token — no user-facing changes
- **Chrome Extension**: The extension connects via WebSocket (ws://127.0.0.1:19825/ext), which has its own `verifyClient` origin check and is unaffected by this change
- **Health checks**: `/ping` remains unauthenticated for daemon-reachability probing
- **Existing token files**: On upgrade, the daemon reuses an existing `daemon.token` if present and valid, so users updating opencli in-place do not need to re-authenticate

## Testing

- Manual verification: `/command` without token returns 401; with valid token passes through to normal processing
- Token file permissions verified: `-rw-------` (0600)
- All existing tests continue to pass with the token auto-injected by the transport layer

Closes #2127


jackwener:
Reviewed for the low-risk PR sweep. I would not merge this branch as-is in this batch.

The goal is reasonable, but this is a daemon protocol/security design change rather than an adapter-local fix. It changes auth semantics for `/command` and `/shutdown`, has no regression coverage for old/new CLI-daemon skew, daemon restart/token lifecycle, status/shutdown behavior, or the extension/daemon compatibility boundary, and it also carries unrelated stale `extension/package-lock.json` version/lockfile churn. The `Closes #2127` reference also appears stale/confusing because #2127 is already a merged PR in this repository, not the daemon-auth issue.

Recommended disposition: defer to a dedicated daemon-auth design PR with protocol compatibility tests and no unrelated lockfile changes, rather than merging during this review pass.

jackwener:
关闭此不完整的重复实现，daemon auth 的真实需求仍由 issue #395 跟踪。该分支只保护 HTTP command/shutdown，未覆盖 WebSocket extension 身份、CLI/daemon 版本错配、token restart lifecycle 和 extension compatibility，并带有无关 stale lockfile churn；现有 #397 至少以同一协议设计同时处理 HTTP 与 WebSocket，应作为后续专门安全 review 的单一分支。

PR #2168, opened 2026-07-23
[codex] fix extension background window focus steal on macOS

### LINKED ISSUE #2167: [Bug]: Background window steals focus on initial creation on macOS
## Description

The first Browser Bridge window created for a background session can activate Chrome on macOS, even though OpenCLI passes `focused: false` to `chrome.windows.create`.

This is the residual initial-creation case from #739. The current grouped-window architecture avoids repeated windows, but a newly created owned container still interrupts the foreground application once.

## Steps to reproduce

1. Use macOS with no existing OpenCLI-owned browser window.
2. Set `OPENCLI_WINDOW=background`.
3. Keep another application focused.
4. Run:

```sh
opencli browser focus-probe open https://example.com/ --window background
```

## Actual behavior

Chrome creates a normal window and becomes the foreground application.

## Expected behavior

The owned window is created without activating Chrome or stealing keyboard focus.

## Environment

- OpenCLI: 1.8.6
- Extension: 1.0.22
- macOS: 26.5.2 (25F84)
- Chrome: 150.0.7871.181
- Node.js: 22.22.0

## Proposed fix

Create background owned windows with `state: minimized` and omit `focused` and window bounds. Chrome does not allow minimized creation to be combined with `left`, `top`, `width`, or `height`. Keep the existing focused, sized create data for foreground mode.

Unlike the post-create minimize attempted in #1232, minimizing in the initial `windows.create` call prevents the visible focus flash.

Daily-AC:
## Description

On macOS, creating a normal Chrome window with `focused: false` can still activate Chrome and interrupt the foreground application.

This change creates background owned-container windows directly in the minimized state. It omits bounds for that path because Chrome rejects minimized creation when `left`, `top`, `width`, or `height` are also supplied. Foreground windows retain the existing focused 1280x900 behavior.

The regression test now checks the complete create payload for both modes, and the generated extension bundle is updated.

Related issue: Closes #2167

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] My code follows the existing project style
- [x] I have performed a self-review
- [x] I have added or updated tests
- [x] New and existing targeted tests pass locally
- [x] The extension production build succeeds
- [x] I have tested the change with a real Chrome extension session

## Screenshots / Output

```text
npx vitest run --project extension extension/src/background.test.ts
Test Files  1 passed (1)
Tests       71 passed (71)

npm run build
dist/background.js  92.53 kB
built successfully
```

Manual macOS verification:

- Started with one user Chrome window and Orca focused.
- Opened `https://example.com/` in a fresh background OpenCLI session.
- Sampled the frontmost application every 50 ms while the command ran.
- Every sample remained `Orca`; the command returned successfully.
- Released the session and removed the probe window afterward.

`npm run typecheck` currently reports seven errors in existing test code. Running the same command against a clean `upstream/main` worktree reports the identical seven errors, so this PR does not add typecheck failures.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -1447,13 +1447,18 @@ async function ensureOwnedContainerWindowUnlocked(role, initialUrl, mode = "back
     };
   }
   const startUrl = initialUrl && isSafeNavigationUrl(initialUrl) ? initialUrl : BLANK_PAGE;
-  const win = await chrome.windows.create({
+  const createData = mode === "background" ? {
     url: startUrl,
-    focused: mode === "foreground",
+    state: "minimized",
+    type: "normal"
+  } : {
+    url: startUrl,
+    focused: true,
     width: 1280,
     height: 900,
     type: "normal"
-  });
+  };
+  const win = await chrome.windows.create(createData);
   container.windowId = win.id;
   await persistRuntimeState();
   console.log(`[opencli] Created owned ${role} window ${container.windowId} (start=${startUrl})`);
--- a/extension/src/background.test.ts
+++ b/extension/src/background.test.ts
@@ -1252,7 +1252,7 @@ describe('background tab isolation', () => {
     const { chrome, tabs, groups } = createChromeMock();
     let nextWindowId = 20;
     let nextTabId = 200;
-    chrome.windows.create = vi.fn(async ({ url, focused, width, height, type }: any) => {
+    chrome.windows.create = vi.fn(async ({ url, focused, width, height, type, state }: any) => {
       const windowId = nextWindowId++;
       const tab: MockTab = {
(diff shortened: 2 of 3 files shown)


Daily-AC:
@jackwener @Astro-Han Gentle review ping when you have a chance.

This fixes the macOS focus steal caused by creating the automation window visible before minimizing it. It has been verified against a real Chrome session, and all functional checks pass. The audit failure is from the existing `js-yaml` advisory and is unrelated to this change.

This PR overlaps #2169 in `background.ts`; I am happy to rebase whichever lands second.

PR #2169, opened 2026-07-23
[codex] add browser tab grouping toggle

### LINKED ISSUE #2069: [Feature]: Option to disable Chrome tab group creation ("OpenCLI Browser" orange group)
### Problem

Every time OpenCLI opens a browser tab via `opencli browser <session> open ...`, the Chrome extension creates (or reuses) a **tab group** named **"OpenCLI Browser"** with orange color. Over time, empty/abandoned groups accumulate in Chrome, cluttering the tab bar:

- Restarting Chrome → old window gone, new window + new group created next time
- Using different session names → potentially multiple groups
- No way to prevent this behavior

This is especially disruptive for users who already have structured tab management (e.g., custom manual groups, Tab Groups for projects) — the auto-created "OpenCLI Browser" group feels invasive.

### Root Cause

In the Chrome extension's `background.js`:

```
const CONTAINER_TAB_GROUP_TITLE = {
  interactive: "OpenCLI Browser",
  automation: "OpenCLI Adapter",
};
const OWNED_TAB_GROUP_COLOR = "orange";
```

The `ensureOwnedContainerGroup()` → `createOwnedGroup()` path always creates/ensures a tab group. There is no configuration to opt out.

### Requested Solution

Add a **configuration option** to disable automatic tab group creation, ideally in one of these forms:

1. **Environment variable** — e.g. `OPENCLI_NO_TAB_GROUP=1` (simplest, easy to set in .zshrc)
2. **CLI flag** — e.g. `opencli --no-group browser <session> open ...` or through the `--profile` mechanism
3. **Extension preference** — a toggle stored via chrome.storage.local that the CLI can set once

### Behavior When Disabled

When the option is active:
- The extension still manages its own owned Chrome window
- Tabs are opened in that window normally, but **not grouped** into a named/colored tab group
- All other functionality (session management, tab lease, idle timeout) remains unchanged

### Workaround

Users can currently use `opencli browser <session> bind` to bind to an existing tab (avoids creating new windows/groups), but this is not suitable for all workflows (e.g. automation, multi-session browsing).

### Environment

- OpenCLI: (check with `opencli --version`)
- Extension: 1.0.21
- Chrome: latest

Daily-AC:
## Description

Chrome automatically preserves named tab groups in its Saved Tab Groups bar. When OpenCLI browser windows are closed or recreated across browser or extension lifecycles, closed `OpenCLI Browser` entries can accumulate even though the live-group convergence logic from #2098 correctly keeps one active group.

The extensions API can query and mutate live `chrome.tabGroups`, but it has no API for deleting Chrome saved-group entries. This change therefore prevents new saved entries at the source by adding a **Group browser tabs** toggle to the extension popup.

The setting defaults to enabled, preserving current behavior. When disabled:

- live `OpenCLI Browser` tabs are ungrouped;
- the in-session group cache and ledger are cleared;
- future browser sessions stay ungrouped;
- owned window and lease tracking continue unchanged.

The preference is loaded before startup reconciliation and stored in `chrome.storage.local`. Active adapter grouping is unaffected because visible adapter groups were removed in #1925. Existing profiles can still contain legacy saved `OpenCLI Adapter` entries created by older releases; like previously saved `OpenCLI Browser` entries, Chrome requires users to remove them once from the saved tab groups bar.

This revisits the approach from closed PR #1463 after the explicit user scenario requested in #2069. Unlike that older PR, the setting is exposed in the extension popup, is scoped to the remaining browser group path, and cleans up live groups when switched off.

Related issue: Closes #2069

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactor
- [ ] Performance improvement
- [x] Test update

## Checklist

- [x] My code follows the existing project style
- [x] I have performed a self-review
- [x] I have added or updated tests
- [x] New and existing targeted tests pass locally
- [x] The extension production build succeeds
- [x] I have tested the change with a real Chrome extension session

## Screenshots / Output

```text
npx vitest run --project extension extension/src/background.test.ts
Test Files  1 passed (1)
Tests       73 passed (73)

npm run build
dist/background.js  94.49 kB
built successfully

node --check extension/popup.js
passed
```

Manual verification with Chrome 150 on macOS 26.5.2:

- Disabled **Group browser tabs** from the real extension popup.
- Opened two named browser sessions into one owned window.
- Both tabs remained ungrouped.
- The Saved Tab Groups count did not increase.
- Reloaded the extension and confirmed the toggle remained disabled.
- The daemon reconnected successfully after reload.

`npm run typecheck` reports seven errors in existing test code. The same seven errors are present on a clean `upstream/main` worktree; this PR adds no typecheck failures.

### DIFF
--- a/extension/README.md
+++ b/extension/README.md
@@ -24,3 +24,14 @@ Suggested Chrome Web Store justification for `downloads`:
 > so agents can wait for downloads triggered during an automation workflow. The
 > command filters by a user-provided filename or URL pattern and timeout. We do
 > not modify, redirect, or persist user download history.
+
+## Browser Tab Groups
+
+OpenCLI groups browser-session tabs under `OpenCLI Browser` by default. To keep
+those tabs ungrouped, open the extension popup and turn off **Group browser
+tabs**. The preference is stored in the Chrome profile and takes effect for
+existing live OpenCLI groups as well as future browser sessions.
+
+Chrome does not expose saved tab groups through the extension API. Turning the
+setting off prevents new saved `OpenCLI Browser` entries, but previously saved
+entries must be removed once from Chrome's saved tab groups bar.
--- a/extension/README.md
+++ b/extension/README.md
@@ -34,4 +34,5 @@ existing live OpenCLI groups as well as future browser sessions.
 
 Chrome does not expose saved tab groups through the extension API. Turning the
 setting off prevents new saved `OpenCLI Browser` entries, but previously saved
-entries must be removed once from Chrome's saved tab groups bar.
+`OpenCLI Browser` entries and legacy `OpenCLI Adapter` entries must be removed
+once from Chrome's saved tab groups bar.
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -961,6 +961,7 @@ const IDLE_TIMEOUT_DEFAULT = 3e4;
 const IDLE_TIMEOUT_INTERACTIVE = 6e5;
 const IDLE_TIMEOUT_NONE = -1;
 const REGISTRY_KEY = "opencli_target_lease_registry_v2";
+const BROWSER_TAB_GROUPS_ENABLED_KEY = "opencli_browser_tab_groups_enabled";
 const LEASE_IDLE_ALARM_PREFIX = "opencli:lease-idle:";
 const CONTAINER_TAB_GROUP_TITLE = {
   interactive: "OpenCLI Browser",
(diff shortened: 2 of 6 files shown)


Daily-AC:
@jackwener @Astro-Han Gentle review ping when you have a chance.

This adds an opt-out for OpenCLI saved browser tab groups and has been verified against a real Chrome session, including persistence across extension reloads. All functional checks pass; the audit failure is from the existing `js-yaml` advisory and is unrelated to this change.

This PR overlaps #2168 in `background.ts`; I am happy to rebase whichever lands second.

Issue #2024, opened 2026-06-24
`twitter post` fails — X home page React does not hydrate in automation browser


murongg:
## Environment

- OpenCLI: v1.8.4
- OS: macOS (Apple Silicon)
- Browser: Chrome 149 (automation browser via OpenCLI daemon)
- X session: valid (logged in, confirmed by `twitter like` and `twitter reply` working)

## Bug

`opencli twitter post "text"` fails with:

```
Error: Selector not found: [data-testid="tweetTextarea_0"]
```

## Root Cause

X.com home page and `/compose/post` do not fully render in the automation browser. The HTML shell loads but React does not hydrate — **zero `data-testid` elements** exist on the page.

### Debug output (added instrumentation to post.js):

```
url: "https://x.com/home"
reactRootChildren: 2152    (minimal — just the shell)
testIds: []                 (empty — React never hydrated)
hasTimeline: false
hasSideNav: false
title: ""                   (empty — React never set it)
bodyChildren: 21
```

The page is just a static HTML skeleton. No timeline, no compose button, no interactive elements.

### `/compose/post` also fails

Direct navigation to `https://x.com/compose/post` shows a `ScriptLoadFailure` page with the message "Something went wrong, but don't fret — let's give it another shot." and a note about privacy extensions.

## What still works

| Command | Status | Why |
|---------|--------|-----|
| `twitter like` | ✅ | Navigates to a specific tweet URL — React hydrates on tweet pages |
| `twitter reply` | ✅ | Same — operates on an existing tweet page |
| `twitter search` | ✅ | Search results page renders fine |
| `twitter timeline` | ✅ | Home timeline data is fetched via API, not DOM |
| `twitter post` | ❌ | Needs the compose UI on home or /compose/post — both fail to hydrate |

## Suspected cause

X has tightened anti-bot detection specifically on the home page and standalone compose route. Tweet pages (`/user/status/id`) appear to have lighter protection, which is why `like`/`reply`/`search` still work.

The Chrome 149 automation browser may be fingerprinted (navigator.webdriver, CDP detection, etc.) and X blocks React initialization on sensitive routes.

## Possible fixes

1. **Navigate to a tweet first, then use the reply composer as a post composer** — since tweet pages hydrate, open a random tweet, click reply, type, then submit. Hacky but might work.
2. **Use X API (v2) for posting** — bypass the DOM entirely for write operations.
3. **Stealth browser patches** — undetected-chromedriver style patches to avoid bot detection (significant effort).
[...]

PR #2179, opened 2026-07-26
fix(twitter): delete fails on reply detail pages where focal tweet has no self-permalink

zhattention:
## Problem

`opencli twitter delete <url>` fails with `Could not find the "More" context menu on the matched tweet` when the URL points to a reply (and in general whenever the focal tweet is the one to delete).

Root cause: on a status detail page, the focal tweet renders its timestamp as **plain text** — there is no `/status/<id>` self-permalink anchor inside its `<article>`. `findTargetArticle` matches articles by that anchor, so it either matches nothing or a different card (e.g. the parent tweet), and the caret lookup scoped to that article fails.

Reproduced on 1.8.6 / macOS while deleting a reply; verified the DOM via the browser bridge: the reply's article contained a visible `[data-testid="caret"]`, but no self-referential status link.

## Fix

Fall back to probing every visible `article [data-testid="caret"]`, sorted so the target article's caret (when matched) is tried first. For each caret: open the menu, look for the localized Delete item (Delete/删除, excluding List/列表), and if absent dismiss with Escape and try the next one.

This is safe by construction: **only the caller's own tweet exposes a Delete menu item**, so opening a wrong tweet's menu is a harmless no-op.

## Tests

- Updated the script-shape assertions to the new caret-probing structure
- Updated the stale-menu-item JSDOM test (it now exercises the probe-and-dismiss path); the guard that a pre-existing stale `删除` item is never clicked still holds
- `npx vitest run clis/twitter/delete.test.js` → 6/6 pass

Related context: found while working around #2024 (write-path hydration blocking) — delete itself is unaffected by that issue once the caret is located.

https://claude.ai/code/session_01VJ8ZjaL1sNofXBaY97T3Cs

### DIFF
--- a/clis/twitter/delete.js
+++ b/clis/twitter/delete.js
@@ -15,35 +15,40 @@ function buildDeleteScript(tweetId) {
               targetArticle = findTargetArticle();
           }
 
-          if (!targetArticle) {
-              return { ok: false, message: 'Could not find the tweet card matching the requested URL.' };
-          }
-
-          const belongsToTargetArticle = (el) => el.closest('article') === targetArticle;
-          const buttons = Array.from(targetArticle.querySelectorAll('button,[role="button"]')).filter(belongsToTargetArticle);
-          // X localizes the "More" caret aria-label (zh-Hans: 更多), so prefer the
-          // language-agnostic data-testid and fall back to a multilingual label match.
-          const moreMenu = Array.from(targetArticle.querySelectorAll('[data-testid="caret"]')).filter(belongsToTargetArticle).find(visible)
-              || buttons.find((el) => visible(el) && /^(More|更多)/.test((el.getAttribute('aria-label') || '').trim()));
-          if (!moreMenu) {
+          // The focal tweet on a status detail page has no self-permalink (its
+          // timestamp is plain text), so findTargetArticle can miss it. Fall
+          // back to probing every visible caret; only the caller's own tweet
+          // exposes a Delete item, so opening the wrong menu is harmless — we
+          // dismiss it and move on.
+          const belongsToTargetArticle = (el) => targetArticle && el.closest('article') === targetArticle;
+          const carets = Array.from(document.querySelectorAll('article [data-testid="caret"]')).filter(visible);
+          carets.sort((a, b) => (belongsToTargetArticle(b) ? 1 : 0) - (belongsToTargetArticle(a) ? 1 : 0));
+          if (!carets.length) {
               return { ok: false, message: 'Could not find the "More" context menu on the matched tweet. Are you sure you are logged in and looking at a valid tweet?' };
           }
 
-          const beforeMenuItems = new Set(document.querySelectorAll('[role="menuitem"]'));
-          moreMenu.click();
-          await new Promise(r => setTimeout(r, 1000));
-
-          const items = Array.from(document.querySelectorAll('[role="menuitem"]'))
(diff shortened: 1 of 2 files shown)


jackwener (changes requested):
Blocking safety issue: when the target article is the exact case this PR is meant to fix (so findTargetArticle returns null), the loop probes carets in DOM order and clicks the first menu item named Delete. The claim that only the caller's own tweet exposes Delete is not enough to identify the requested tweet: a reply detail/self-thread can contain both a parent tweet and the focal reply owned by the same caller, so both menus expose Delete. In that case this code can permanently delete the parent/another own tweet instead of tweetId. The updated tests only assert script shape and stale-menu handling; they do not cover two owned tweet cards. Please keep deletion fail-closed until the focal article can be deterministically tied to tweetId (or use a tweet-id-addressed API/mutation). Probing arbitrary cards and clicking the first destructive action is not safe to merge.

jackwener:
Reviewed as maintainer. **The problem you identified is real, but this fix can delete the wrong tweet — please rework before this can be merged.**

## The failure I reproduced

The new fallback probes *every* visible caret and deletes the first menu that offers Delete:

```js
const carets = Array.from(document.querySelectorAll('article [data-testid="caret"]')).filter(visible);
carets.sort((a, b) => (belongsToTargetArticle(b) ? 1 : 0) - (belongsToTargetArticle(a) ? 1 : 0));
for (const caret of carets) { ...; deleteBtn = findDeleteItem(items); if (deleteBtn) break; }
```

When `targetArticle` is `null` — **precisely the case this PR exists to handle** — `belongsToTargetArticle` is false for every caret, so the sort is a no-op and iteration falls back to plain DOM order.

On a reply detail page, conversation ancestors render **above** the focal tweet. If the user is deleting their own reply in their own thread (a self-thread, which is common), the ancestor is their tweet too, so its menu also offers Delete — and it comes first.

Reproduced in JSDOM with the script this PR generates: focal tweet with no self-permalink, ancestor above it with one, both authored by the user.

```
requested (FOCAL): 2040254679301718161
delete clicked on: ANCESTOR
=> deleted the ancestor, not the requested tweet
```

It halted only because my fixture had no confirmation dialog (`Delete confirmation dialog did not appear.`). On the real page the confirm button exists and the deletion completes.

The premise in the comment — *"only the caller's own tweet exposes a Delete item, so opening the wrong menu is harmless"* — holds for a stranger's tweet, but not for the user's own other tweets on the same page. And this is an irreversible write.

## Why CI did not catch it

The tests assert on the generated **script text** (`expect(script).toContain(...)`), not on behaviour with multiple own-tweets. Every string assertion still passes while the wrong tweet is deleted.

## Suggested direction

Keep the diagnosis — the focal tweet genuinely has no self-permalink — but keep identity verification in the fallback rather than dropping it. Two options that stay small:

1. Derive the focal article from `location.pathname`'s status id (the URL already identifies the tweet) instead of probing menus blindly; or
2. Before clicking Delete, confirm the opened menu's `closest('article')` is either `targetArticle` **or** an article with no self-permalink — i.e. only ever fall back to the *unlinked* focal article, never to a sibling that has its own permalink.

Option 2 is closest to your current shape and would have failed my repro above, since the ancestor *does* carry a self-permalink.

Whatever the shape, please add a test with two of the user's own tweets on one page asserting the requested one is the one deleted — a script-text assertion cannot cover this.

Happy to re-review as soon as it's updated.

PR #2183, opened 2026-07-26
fix(browser): forward preferred profile through health checks

### LINKED ISSUE #2164: bug(browser): forward preferredContextId through health and readiness routing
### Summary

#2073 introduced `preferredContextId` so a persisted default is a soft preference rather than a strict requirement. Command dispatch forwards it, but the health/status/readiness paths still drop it. This makes the configured default ineffective before command dispatch and can make `opencli doctor` or browser startup return `profile_required` whenever two Browser Bridge profiles are connected.

### Reproduction

1. Connect two Browser Bridge profiles.
2. Run `opencli profile use <one-profile>` so `~/.opencli/browser-profiles.json` has that `defaultContextId`.
3. Run `opencli doctor` (or a browser command that reaches `ensureBrowserBridgeReady()`) without `--profile`.

### Actual

The status/health request contains neither `preferredContextId` nor the persisted default. `/status` resolves with no selection and returns `profile_required`; `doctor` reports that no default is selected even though one is configured.

### Expected

The persisted default should remain a soft `preferredContextId` through health, readiness polling, `/status`, and doctor. With the preferred profile connected it should be selected; if stale, retain #2073 behavior: fall back only when one live profile exists, otherwise require an explicit choice.

### Evidence

- [`profileRouteParams`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/browser/profile.ts#L70-L90) correctly distinguishes explicit `contextId` from soft `preferredContextId`.
- [`BrowserBridge.connect()`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/browser/bridge.ts#L28-L46) drops `routing.preferredContextId` when calling `_ensureDaemon`.
- [`fetchDaemonStatus` / `getDaemonHealth`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/browser/daemon-transport.ts#L69-L87) accept only `contextId`.
- [`waitForBridgeReady`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/browser/bridge-readiness.ts#L5-L23) and [`ensureBrowserBridgeReady`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/browser/daemon-lifecycle.ts#L96-L104) forward only `contextId`.
- [`/status`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/daemon.ts#L262-L286) parses only `contextId`, while `/command` already carries both route fields.
- [`runBrowserDoctor()`](https://github.com/jackwener/OpenCLI/blob/5256711a25458e537c5a63d2a6f9c7fd36d0d1eb/src/doctor.ts#L109-L115) calls bare `getDaemonHealth()`.

### Relation to existing work

This is not #1575 (formatter wording) and not the stale-default arbitration addressed by #2073 / #2081. It is the remaining pre-dispatch health/status propagation gap.

I reproduced this with two connected profiles and a persisted default; forwarding the soft preference end-to-end fixes the health probe while preserving the strict-vs-preferred semantics added in #2073.

Missing-Identity:
## Summary

Fixes #2164 by forwarding a persisted `preferredContextId` through the browser daemon health and readiness routes.

- Keep explicit `contextId` routing strict.
- Thread the soft preference through bridge startup, lifecycle readiness, transport URL serialization, `/status`, and doctor.
- Cover preference forwarding in focused unit tests and the real-daemon transport E2E.

## Validation

- Focused unit tests: 60 passed
- TypeScript typecheck
- Real-daemon transport E2E: 7 passed
- `git diff --check`

The full suite was also run; 41 tests failed only because this hosted sandbox has no writable `/root` home for OpenCLI's home-directory fixtures. The production build's `tsx` wrapper could not open its local IPC socket in this sandbox, while its TypeScript compile and direct manifest generation completed successfully.

### DIFF
--- a/src/browser/bridge-readiness.test.ts
+++ b/src/browser/bridge-readiness.test.ts
@@ -54,6 +54,18 @@ describe('waitForBridgeReady', () => {
     expect(fetchHealth).toHaveBeenCalledTimes(3);
   });
 
+  it('forwards a preferred profile on every health poll', async () => {
+    const fetchHealth: HealthFetcher = vi.fn(async () => readyHealth());
+
+    await waitForBridgeReady(fetchHealth, {
+      timeoutMs: 10_000,
+      preferredContextId: 'personal profile',
+      intervalMs: 1,
+    });
+
+    expect(fetchHealth).toHaveBeenCalledWith({ contextId: undefined, preferredContextId: 'personal profile' });
+  });
+
   it('returns the last observed non-ready health when the deadline expires', async () => {
     const fetchHealth: HealthFetcher = vi.fn(async () => notReadyHealth('profile-disconnected'));
 
--- a/src/browser/bridge-readiness.ts
+++ b/src/browser/bridge-readiness.ts
@@ -2,22 +2,22 @@ import type { DaemonHealth } from './daemon-transport.js';
 
 export type { DaemonHealth };
 
-export type HealthFetcher = (opts?: { timeout?: number; contextId?: string }) => Promise<DaemonHealth>;
+export type HealthFetcher = (opts?: { timeout?: number; contextId?: string; preferredContextId?: string }) => Promise<DaemonHealth>;
 
 const DEFAULT_POLL_INTERVAL_MS = 200;
 
 export async function waitForBridgeReady(
   fetchHealth: HealthFetcher,
-  opts: { timeoutMs: number; contextId?: string; intervalMs?: number },
+  opts: { timeoutMs: number; contextId?: string; preferredContextId?: string; intervalMs?: number },
 ): Promise<DaemonHealth> {
   const interval = opts.intervalMs ?? DEFAULT_POLL_INTERVAL_MS;
-  let health = await fetchHealth({ contextId: opts.contextId });
+  let health = await fetchHealth({ contextId: opts.contextId, preferredContextId: opts.preferredContextId });
   if (health.state === 'ready') return health;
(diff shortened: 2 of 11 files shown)


jackwener:
One propagation gap remains in the command transport recovery path, so this does not yet cover the full readiness behavior described in #2164.

`src/browser/daemon-client.ts` computes both `contextId` and `preferredContextId` in `sendCommandRaw()`, but its nested `ensureBridge()` still calls:

```ts
ensureBrowserBridgeReady({
  timeoutSeconds: ...,
  contextId,
  verbose: false,
})
```

That drops the soft preference whenever a fetch failure / daemon replacement enters the ensure-and-retry path. With two live profiles this can still make `/status` return `profile_required` even though a persisted default exists.

The nearby comment says forwarding a stale preference would pin readiness, but `preferredContextId` is deliberately soft: daemon arbitration should select it when live, fall back when exactly one other profile is live, and require a choice only when the fallback is ambiguous. Please forward `preferredContextId` here as well and add a focused recovery-path regression test.

Issue #1825, opened 2026-06-02
[Bug]: Xiaohongshu security block: the note detail page was blocked by risk control.


seveniscat:
### Description

opencli xiaohongshu note 触发风控

```
$ opencli xiaohongshu note https://www.xiaohongshu.com/search_result/688d8fce0000000025024239\?xsec_token\=AB9HP27qLVzkFmokEk4xUr9ynAuPaB2Yy_BtW7sgx2kQ4\=\&xsec_source\=
ok: false
error:
  code: SECURITY_BLOCK
  message: 'Xiaohongshu security block: the note detail page was blocked by risk control.'
  help: The page may be temporarily restricted. Try again later or from a different session.
  exitCode: 1
```

### Steps to Reproduce

opencli xiaohongshu note https://www.xiaohongshu.com/search_result/688d8fce0000000025024239\?xsec_token\=AB9HP27qLVzkFmokEk4xUr9ynAuPaB2Yy_BtW7sgx2kQ4\=\&xsec_source\=

### Expected Behavior

执行结果：

ok: false
error:
  code: SECURITY_BLOCK
  message: 'Xiaohongshu security block: the note detail page was blocked by risk control.'
  help: The page may be temporarily restricted. Try again later or from a different session.
  exitCode: 1

### OpenCLI Version

1.8.1

### Node.js Version

20.x

### Operating System

macOS

### Logs / Screenshots

```shell

```

PR #2207, opened 2026-07-31
fix(xiaohongshu): retry once through a cooldown on risk-control soft blocks (#1825)

LeoLin990405:
## Problem

Reading Xiaohongshu note-detail pages (`note` / `comments` / `download`) back-to-back trips XHS's velocity-based risk control: a soft block that redirects to `website-login/error?error_code=300017`/`300031` or renders `安全限制` / `访问链接异常` (#1825, #962).

Today the **first** block fails the command outright, and an unattended loop just keeps hammering the same endpoint — which is exactly what escalates the risk state toward the account-violation / ban path (#842, #677).

## Change

XHS soft blocks are frequently **transient** per-request challenges, so a single reload after a real cooldown recovers many of them. New shared helper `clis/xiaohongshu/risk-control.js`:

- `readXhsDetailPage(page, { url, extractJs, securityHelp, settleMinS, settleMaxS, ... })` — navigate → settle-jitter → extract; **only** on a `securityBlock` payload, wait one long randomized cooldown (8–18s) and reload+re-extract **exactly once**; if still blocked, throw `SECURITY_BLOCK`.
- The single-retry cap is **structural** — a guarded `if`, not a caller-tunable retry count — so it can never devolve into a hammer loop. `retryOnBlock: false` opts into the previous fail-fast behaviour.
- `jitterSeconds` / `isSecurityBlock` are small pure helpers.

`note` / `comments` / `download` now share this helper, replacing three copies of the inline `securityBlock` detection (download keeps its 1–3s settle; the others 2–5s).

### Scope (honest limitation)

This makes each detail read gentler and recovers transient blocks. It does **not** throttle request velocity **across separate CLI invocations** — capping cumulative velocity needs session-level pacing (the daemon already holds a persistent session, so that's the natural home). I'm happy to follow up with that as a separate change if you're open to it.

## Tests

`readXhsDetailPage` is exercised offline with a mock page:
- transient block → single cooldown reload → recovers
- persistent block → exactly two `goto`/`evaluate` calls, then `SECURITY_BLOCK` (proves the structural cap — never hammers)
- `retryOnBlock: false` → one navigation, fail fast
- custom settle bounds; non-block malformed payloads pass straight through to the caller

```
npx vitest run clis/xiaohongshu/     # 291 passed (existing note/comments/download tests unchanged)
npm run typecheck                    # clean
npm run check:typed-error-lint       # new=0
npm run check:silent-column-drop     # new=0
```

The cooldown/settle bounds and the one-retry policy are intentionally simple constants — easy to tune if you'd prefer different numbers.

### DIFF
--- a/clis/xiaohongshu/comments.js
+++ b/clis/xiaohongshu/comments.js
@@ -6,8 +6,9 @@
  * the --with-replies flag.
  */
 import { cli, Strategy } from '@jackwener/opencli/registry';
-import { AuthRequiredError, CliError, CommandExecutionError, EmptyResultError } from '@jackwener/opencli/errors';
+import { AuthRequiredError, CommandExecutionError, EmptyResultError } from '@jackwener/opencli/errors';
 import { parseNoteId, buildNoteUrl } from './note-helpers.js';
+import { readXhsDetailPage } from './risk-control.js';
 
 const XHS_PROFILE_HREF_SELECTOR = '.author-wrapper a[href*="/user/profile/"], a.name[href*="/user/profile/"], a.user-name[href*="/user/profile/"], a[href*="/user/profile/"]';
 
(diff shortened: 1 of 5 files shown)


jackwener:
Merged as `75c85e57`. Post-merge on main: `clis/xiaohongshu` 337/337, `validate xiaohongshu` 25 commands / 0 errors / 0 warnings.

The property I most wanted to verify is the retry cap, because getting it wrong on a risk-control path is worse than the original bug — and it holds **structurally**, not by convention:

```js
// At most ONE retry — a single `if`, never a loop.
if (retryOnBlock && isSecurityBlock(data)) { ... }
```

An `if` cannot degrade into hammering the way a loop with a counter can. That, plus the comment explaining *why* (escalation toward the account-violation path, #842/#677), is the right shape for this kind of fix.

Also good: the module states what it does **not** do — no throttling across separate CLI invocations, called out as a follow-up rather than silently implied. That keeps the next reader from assuming velocity is globally capped.

Extracting `risk-control.js` adds a file but removes the duplicated block handling from `note` / `comments` / `download`, so the surface is consolidated rather than grown, and `jitterSeconds(min, max, rand = Math.random)` keeps the timing testable without a clock mock.

Verified before merging: merge-tree clean, 296/296 on the branch, typecheck clean, hosted CI green on the exact head.

PR #2224, opened 2026-08-03
fix(boss): retry detached read once

shaomingbo:
## Summary

- retry `bossFetch` once when a `GET` evaluation fails with `detached_mid_command`
- keep the retry bounded to one fresh evaluation
- never retry `POST`, because a detached mid-command write may already have completed

## Root cause

The Browser Bridge correctly treats `detached_mid_command` as non-retryable at the generic `Page.evaluate` layer because the command outcome is unknown. BOSS `bossFetch` knows whether its XHR is a read or a write, so it can safely replay a detached `GET` without weakening the core write-safety boundary.

An isolated BOSS search canary reached `bossFetch` and failed with `Boss API request failed: Detached while handling command.` after navigation had succeeded. The trace reported no failed network requests. The patched adapter was not retried live because the canary authorization explicitly allowed no retry.

## Validation

- regression test failed before the change and passed after it
- verifies a detached `GET` succeeds on one replay
- verifies a second detached `GET` is propagated after exactly two evaluations
- verifies a detached `POST` is propagated after one evaluation
- BOSS adapter tests: 5 files, 38 tests passed
- `npm run typecheck`
- `npm run build`
- `git diff --check`


jackwener:
Reviewed as maintainer. **The change is sound and CI is now fully green — but this PR is still marked as a draft, so I have not merged it.** Flip it to "Ready for review" when you are happy and it can go in.

## Review

The retry is correctly bounded to reads:
```js
method === GET && error.code === detached_mid_command
```
`method` defaults to `GET` and write paths pass `POST` explicitly (`utils.js:436`), so a write can never enter the retry. Typed errors stay intact — `throwBossFetchError` re-throws `AuthRequiredError` / `CommandExecutionError` unchanged rather than flattening them into a generic message.

**Mutation-checked:** disabling the retry branch fails exactly two tests, and the second one is the valuable one — *"propagates a second detached GET without entering a retry loop"*. That pins the retry as bounded rather than just present.

Local: `clis/boss` 48/48, typecheck clean, merge-tree clean.

## I pushed one commit: lockfile refresh (`12407c2b`)

Security Audit was failing, but **not because of your change**. The branch carried an old lockfile:
```
this branch : js-yaml 4.3.0   undici 6.27.0    <- matched GHSA advisories
main        : js-yaml 4.3.1   undici 7.29.0    <- already patched
```
I merged `main` in to refresh it. After that: Security Audit / CI / Doc Check all green, and `clis/boss` still 48/48 with typecheck clean.

Nothing else in your diff was touched.

shaomingbo:
Hi @jackwener, friendly ping 🙂 The PR has been flipped to "Ready for review" since Sep 1 and all CI checks are green. Whenever you have a moment, it would be great to get this merged. Thanks!

PR #2257, opened 2026-08-05
fix(browser): dismiss a page's beforeunload prompt instead of wedging the session

Benjamin-eecs:
## Description

A page that arms `beforeunload` can wedge a browser session: the next navigation opens the native "Leave site?" prompt, nothing in the CLI can answer it (the daemon owns the CDP session, so `browser dialog dismiss` reports no dialog), navigation never returns, and every later `evaluate` reads an empty document. Any adapter that clicks before navigating can supply the user gesture the prompt needs; `twitter post` wedged a session for me twice.

Both attach paths, the extension and the direct-CDP `CDPBridge`, install two halves: a guard registered through `Page.addScriptToEvaluateOnNewDocument`, early enough to stop the site's own handler, and a `Page.javascriptDialogOpening` handler that accepts only `beforeunload` dialogs, covering documents already loaded at attach. `alert`, `confirm` and `prompt` stay with `browser dialog`.

Related issue: none filed; #1972 is @biospark33's twitter-scoped PR for the same wedge, and this installs the guard for every adapter in the layer that runs before page scripts.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Navigation-level A/B on Chrome 151.0.7922.76 over raw CDP: a page arms `beforeunload` at load, receives a real click for the user gesture, then navigates away. Without the guard the prompt opens and the navigation never completes, which is the wedge; with the guard pre-injected the navigation completes with no dialog:

```
UNGUARDED {"dialogOpened":true,"dialogType":"beforeunload","navigationCompleted":false}
GUARDED   {"dialogOpened":false,"dialogType":null,"navigationCompleted":true}
```

`npx vitest run src/browser` 433 / 433 and the extension suite 99 / 99, with tests for the guard, the injection at attach, the `CDPBridge` dialog accept, keeping the two guard copies in step, and leaving non-`beforeunload` dialogs alone. Typecheck, both lint gates and doc coverage pass. `extension/dist/background.js` is rebuilt, as the repo tracks it.

Not verified end to end through a loaded build of this extension, since the daemon tracks one extension connection at a time and the installed build holds the slot.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -3,6 +3,21 @@ const DAEMON_HOST = "localhost";
 const DAEMON_WS_URL = `ws://${DAEMON_HOST}:${DAEMON_PORT}/ext`;
 const DAEMON_PING_URL = `http://${DAEMON_HOST}:${DAEMON_PORT}/ping`;
 
+const BEFORE_UNLOAD_GUARD_JS = `
+  (() => {
+    const holder = EventTarget.prototype;
+    const key = '__lsnUnload';  // looks like an internal listener cache
+    if (Object.getOwnPropertyDescriptor(holder, key)) return;
+    try {
+      Object.defineProperty(holder, key, { value: true, enumerable: false, configurable: true });
+    } catch {}
+    window.addEventListener('beforeunload', (event) => {
+      event.stopImmediatePropagation();
+      event.returnValue = '';
+    }, true);
+    window.onbeforeunload = null;
+  })()
+`;
 const attached = /* @__PURE__ */ new Set();
 const tabFrameContexts = /* @__PURE__ */ new Map();
 const frameTargets = /* @__PURE__ */ new Map();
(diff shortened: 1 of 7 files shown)


jackwener:
Closing as-is after core review. Automatically suppressing and accepting beforeunload across every adapter on the user's real Chrome can discard unsaved user work; that is a product/safety policy change, not a low-risk transport fix. The raw-CDP A/B proves the wedge but does not validate the tracked loaded extension path. A replacement needs an explicit ownership/opt-in boundary (for example only agent-created disposable tabs) plus a loaded-extension E2E proving both recovery and preservation of user-owned unsaved pages.

Issue #1883, opened 2026-06-06
auth refresh: include recovery command for not_logged_in rows


jackwener:
## Context
PR #1881 added `opencli auth refresh`. Review noted that `not_logged_in` rows should be directly actionable for agents instead of requiring them to synthesize the login command.

## Goal
Add a safe, structured recovery hint to auth refresh/status rows where appropriate, for example:

```json
{
  "site": "github",
  "status": "not_logged_in",
  "recovery": "opencli github login"
}
```

## Requirements
- No secrets or account identifiers in recovery fields.
- Keep existing table output readable.
- Apply consistently to `auth refresh` and consider `auth status` if the row shape supports it cleanly.
- Align with existing AuthRequiredError recovery-command style.

Follow-up from PR #1881 / thread #OpenCLI:7bbdfcd2.

PR #2258, opened 2026-08-06
fix(auth): add recovery commands to auth rows

Ubehind:
## Description

Add a safe, structured `recovery` field to authentication status and refresh rows when a site is not logged in. The field provides the deterministic command `opencli <site> login` without exposing domains, account identifiers, or adapter error details.

Related issue: #1883

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Validation

- `npx vitest run --project unit src/commands/auth.test.ts`
- `npx tsc --noEmit`
- `npm run build`
- `npm test`
- `npm run check:silent-column-drop`
- `npm run check:typed-error-lint`
- `git diff --check`

### DIFF
--- a/src/commands/auth.test.ts
+++ b/src/commands/auth.test.ts
@@ -60,7 +60,7 @@ describe('auth status collection', () => {
     const rows = await collectAuthStatus({ sites: 'alpha' });
 
     expect(rows).toEqual([
-      { site: 'alpha', status: 'logged_in', logged_in: true, identity: '', checked: 'quick', error: '' },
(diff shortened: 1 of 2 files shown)


jackwener:
Closing from the fix sweep because this adds a new auth-output and recovery-command UX contract rather than fixing a demonstrated regression. It changes every status/refresh row and table schema to carry a mostly-empty column. Please reopen as a focused feature/RFC if desired, with product acceptance for the structured recovery shape and end-to-end proof that the advertised login command is valid across the intended site set.

PR #2262, opened 2026-08-06
fix(browser): thread the preferred profile through readiness and status checks

### LINKED ISSUE #2259: Default profile ignored by daemon readiness checks when >=2 profiles connected (doctor FAIL / "Multiple Browser Bridge profiles are connected")
## Bug: default profile is ignored by daemon readiness checks when ≥2 profiles are connected

**Version:** @jackwener/opencli 1.8.6 (daemon v1.8.6, extension v1.0.22)
**Platform:** macOS, Node v25.0.0

### Summary

When two or more Chrome/Edge profiles are connected to the daemon and a default
profile has been set via `opencli profile use <name>`, commands that don't
pass `--profile` explicitly still fail with:

```
✖  Multiple Browser Bridge profiles are connected
✖  Hint: Select one with --profile <name>, OPENCLI_PROFILE=<name>, or opencli profile use <name>.
```

`opencli doctor` also reports connectivity FAIL with "no default profile was
selected" even though `~/.opencli/browser-profiles.json` clearly has
`defaultContextId` set, and `opencli profile list` shows the profile marked
`default`.

### Root cause

The default profile is resolved as a **soft preference**
(`resolveProfileSelection()` → `source: 'preferred'` → `preferredContextId`),
as opposed to an explicit `--profile`/`OPENCLI_PROFILE` selection which is a
**hard requirement** (`contextId`). This distinction is intentional (see
comments in `browser/profile.js` and `browser/bridge.js`) so a stale default
doesn't block startup.

However, the pre-flight daemon-readiness path only ever threads `contextId`
through — `preferredContextId` is dropped before it reaches the daemon's
`/status` endpoint:

- `browser/bridge.js` `BrowserBridge.connect()` → `_ensureDaemon(timeout, routing.contextId)` — drops `routing.preferredContextId`
- `browser/daemon-lifecycle.js` `ensureBrowserBridgeReady()` → `getDaemonHealth({ contextId })` and `waitForBridgeReady(getDaemonHealth, { timeoutMs, contextId })` — never receives/forwards a preferred id
- `browser/bridge-readiness.js` `waitForBridgeReady()` → `fetchHealth({ contextId: opts.contextId })` — same
- `browser/daemon-transport.js` `fetchDaemonStatus()` → only serializes `contextId` into the `/status` query string, never `preferredContextId`
- `daemon.js` `/status` handler → `resolveExtensionConnection(requestedContextId)` — called with only one arg, so the daemon-side `preferredContextId` arbitration in `resolveProfileRoute` (which *is* correctly implemented for `/command` POSTs) never gets exercised for `/status`

So `doctor`, and any command relying on `ensureBrowserBridgeReady`/health checks
before dispatch, always see "0 explicit + 2 connected profiles, no preference"
and immediately hit the `profile_required` ambiguity error — even with a
default profile configured. The interesting part: the actual `/command` POST
*does* include `preferredContextId` (`daemon-client.js` `sendCommandRaw`), so
if a command manages to skip the readiness pre-check (or you pass `--profile`
explicitly), it dispatches correctly. This makes the bug intermittent-looking:
`--profile <name>` always works, but relying on the configured default does not.

### Repro

```
$ opencli profile list
Connected Browser Bridge profiles

  t8gj7s42 chrome default — connected v1.0.22
  j4zp6kmp — connected v1.0.22

$ cat ~/.opencli/browser-profiles.json
{
  "version": 1,
  "aliases": { "chrome": "t8gj7s42" },
  "defaultContextId": "t8gj7s42"
}

$ opencli browser main state
✖  Multiple Browser Bridge profiles are connected
✖  Hint: Select one with --profile <name>, OPENCLI_PROFILE=<name>, or opencli profile use <name>.
Run opencli profile list to see connected profiles.

$ opencli doctor
...
[FAIL] Connectivity: failed (Multiple Browser Bridge profiles are connected)
Issues:
  • Multiple Chrome profiles are connected to the daemon, but no default profile was selected.
  ...

# Direct daemon query confirms: preferredContextId is simply never reaching /status
$ curl -s -H "X-OpenCLI: 1" "http://127.0.0.1:19825/status" | jq .profileRequired
true
$ curl -s -H "X-OpenCLI: 1" "http://127.0.0.1:19825/status?preferredContextId=t8gj7s42" | jq .profileRequired
false
```

I also verified this is not Chrome-specific — reproduced identically with a
second profile bound to Microsoft Edge (`j4zp6kmp`): `opencli --profile
j4zp6kmp browser edge open/state/screenshot` works fine and correctly drives
Edge, but omitting `--profile` (relying on the Chrome profile's configured
default) hits the same ambiguity error and `doctor` FAIL, regardless of which
browser is the default.

### Fix

Thread `preferredContextId` end-to-end alongside `contextId` in the readiness
path, mirroring what `daemon-client.js`'s `sendCommandRaw` already does for
`/command`:

1. `browser/bridge.js`: pass `routing.preferredContextId` into `_ensureDaemon`, and forward it into `ensureBrowserBridgeReady({ contextId, preferredContextId })`.
2. `browser/daemon-lifecycle.js`: accept `opts.preferredContextId` in `ensureBrowserBridgeReady`, forward to `getDaemonHealth({ contextId, preferredContextId })` and `waitForBridgeReady(getDaemonHealth, { timeoutMs, contextId, preferredContextId })`.
3. `browser/bridge-readiness.js`: forward `opts.preferredContextId` in both `fetchHealth(...)` calls inside `waitForBridgeReady`.
4. `browser/daemon-transport.js`: serialize `preferredContextId` as an additional `/status` query param in `fetchDaemonStatus`.
5. `daemon.js`: read `preferredContextId` from the `/status` query string and pass it as the second argument to `resolveExtensionConnection(requestedContextId, preferredContextId)` (the daemon-side arbitration in `resolveProfileRoute` already supports this — it's only the caller that was missing it).

I have a local patch implementing exactly this that resolves the issue —
`opencli doctor` now reports `[OK] Connectivity` and `opencli browser <session>
state` succeeds without `--profile`, on both the Chrome and Edge profiles.
Happy to open a PR if useful.

Benjamin-eecs:
## Description

With two or more profiles connected and a default set via `opencli profile use`, every command relying on the readiness pre-check failed with `Multiple Browser Bridge profiles are connected`, and `doctor` reported FAIL claiming no default was selected (#2259). The readiness path dropped `preferredContextId` before it reached the daemon: `/status` only ever received `contextId`, so the daemon-side arbitration `/command` already exercises never ran for health checks.

The preferred id now rides the same params as the explicit one through the readiness path into the `/status` handler, which passes it to `resolveExtensionConnection` exactly as `/command` does. `doctor` threads the resolved selection too, so an explicit `OPENCLI_PROFILE` is likewise diagnosed as the profile it names. Arbitration semantics are unchanged (#2073): a stale default still falls back or fails fast rather than pinning readiness.

The diagnosis and the fix plan are @lyc10031's, laid out in the issue; this implements them with tests.

Related issue: Closes #2259

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Screenshots / Output

Live, two profiles connected (the real Chrome profile `fvqtde6b` plus a second synthetic extension connection), default set to `fvqtde6b`. Before, on v1.8.6:

```
$ opencli browser main state
✖  Multiple Browser Bridge profiles are connected
✖  Hint: Select one with --profile <name>, OPENCLI_PROFILE=<name>, or opencli profile use <name>.
Run opencli profile list to see connected profiles.

$ opencli doctor
opencli v1.8.6 doctor (node v26.0.0)

[OK] Daemon: running on port 19825 (v1.8.6)
[MISSING] Extension: not connected

Profiles:
  • fvqtde6b: connected v1.0.22, default
  • zvyfake1: connected v1.0.22
[FAIL] Connectivity: failed (Multiple Browser Bridge profiles are connected)

Issues:
  • Multiple Chrome profiles are connected to the daemon, but no default profile was selected.
  Run opencli profile list, then opencli profile use <name>, or pass --profile <name>.
  • Browser connectivity test failed: Multiple Browser Bridge profiles are connected
```

After, same setup on this branch:

```
$ opencli browser main state
URL: about:blank

url: about:blank
title: 
viewport: 1280x568
---
---
interactive: 0 | iframes: 0

$ opencli doctor
opencli v1.8.6 doctor (node v26.0.0)

[OK] Daemon: running on port 19825 (v1.8.6)
[OK] Extension: connected (v1.0.22)

Profiles:
  • fvqtde6b: connected v1.0.22, default
  • zvyfake1: connected v1.0.22
[OK] Connectivity: connected in 0.1s

Everything looks good!
```

The daemon-level flip on the same run: `/status` with no params keeps `profileRequired: true`; `?preferredContextId=fvqtde6b` returns `profileRequired: false` with `contextId: fvqtde6b`; a stale preferred with two live profiles stays `true`, keeping the #2073 fallback semantics.

Every touched source file is individually pinned: reverting any one of the seven alone fails at least one named test, with the daemon-side `/status` forwarding pinned end-to-end by a real-daemon e2e case in `tests/e2e/daemon-transport.test.ts`. The six touched suites pass 106 / 106 (`npx vitest run` over src/browser.test.ts, bridge-readiness, daemon-client, doctor, daemon, and tests/e2e/daemon-transport); typecheck, both lint gates and doc coverage pass; `cli-manifest.json` unchanged.

### DIFF
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -170,6 +170,31 @@ describe('BrowserBridge state', () => {
     await expect(bridge.connect({ timeout: 0.1 })).rejects.toThrow('Browser Bridge extension not connected');
   });
 
+  it('threads preferredContextId into every readiness health read', async () => {
+    const { PKG_VERSION } = await import('./version.js');
+    const spy = vi.spyOn(daemonTransport, 'getDaemonHealth').mockResolvedValue({
+      state: 'no-extension',
+      status: {
+        ok: true,
+        pid: 999999,
+        uptime: 0,
+        daemonVersion: PKG_VERSION,
+        extensionConnected: false,
+        pending: 0,
+        memoryMB: 0,
+        port: 0,
+      },
+    });
+
+    const bridge = new BrowserBridge();
+
+    await expect(bridge.connect({ timeout: 0.1, preferredContextId: 'zvypsyje' })).rejects.toThrow('Browser Bridge extension not connected');
+    expect(spy.mock.calls.length).toBeGreaterThan(1);
+    for (const call of spy.mock.calls) {
+      expect(call[0]).toMatchObject({ preferredContextId: 'zvypsyje' });
+    }
+  });
+
   it('attempts stale daemon replacement when daemonVersion is missing', async () => {
     vi.spyOn(daemonTransport, 'getDaemonHealth').mockResolvedValue({
       state: 'no-extension',
--- a/src/browser/bridge-readiness.test.ts
+++ b/src/browser/bridge-readiness.test.ts
@@ -54,6 +54,19 @@ describe('waitForBridgeReady', () => {
     expect(fetchHealth).toHaveBeenCalledTimes(3);
   });
 
+  it('forwards preferredContextId to every health poll', async () => {
+    const sequence: DaemonHealth[] = [notReadyHealth('profile-required'), readyHealth()];
+    let i = 0;
+    const fetchHealth: HealthFetcher = vi.fn(async () => sequence[i++] ?? readyHealth());
+
+    await waitForBridgeReady(fetchHealth, { timeoutMs: 10_000, intervalMs: 1, preferredContextId: 'zvypsyje' });
+
+    expect(vi.mocked(fetchHealth).mock.calls.length).toBeGreaterThan(1);
+    for (const call of vi.mocked(fetchHealth).mock.calls) {
+      expect(call[0]).toMatchObject({ preferredContextId: 'zvypsyje' });
+    }
+  });
+
   it('returns the last observed non-ready health when the deadline expires', async () => {
     const fetchHealth: HealthFetcher = vi.fn(async () => notReadyHealth('profile-disconnected'));
 
--- a/src/browser/bridge-readiness.ts
+++ b/src/browser/bridge-readiness.ts
@@ -2,22 +2,22 @@ import type { DaemonHealth } from './daemon-transport.js';
 
 export type { DaemonHealth };
 
-export type HealthFetcher = (opts?: { timeout?: number; contextId?: string }) => Promise<DaemonHealth>;
+export type HealthFetcher = (opts?: { timeout?: number; contextId?: string; preferredContextId?: string }) => Promise<DaemonHealth>;
 
 const DEFAULT_POLL_INTERVAL_MS = 200;
 
 export async function waitForBridgeReady(
   fetchHealth: HealthFetcher,
-  opts: { timeoutMs: number; contextId?: string; intervalMs?: number },
+  opts: { timeoutMs: number; contextId?: string; preferredContextId?: string; intervalMs?: number },
 ): Promise<DaemonHealth> {
   const interval = opts.intervalMs ?? DEFAULT_POLL_INTERVAL_MS;
-  let health = await fetchHealth({ contextId: opts.contextId });
+  let health = await fetchHealth({ contextId: opts.contextId, preferredContextId: opts.preferredContextId });
   if (health.state === 'ready') return health;
 
   const deadline = Date.now() + opts.timeoutMs;
   while (Date.now() < deadline) {
     await new Promise<void>((resolve) => setTimeout(resolve, interval));
-    health = await fetchHealth({ contextId: opts.contextId });
+    health = await fetchHealth({ contextId: opts.contextId, preferredContextId: opts.preferredContextId });
     if (health.state === 'ready') return health;
   }
   return health;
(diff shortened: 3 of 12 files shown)


jackwener (approved):
Exact-head review of `a62c17a23b239fbeb4fcaf4021ab8ab9f99d4671`: LGTM.

Issue #2259 is still reproducible on current main: the configured default is represented as `preferredContextId`, but readiness/status drop that field, so a two-profile setup reaches `/status` without the daemon's existing soft-preference arbitration. This patch threads the already-defined preference end to end without changing fallback semantics.

Independent verification passed: 99 focused unit tests across BrowserBridge/readiness/daemon-client/doctor/daemon, typecheck, and `git diff --check`. Hosted CI, Doc Check, and Security Audit are terminal green on this exact SHA, and the PR is clean against current main `a86d6470`. No remaining blocker.

PR #2264, opened 2026-08-07
fix(tiktok): protect slow proxy publish attempts

myluke:
## Summary
- apply one shared command deadline across TikTok upload, preparation, write, and receipt phases
- keep slow uploads alive while byte progress continues, and fence each upload capture to its own attempt
- classify ambiguous post-write outcomes as `publish_outcome_unknown`; do not fall back or retry after a write may have started

## Validation
- `CI=true npx vitest run clis/tiktok` — 124 passed (current suite count)
- `CI=true npm run typecheck`

The requested 109-test count is no longer current; the validated TikTok suite now contains 124 tests.


myluke:
Closing because this branch is based on EasyMetaAu/opencli main and would include unrelated fork history against upstream. The clean task-scoped PR is being opened against EasyMetaAu/opencli instead.

PR #2266, opened 2026-08-07
fix(chat): preserve Claude model and wait for Kimi replies

bulexu:
## Description

Fixes two chat-adapter regressions without carrying unrelated adapter churn:

- `claude ask` no longer switches the UI back to Sonnet unless `--model` is explicitly passed. The `model` argument no longer has a default in the manifest, and explicit model changes still fail fast inside an existing conversation.
- `kimi ask` no longer returns while Kimi is still generating a stable-looking intermediate/thinking message. It now waits for the generation control state to clear, strips tool-call/thinking containers from extracted assistant text, and keeps the returned `ReplyPreview` focused on the user-facing answer.
- Kimi SVG clicking now handles React-style controls where inline `onclick` is empty. It preserves the nearest SVG parent as the fallback and only replaces it with an ancestor that is explicitly recognizable as clickable.

Dropped from the original PR head: unrelated Instagram caption/formatting changes and broad Kimi file reformatting.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

## Verification

Local:

```
npx vitest run clis/claude/ask.test.js clis/kimi/kimi.test.js clis/instagram/instagram.test.js --reporter=verbose
# 3 files, 57 tests passed

npm run typecheck
npx tsx src/main.ts validate claude
npx tsx src/main.ts validate kimi
npm run build
git diff --check
```

GitHub PR checks on `2807f261344b441ec6667afa382a1708887fd148` are green after maintainer workflow approval: CI, Security Audit, and Doc Check.

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -7312,9 +7312,8 @@
       {
         "name": "model",
         "type": "str",
-        "default": "sonnet",
         "required": false,
-        "help": "Model to use: sonnet, opus, or haiku",
+        "help": "Switch to sonnet, opus, or haiku before sending. Omit to keep the current selection.",
         "choices": [
           "sonnet",
           "opus",
--- a/clis/claude/ask.js
+++ b/clis/claude/ask.js
@@ -21,7 +21,7 @@ export const askCommand = cli({
         { name: 'prompt', positional: true, required: true, help: 'Prompt to send' },
         { name: 'timeout', type: 'int', default: 120, help: 'Max seconds to wait for response' },
         { name: 'new', type: 'boolean', default: false, help: 'Start a new chat before sending' },
-        { name: 'model', default: 'sonnet', choices: ['sonnet', 'opus', 'haiku'], help: 'Model to use: sonnet, opus, or haiku' },
+        { name: 'model', choices: ['sonnet', 'opus', 'haiku'], help: 'Switch to sonnet, opus, or haiku before sending. Omit to keep the current selection.' },
         { name: 'think', type: 'boolean', default: false, help: 'Enable Adaptive thinking' },
         { name: 'file', help: 'Attach a file (image, PDF, text) with the prompt' },
     ],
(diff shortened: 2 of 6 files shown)


jackwener (changes requested):
Blocking on one behavioral issue in exact head `f094f947`.

`clis/kimi/_utils.js:93-99`: the new helper says it falls back to the nearest SVG parent when no obvious clickable ancestor can be detected, but the loop unconditionally promotes the target once on `i === 0`. For a React-style control shaped like `<div class="wrapper"><div onClick><svg name="Send"/></div></div>` (where React delegation leaves `onclick` empty and neither div has a heuristic class/cursor), this clicks `wrapper`, not the direct click owner. That is precisely the React case this change is intended to repair, and this shared helper also serves Copy/Refresh/Like/Share/SignOut/Close/LeftBar.

I reproduced it with jsdom: a click listener on the direct parent was not invoked; the generic wrapper was the returned/clicked target. The test at `clis/kimi/kimi.test.js:133-140` only asserts that generated script text contains three strings, so it cannot catch the behavior.

Minimal fix: preserve `svg.parentElement || svg` as the fallback; scan candidates upward for up to six levels and replace the target only when `isClickTarget(candidate)` succeeds. Add a behavioral DOM test covering (1) a generic direct React click owner falling back to the direct parent and (2) a recognizable grandparent button/class being selected. The rest of the scoped diff and exact-head verification looks sound; I can re-review immediately after this is corrected.

jackwener (approved):
Exact-head re-review of `2807f261344b441ec6667afa382a1708887fd148`: LGTM.

The prior blocker is resolved correctly. `clickBySvgNameScript` preserves the direct SVG parent as a fixed fallback and only replaces it when the bounded upward scan finds an explicitly recognizable click target. The new jsdom tests behaviorally verify both the generic React direct-parent fallback and recognizable grandparent selection, so the regression is no longer hidden behind source-string assertions.

Independent verification passed: 57 focused Claude/Kimi/Instagram tests, typecheck, build/manifest regeneration with a clean tracked tree, and `git diff --check`. Hosted CI, Doc Check, and Security Audit are terminal green on this exact SHA. The cleaned diff is scoped to Claude model preservation and Kimi send/reply stability; the unrelated Instagram churn remains removed. No remaining review blocker.

PR #2272, opened 2026-08-10
fix(1688): extract detail images from shadow DOM with lazy-render scrolling

HARRY-BEAR:
## Problem

The 1688 `assets` / `download` commands return `detail_count: 0` on real offer pages, even though the page contains 7–16 detail images. Only main images and videos were extracted.

## Root cause

Two compounding issues:

1. **Shadow DOM pierce failure (structural).** The product detail section lives inside the shadow root of a custom element (`v-detail-e`, class `html-description`). The old code queried selectors like `.html-description img` through `queryAllDeep`, which runs the same selector inside each shadowRoot — but `.html-description` is the *host* element in light DOM, so the selector can never match inside the shadow root. The detail group was always empty.

2. **Lazy rendering (timing).** The detail content renders only after the user scrolls near it. The old `autoScroll({ times: 3, delayMs: 400 })` is not enough for long pages (~13k–18k px), so the lazy images never materialized before extraction.

## Fix

- Collect all `img` / `source` elements while walking shadow roots, then decide membership by walking the **shadow host chain** with `closest()` against the known detail containers (`.html-description`, `.de-description-detail`, `#detailContentContainer`, `.desc-lazyload-container`).
- Scroll more aggressively (`autoScroll` 6+4 × 500ms), then `scrollIntoView` the detail container and settle for 3s before extraction.

## Verification

Real-browser results before → after:

| offer | detail before | detail after |
| --- | --- | --- |
| 997412525461 | 0 | 16 |
| 954578900038 | 0 | 13 |
| 871254382620 | 0 | 7 |

`download` end-to-end lands `*_detail_NN.jpg` files; main/video counts unchanged (no regression).

Added a jsdom regression test (`assets.test.js`) covering:
- images inside the `v-detail-e` shadow root → detected as detail
- light-DOM main gallery images → not detected as detail
- light-DOM plain-class detail containers → detected as detail

All 15 tests in `clis/1688` pass; `npx tsc --noEmit` is clean.

### DIFF
--- a/clis/1688/assets.js
+++ b/clis/1688/assets.js
@@ -3,19 +3,21 @@ import { assertAuthenticatedState, buildDetailUrl, buildProvenance, cleanText, e
 // 1688 商品详情区位于自定义元素 v-detail-e 的 shadow DOM 内（懒渲染），
 // 普通 CSS selector 无法穿透 shadowRoot，需沿 shadow host 链判断归属。
 export const DETAIL_CONTAINER_SELECTOR = '.de-description-detail, #detailContentContainer, .html-description, .desc-lazyload-container';
-export function inDetailContainer(el) {
+export function inDetailContainer(el, selector = DETAIL_CONTAINER_SELECTOR) {
     let node = el;
     while (node) {
+        // Check ancestors within the current root first: a detail container can
+        // be a plain element inside a shadow root, not only the host itself.
+        if (node.closest && node.closest(selector))
+            return true;
         const rootNode = node.getRootNode ? node.getRootNode() : null;
         if (rootNode && rootNode.host) {
             const host = rootNode.host;
-            if (host && host.matches && host.matches(DETAIL_CONTAINER_SELECTOR))
+            if (host && host.matches && host.matches(selector))
                 return true;
             node = host;
         }
         else {
-            if (node.closest && node.closest(DETAIL_CONTAINER_SELECTOR))
-                return true;
             node = null;
         }
     }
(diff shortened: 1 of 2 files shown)


jackwener:
Reviewed as maintainer and pushed fixes as `47578743`. Merging.

**The diagnosis is correct and a good catch.** `queryAllDeep` runs the *whole* selector inside each root independently, so when `.html-description` is itself the shadow **host**, `.html-description img` matches nowhere — not on `document` (the img lives in the shadow root) and not on the shadow root (the host isn't inside it). The detail group was provably always empty for shadow-hosted content. Walking the host chain is the right fix.

What I changed:

**1. Restored the two tests this PR replaced.** `normalizeAssets` (grouping, counts, `blob:` filtering) and `normalizeMediaUrl` both lost *all* coverage. The 14 → 15 count hid it, because three new cases were added while two existing ones were deleted, and the description mentioned only "Added a jsdom regression test". They're now back alongside the new ones — 17 total.

**2. Injected the module-level `inDetailContainer` via `toString()` instead of keeping a hand-copied twin inside the evaluated script.** As written, the unit tests exercised the module copy while the browser ran the inline copy, so the tests could pass while the shipped traversal was broken, and the two could drift silently. This is the convention already used in `clis/gov-policy/search.js` and `clis/codex/sidebar.js`.

**3. Closed a traversal gap:** the walk now checks `node.closest(selector)` at each level, not just `host.matches(...)`. A detail container that is a plain element *inside* a shadow root — rather than the host itself — was previously missed.

**4. Replaced `autoScroll(6) + autoScroll(4) + wait(3)` with one `autoScroll`, a `scrollIntoView`, and a bounded poll.** `autoScrollJs` keeps no state between calls (`src/browser/dom-helpers.ts`), so `6 + 4` was byte-for-byte equivalent to a single `10` — the comment about a "second confirmation pass" described something that doesn't happen. Worse, its MutationObserver only short-circuits when `scrollHeight` *grows*, so once the page stopped growing every remaining iteration burned the full delay, and the fixed `page.wait(3)` was then paid on **every** `assets`/`download` call even when the content had already rendered. The poll now returns as soon as the deep detail-image count is stable (capped ~5s). Since `defaultSrcProps` already includes `data-lazyload-src`/`data-src`, URLs are readable without images entering the viewport — only the shadow content needs to render.

**5.** Swapped the fixture's `<source srcset>` for an ``: `defaultSrcProps` doesn't read `srcset`, so asserting on that element implied coverage the adapter doesn't have.

Verified: `clis/1688/` 17/17, typecheck clean, `check:typed-error-lint` and `check:silent-column-drop` both 0 new violations, manifest unchanged (no arg/column changes).

**One thing worth answering in a follow-up rather than blocking here:** `assets.js` already reads `window.context.result.data.gallery.fields`, and `item.js`/`store.js` read `result.data.shippingServices.fields` — so `result.data` is a keyed module map and a description/detail module may well live there too. If it does, the whole DOM traversal could be replaced by a JSON read, which would be far more robust. You have live access; `Object.keys(window.context?.result?.data ?? {})` on a real offer page would settle it. I'm not holding this PR for it because `main` already scrapes the DOM for the `detail` group — this fixes existing scraping rather than introducing it, so the JSON path is a separate improvement.

PR #2279, opened 2026-08-12
fix(browser): add cross-origin frame enumeration fallback

YishanCoding:
## Summary
- preserve DOM snapshot `[F#]` order for cross-origin iframes
- lazily fall back from `Page.getFrameTree` to current-tab-owned `Target.getTargets(type=iframe)` candidates
- fail closed with `frame_enumeration_mismatch` for missing, ambiguous, cross-tab, or stale iframe identity
- route `eval --frame` through the same resolved target

## Safety
- fallback runs only when the normal frame tree misses a DOM-observed frame
- candidates must have a CDP `parentId` chain reaching the current tab target
- duplicate URL/name matches are rejected
- snapshot iframe element identity/order is revalidated before resolution

## Tests
- `npx vitest run --project extension`: 105 passed
- `npx vitest run --project unit src/browser/dom-snapshot.test.ts`: 34 passed
- `npm run typecheck`: 0 errors
- `npm run typecheck --prefix extension`: 0 errors
- `npm test`: 6743 passed, 1 failed, 1 skipped

The one full-suite failure is unchanged from the pre-change baseline and unrelated: `clis/huodongxing/events.test.js > filters Huodongxing loose date results by parsed event date overlap`; expected the `tomorrow` event plus range/same-day, but received only range/same-day.

Validation used mock CDP tests only; no real browser or Judge.me page was opened.

### DIFF
--- a/extension/src/background.test.ts
+++ b/extension/src/background.test.ts
@@ -339,6 +339,303 @@ describe('background tab isolation', () => {
     ]);
   });
 
+  it('falls back to iframe targets when the DOM sees a cross-origin frame missing from the frame tree', async () => {
+    const { chrome } = createChromeMock();
+    chrome.debugger.sendCommand = vi.fn(async (_target: unknown, method: string, params?: { expression?: string }) => {
+      if (method === 'Runtime.enable') return {};
+      if (method === 'Runtime.evaluate') {
+        if (params?.expression === '1') return { result: { value: 1 } };
+        return {
+          result: {
+            value: [{ url: 'https://frame.example/reviews', name: 'app-iframe' }],
+          },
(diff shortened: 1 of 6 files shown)


jackwener:
Closing in the latest-40 fix sweep because this is not safe to merge as submitted. It changes protocol-sensitive cross-origin iframe routing across extension and core using mock-only validation, and it modifies extension/src without the tracked extension/dist/background.js rebuild users actually load. A replacement needs a current loaded-extension E2E on a real cross-origin iframe, the rebuilt dist artifact, and a narrower staged diff that proves target identity and stale/cross-tab rejection end to end.

Issue #2242, opened 2026-08-04
[Bug]: codex commands attach to the avatar overlay target and fail with SELECTOR errors


Benjamin-eecs:
### Description

When the Codex app runs with `--remote-debugging-port`, it exposes two CDP page targets: the avatar overlay (`app://-/index.html?initialRoute=%2Favatar-overlay`) and the main window (`app://-/index.html`). The overlay is listed first in `/json` and keeps that position after `Target.activateTarget` and after activating the app. The connection layer picks it, so every codex command that reads the window DOM runs against the overlay and fails with a SELECTOR error, for example `Could not find element: Codex Composer input element` from `codex send`. Closing the overlay page over CDP (`/json/close/<targetId>`) makes all commands work until the next app start. Observed with Codex 26.727.51351 (Chromium 150.0.7871.182).

### Steps to Reproduce

1. Quit Codex, then start it with the debugging port: `/Applications/Codex.app/Contents/MacOS/ChatGPT --remote-debugging-port=9238`
2. Run `opencli codex send "hello"`
3. See the SELECTOR error; `curl http://127.0.0.1:9238/json` shows the avatar overlay listed before the main window

### Expected Behavior

codex commands attach to the main window rather than the overlay when both are inspectable.

### OpenCLI Version

1.8.6

### Node.js Version

Other

### Operating System

macOS

### Logs / Screenshots

```shell
$ opencli codex send "/review"
ok: false
error:
  code: SELECTOR
  message: 'Could not find element: Codex Composer input element'
  help: The page UI may have changed. Please report this issue.
  exitCode: 1

$ curl -s http://127.0.0.1:9238/json | jq -r '.[] | select(.type=="page") | .url'
app://-/index.html?initialRoute=%2Favatar-overlay
app://-/index.html
```

PR #2281, opened 2026-08-13
fix(codex): handle current desktop CDP layout

yoruuuchan:
## Summary
- deprioritize Codex's auxiliary avatar-overlay CDP target so the main renderer is selected
- make `codex ask` track the latest assistant message identity instead of relying on virtualized turn/message counts
- add regressions for both behaviors

## Reproduction
On Windows with OpenCLI 1.8.6 and current Codex Desktop, CDP 9238 exposes both:
- `app://-/index.html?initialRoute=%2Favatar-overlay`
- `app://-/index.html`

Because both targets otherwise receive the same effective score, `/json` order can select the overlay. `codex status` then reports Connected, while `codex ask` cannot find the composer.

After selecting the main renderer, prompts can be sent. Current Codex also virtualizes the thread DOM, so the number of `[data-content-search-turn-key]` elements may stay unchanged while a new assistant response replaces the visible turn. Assistant messages expose stable per-message identity through `data-content-search-unit-key` / `data-response-annotation-target`, which this change uses for polling, with the old turn-count behavior kept as a fallback for older builds.

## Validation
- real Windows Codex Desktop E2E through CDP: `askCommand.func(..., timeout: 60)` returned `DIRECT60_OK`
- JS syntax checks passed for the modified Codex command and regression test
- regression tests added for CDP target selection and response detection

Local Vitest startup was blocked by a Windows Rolldown/native-binding lock in the temporary dependency install, so repository CI is the source of truth for the full test matrix.

### DIFF
--- a/clis/codex/ask.js
+++ b/clis/codex/ask.js
@@ -22,11 +22,21 @@ export const askCommand = cli({
             throw new ArgumentError('--timeout must be a positive integer (seconds)');
         }
         const selected = await openCodexConversation(page, kwargs);
-        // Snapshot the current content length before sending
-        const beforeLen = await page.evaluate(`
+        // Snapshot the latest assistant identity. Codex virtualizes older turns, so
+        // message/turn counts can stay constant even after a new response appears.
+        const beforeState = await page.evaluate(`
       (function() {
-        const turns = document.querySelectorAll('[data-content-search-turn-key]');
-        return turns.length;
+        const messages = Array.from(document.querySelectorAll('[data-markdown-text-style="assistant-message"]'));
+        const lastMessage = messages[messages.length - 1] || null;
+        const unit = lastMessage?.closest('[data-content-search-unit-key]');
+        const annotation = lastMessage?.closest('[data-response-annotation-target]');
+        return {
+          assistantKey: unit?.getAttribute('data-content-search-unit-key')
+            || annotation?.getAttribute('data-response-annotation-target')
+            || '',
+          assistantCount: messages.length,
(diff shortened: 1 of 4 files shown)


jackwener:
Lane C review note: I would not merge this PR as-is.

The Codex overlay target problem is real, but the target-selection half is `src/browser/cdp.ts` core behavior, not adapter-local. There is also a narrower dedicated PR for the same root cause: #2244, which breaks ties generically toward the unrouted main Electron document instead of hard-coding the Codex `avatar-overlay` route. That is the smaller/Ockham fix for #2242.

This PR also bundles a separate `codex ask` response-virtualization change in `clis/codex/ask.js`. That may be a real adapter fix, but it should be split and reviewed on its own evidence rather than merged together with a superseded core target-selection change.

Local exact-head checks I ran for this PR:
- `npm run build` -> PASS
- `npx vitest run clis/codex/ask.test.js clis/codex/sidebar.test.js src/browser.test.ts` -> 58/58 PASS
- `node dist/src/main.js validate codex` -> PASS, 16 commands

Current-main merge-tree also reports overlapping edits in `src/browser.test.ts`, so it is not a clean lane C merge candidate.

jackwener:
#2244 has now merged as the generic core fix for #2242 (main f909f1e2), so this mixed PR is superseded on its target-selection half. Closing it to prevent the superseded hard-coded route logic from being merged later. If the Codex response-virtualization problem still reproduces on current main, please split only that adapter-local change into a new PR with its own trace and tests.

PR #2282, opened 2026-08-13
fix(extension): omit credentials from daemon ping (fixes #2278)

### LINKED ISSUE #2278: Extension silently never connects when daemon /ping returns 431 (localhost cookie overflow); ping preflight swallows status
## Summary

The extension can silently fail to connect **forever**, with no error anywhere, when the daemon's HTTP `/ping` returns `431 Request Header Fields Too Large`. The popup sits at "Reconnecting…", the service worker console is clean, and the daemon logs show **zero** connection attempts.

Root cause: the ping preflight in `connectAttempt()` treats any non-OK response the same as "daemon down", and both failure branches are silent.

## Environment

- opencli `1.8.6` (npm, global)
- Extension `1.0.22` (unpacked, from the v1.8.6 release asset)
- macOS 15, Node v22.22.3
- Browser: Brave (Chromium-based) — not Chrome-specific, see below

## What happens

`background.js`:

```js
async function connectAttempt() {
  ...
  try {
    const res = await fetch(DAEMON_PING_URL, { signal: AbortSignal.timeout(1e3) });
    if (!res.ok) {
      scheduleReconnect();   // no log, no status
      return;
    }
    ...
  } catch {
    scheduleReconnect();     // empty catch, no log
    return;
  }
  // new WebSocket(...) is never reached
}
```

Browsers attach the **whole `localhost` cookie jar** to that `fetch`. Cookies on `localhost` are shared across ports, so anyone running local dev servers accumulates them — in my case four local Supabase sessions, whose chunked JWT auth cookies (`sb-<ref>-auth-token.0/.1`, ~4.7 KB each) totalled **17,475 bytes** for `localhost`.

Node's default `--max-http-header-size` is **16384**, so the daemon answers `431`, `res.ok` is `false`, and the WebSocket is never attempted. The retry loop repeats this forever.

## Why it is hard to diagnose

- Nothing is logged — the `catch` is empty and the `!res.ok` branch drops the status.
- `curl http://localhost:19825/ping` returns `200`, because curl sends no cookies. Every server-side check looks healthy.
- Running `new WebSocket('ws://localhost:19825/ext')` by hand from the extension's service worker console **succeeds**, which makes it look like the transport is fine and misdirects the investigation. (I don't have an explanation for the fetch/WebSocket asymmetry here — only the observation.)
- The only thing that surfaced it was replaying the extension's own request and printing `res.status` rather than `res.ok`.

## Reproduction

1. Accumulate >16 KB of cookies on `http://localhost` (a few local Supabase logins will do it).
2. Start the daemon normally and load the extension.
3. Popup shows "Reconnecting…" indefinitely; `opencli doctor` reports `[MISSING] Extension: not connected`; daemon logs stay empty.
4. In the extension's service worker console:

```js
fetch('http://localhost:19825/ping', { signal: AbortSignal.timeout(1000) })
  .then(r => console.log(r.status, r.ok));   // 431 false
```

Server-side confirmation:

```bash
BIG=$(python3 -c "print('x'*20000)")
curl -o /dev/null -w "%{http_code}\n" -H "Cookie: junk=$BIG" http://localhost:19825/ping   # 431
curl -o /dev/null -w "%{http_code}\n" http://localhost:19825/ping                          # 200
```

## Workaround

Start the daemon with a larger header limit; the extension then connects on its own immediately:

```bash
NODE_OPTIONS=--max-http-header-size=131072 node .../dist/src/daemon.js
```

## Suggested fixes

1. **Spawn the daemon with a larger `--max-http-header-size`** (e.g. 128 KB) in `daemon-lifecycle.js`. This removes the failure mode for everyone, since the daemon cannot control what cookies the browser attaches.
2. **Don't let the ping gate the WebSocket.** The preflight adds no safety — the WebSocket connection fails on its own if the daemon is down. Attempting it unconditionally would have avoided this entirely.
3. **Log the failure.** At minimum `console.warn('[opencli] ping failed', res.status)` and a non-empty `catch`. A silent empty `catch` on the only path to connectivity makes this class of bug undebuggable from the extension side.
4. Optionally, send the ping with `credentials: 'omit'` so the cookie jar is not attached in the first place.

kagura-agent:
## Problem

Fixes #2278. The extension can silently stop before its WebSocket connection when the daemon `/ping` returns HTTP 431.

A large `localhost` cookie jar can exceed Node's default HTTP header limit. Because the ping used browser-default credentials, Chromium attached those cookies, the daemon answered 431, and the extension silently scheduled another retry without reaching the WebSocket step.

## Fix

- Send only the extension `/ping` request with `credentials: 'omit'`, so browser cookies are not attached.
- Log non-OK ping status codes before scheduling the normal reconnect.
- Keep connection errors silent because a stopped daemon is the expected idle state.
- Rebuild and commit `extension/dist/background.js`, which is the bundle users load.

This deliberately does not raise the daemon header ceiling globally. The reproduced failure is the HTTP ping, omitting credentials resolves that path, and the reporter explicitly confirmed that a direct WebSocket connection succeeds with the same cookie jar. Expanding parsing limits for every daemon HTTP endpoint and WebSocket upgrade therefore lacks supporting evidence.

## Verification

Exact head `4c03e305`:

- extension tests passed (source head: 97/97; rebuilt-dist follow-up: focused 72/72)
- `npm run typecheck` passed
- `cd extension && npm run build` passed
- hosted CI, extension build, security audit, and docs checks all green
- `git diff --check` passed

Final diff: `extension/src/background.ts`, its regression test, and the rebuilt tracked `extension/dist/background.js`.

### DIFF
--- a/extension/dist/background.js
+++ b/extension/dist/background.js
@@ -861,8 +861,12 @@ function connect() {
 async function connectAttempt() {
   if (isDaemonSocketActive()) return;
   try {
-    const res = await fetch(DAEMON_PING_URL, { signal: AbortSignal.timeout(1e3) });
+    const res = await fetch(DAEMON_PING_URL, {
+      signal: AbortSignal.timeout(1e3),
+      credentials: "omit"
+    });
     if (!res.ok) {
+      console.warn(`[opencli] daemon ping failed: HTTP ${res.status}`);
       scheduleReconnect();
       return;
     }
--- a/extension/src/background.test.ts
+++ b/extension/src/background.test.ts
@@ -877,6 +877,30 @@ describe('background tab isolation', () => {
     expect(mod.__test__.getReconnectAttempts()).toBe(0);
   });
 
+  it('pings without credentials and logs a non-OK status instead of swallowing it', async () => {
+    const { chrome } = createChromeMock();
+    vi.stubGlobal('chrome', chrome);
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const fetchMock = vi.fn(async () => ({ ok: false, status: 431 }));
+    vi.stubGlobal('fetch', fetchMock);
+
+    await import('./background');
+
+    await vi.waitFor(() => {
+      expect(fetchMock).toHaveBeenCalled();
+    });
+
+    // The ping must not attach the localhost cookie jar — that is what pushes
+    // the request past Node's header limit and makes the daemon answer 431.
+    expect(fetchMock.mock.calls[0][1]).toMatchObject({ credentials: 'omit' });
+    // A non-OK ping must be logged, not silently swallowed.
+    expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('HTTP 431'));
+    // The WebSocket must not be attempted after a failed ping.
+    expect(MockWebSocket.instances).toHaveLength(0);
+
+    warnSpy.mockRestore();
+  });
+
   it('ignores daemon commands delivered to a superseded WebSocket', async () => {
     const { chrome } = createChromeMock();
     vi.stubGlobal('chrome', chrome);
--- a/extension/src/background.ts
+++ b/extension/src/background.ts
@@ -138,14 +138,24 @@ async function connectAttempt(): Promise<void> {
   if (isDaemonSocketActive()) return;
 
   try {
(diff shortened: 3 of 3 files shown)


kagura-agent:
Hi @jackwener 👋 — gentle ping. This fix for #2278 (silent 431 wedge from localhost cookie overflow) has been open for 3 days with all CI green. Would appreciate a review when you have a moment. Happy to adjust anything.

PR #2286, opened 2026-08-14
fix(browser): honor OPENCLI_CDP_ENDPOINT for website CLIs (remote-Chrome mode)

luckydududu:
## Problem

[`docs/advanced/remote-chrome.md`](https://github.com/jackwener/OpenCLI/blob/main/docs/advanced/remote-chrome.md) promises that setting `OPENCLI_CDP_ENDPOINT` lets OpenCLI drive a remote Chrome (headless servers, CI — the doc even ships a CI recipe with `xvfb` + `--remote-debugging-port`). In practice the variable is only honored for **Electron app** CLIs: website CLIs always route through the Browser Bridge extension (`getBrowserFactory` → `BrowserBridge`), which cannot be installed in many remote/headless setups. So `opencli twitter timeline` etc. fail with `Browser Bridge extension not connected` even when a perfectly good remote Chrome is reachable.

## Fix

- **runtime**: `getBrowserFactory` returns `CDPBridge` for any site when `OPENCLI_CDP_ENDPOINT` is set (explicit opt-in; default behavior unchanged, Electron unchanged).
- **execution**: pass the manual endpoint for non-Electron sites. No localhost port probe here — the endpoint may live on another machine; `CDPBridge` already validates it when fetching `/json`.
- **cdp**: website sessions open a **dedicated tab** via `/json/new` (PUT, with GET fallback for pre-111 Chrome) instead of attaching to whichever existing tab ranks first — a shared remote browser may have the user's real tabs open, and hijacking one to navigate to the site would be destructive. The tab is closed on `close()` (best-effort). `ws://` endpoints and the existing attach path (Electron / no `dedicatedTarget`) keep their previous behavior, including `OPENCLI_CDP_TARGET` ranking.

## Tests

- Dedicated-tab lifecycle against a real local HTTP server: create → close sequence, `PUT`→`GET` fallback on 405, and legacy attach path unchanged (`GET /json` only).
- Factory routing unit tests (env unset → `BrowserBridge`; set → `CDPBridge`).
- `npx tsc --noEmit` clean; full `npm test` — the only failures are 2 pre-existing ones in `src/download/index.test.ts` (cross-domain redirect cookies), reproducible on a clean `main` checkout in the same environment.

## Verified end-to-end

Against a real Chrome on another machine over the network (no extension installed in that Chrome):

```console
$ export OPENCLI_CDP_ENDPOINT="http://<remote-host>"
$ opencli twitter whoami
logged_in: true
username: <redacted>
$ opencli twitter timeline --type for-you --limit 5 --format json
# 5 tweets, correct content
```

Tab hygiene confirmed: repeated runs leave no extra tabs in the shared browser.

## Docs

`remote-chrome.md` updated to state that website CLIs work in this mode, describe the dedicated-tab semantics, and clarify `ws://` endpoint behavior.

### DIFF
--- a/docs/advanced/remote-chrome.md
+++ b/docs/advanced/remote-chrome.md
@@ -49,6 +49,21 @@ OpenCLI attaches to that exact target instead of opening a dedicated tab
 dedicated tab, e.g. for Electron apps).
 :::
 
+#### Dedicated tab lifecycle
+
+| Situation | What happens to the tab |
+| --- | --- |
+| Command finishes (default, `--site-session ephemeral`) | Closed |
+| `--keep-tab true`, or `--site-session persistent` (which implies keep-tab) | Left open |
+| WebSocket handshake or CDP setup fails | Closed, even with `--keep-tab true` — nothing ever attached to it |
+| Command fails mid-run (navigation error, adapter throw) | Closed, unless keep-tab applies |
+
+A kept tab is **not** reused by the next command: each command opens its own tab.
+Nothing is lost by that — cookies and login state live in the shared Chrome profile,
+not in the tab — so a persistent site session still sees the same logged-in state.
+Keep-tab is there for inspecting what a command did, not for session continuity.
+
+
 ### 4. Verify
 
 ```bash
--- a/docs/advanced/remote-chrome.md
+++ b/docs/advanced/remote-chrome.md
@@ -37,6 +37,18 @@ Use `127.0.0.1` instead of `localhost` in the SSH command to avoid IPv6 resoluti
 export OPENCLI_CDP_ENDPOINT="http://127.0.0.1:9222"
 ```
 
+With the endpoint set, website CLIs (e.g. `opencli twitter timeline`) run over CDP
+directly — the Browser Bridge extension is not required in this mode. Each command
+opens a **dedicated tab** on the remote browser and closes it when done, so your
+existing tabs are never hijacked. Electron app CLIs honor the same variable as before.
+
+::: tip
+The endpoint may also be a `ws://` URL pointing at a specific target. In that case
+OpenCLI attaches to that exact target instead of opening a dedicated tab
+(`OPENCLI_CDP_TARGET` filters targets when using an `http://` endpoint without a
+dedicated tab, e.g. for Electron apps).
+:::
+
 ### 4. Verify
 
 ```bash
--- a/src/browser/cdp.test.ts
+++ b/src/browser/cdp.test.ts
@@ -4,12 +4,18 @@ const { MockWebSocket } = vi.hoisted(() => {
   class MockWebSocket {
     static OPEN = 1;
     static lastInstance: MockWebSocket | undefined;
+    /** Set to 'error' to simulate a WebSocket handshake that never connects. */
+    static nextConnect: 'open' | 'error' = 'open';
     readyState = 1;
     private handlers = new Map<string, Array<(...args: unknown[]) => void>>();
 
     constructor(_url: string) {
       MockWebSocket.lastInstance = this;
-      queueMicrotask(() => this.emit('open'));
+      const outcome = MockWebSocket.nextConnect;
+      queueMicrotask(() => {
+        if (outcome === 'error') this.emit('error', new Error('connect ECONNREFUSED'));
+        else this.emit('open');
+      });
     }
 
     on(event: string, handler: (...args: unknown[]) => void): void {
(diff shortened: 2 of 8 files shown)


jackwener:
Reviewed in the latest-40 sweep. The documented remote-Chrome mismatch appears real and the reported live E2E is useful, but this is not a low-risk merge candidate: one PR changes website backend dispatch, DevTools HTTP target creation/closure, PUT→GET compatibility, and session cleanup semantics. I am leaving it open for a dedicated core review rather than merging or rejecting it here. Before merge, please rebase after #2244 and add failure-path evidence that a target created successfully is still closed when WebSocket connect/navigation fails, plus a real-Chrome assertion for the exact /json/new URL form and explicit keep-tab/persistent-session semantics.

luckydududu:
Thanks for the review — all three points are addressed. Rebase is in `3901d335`, the rest in `12a07256`.

**1. Rebase after #2244** — done, rebased onto current `main` (`49907e53`). The rebase was clean, but one context change is worth flagging: #2221 replaced `shouldUseBrowserSession(cmd)` with `siteSession !== null`, so the remote-Chrome branch now lives under the site-session block. That is what made point 3 a real semantics question rather than a docs question, see below.

**2. Failure-path evidence: a created target is still closed when the WebSocket connect or navigation fails**

It was only true by accident before: `browserSession()` closes the factory in a `finally`, so the tab did come down — but `connect()` itself relied on a well-behaved caller. It no longer does. `connect()` now discards its own target on any failure between `/json/new` and a usable session (refused socket, connect timeout, failing `Page.enable`), and does so even when keep-tab was requested, because nothing ever attached to that tab.

Covered at both levels:

- unit (`src/browser/cdp.test.ts`): tab closed after a failed handshake; tab closed after a failed CDP setup; tab still discarded on a failed handshake even with `keepTab: true`
- real Chrome (new e2e, below): the failed handshake is produced by fronting the real DevTools endpoint with a proxy that rewrites the returned `webSocketDebuggerUrl` to a closed port — so Chrome really does create the tab that then has to be cleaned up — plus a failed-navigation case

**3. Real-Chrome assertion for the exact `/json/new` URL form, and keep-tab / persistent-session semantics**

New opt-in e2e at `tests/e2e/remote-cdp-chrome.test.ts`, gated by `OPENCLI_REMOTE_CDP_E2E=1` and skipped when no Chrome binary is found (same discovery helper shape as `browser-ax-chrome.test.ts`). It launches Chrome with `--remote-debugging-port=0`, reads `DevToolsActivePort`, and asserts against a browser that actually enforces the protocol:

```
✓ creates a blank page target through PUT /json/new?url=about:blank
✓ rejects a bare GET /json/new, which is why the PUT-first fallback exists
✓ opens its own tab and removes it on close(), leaving the browser as it was
✓ leaves its tab open when keepTab is requested
✓ removes its tab when the WebSocket handshake fails
✓ removes its tab when navigation fails after connecting

Test Files  1 passed (1)   Tests  6 passed (6)
```

Run against Google Chrome for Testing 151.0.7922.34. The `?url=about:blank` query is asserted rather than assumed: without it Chrome opens the profile's new-tab page, which navigates on its own and races the adapter's first `goto`. The unit test asserts the exact request line `PUT /json/new?url=about:blank` too.

Keep-tab was genuinely wrong, not just undocumented: `close()` removed the tab regardless of `--keep-tab true` or `--site-session persistent` (which implies keep-tab via `resolveKeepTab`). The flag now reaches `CDPBridge` and leaves the tab open. Explicit semantics, now in `docs/advanced/remote-chrome.md`:

| Situation | Tab |
| --- | --- |
| Command finishes (default, ephemeral) | Closed |
| `--keep-tab true`, or `--site-session persistent` | Left open |
| Handshake / CDP setup fails | Closed, even with keep-tab |
| Command fails mid-run | Closed, unless keep-tab applies |

A kept tab is deliberately **not** reused by the next command: each command opens its own. Nothing is lost, because cookies and login state live in the shared Chrome profile rather than in the tab, so a persistent site session still sees the same logged-in state. Keep-tab is for inspecting what a command did, not for session continuity — reusing a tab across processes would need durable target identity that the DevTools HTTP API does not give us.

**Checks**: `tsc --noEmit`, `npm test` (7292 passed), `check:typed-error-lint` and `check:silent-column-drop` both at baseline with 0 new. Note `src/download/index.test.ts` has 2 failures — they reproduce on unmodified `main` at `49907e53`, so they are pre-existing and unrelated to this PR.

PR #2397, opened 2026-08-25
fix(linux-do): use current session for whoami

zhuxiucai:
## Summary

- replace Linux.do's removed `meta[name="current-user-username"]` identity check with Discourse's authenticated `/session/current.json`
- validate `current_user.id` and `current_user.username` before returning a successful `whoami` result
- add adapter regression tests for the endpoint choice, missing session cookies, typed failures, and malformed success payloads

## Problem

On OpenCLI 1.8.7, a valid Linux.do session can have the `_t` cookie and visibly logged-in account controls while the page no longer exposes `meta[name="current-user-username"]`. `opencli linux-do whoami` therefore reports `AUTH_REQUIRED` even though the user is logged in. The subsequent `/u/<self>.json` path is also unnecessary for identifying the current session.

Discourse's `/session/current.json` remains available with the same authenticated browser session and returns the canonical `current_user` identity.

## Verification

- `npx vitest run --project adapter clis/linux-do/auth.test.js` — 6 passed
- `npm run typecheck` — passed
- `node dist/src/main.js validate linux-do` — 10 commands checked, 0 errors, 0 warnings
- `node dist/src/main.js linux-do whoami -f json` — verified against a real logged-in Linux.do session
- `npm test` — 625 test files passed; 7,248 tests passed, 1 skipped

### DIFF
--- a/clis/linux-do/auth.js
+++ b/clis/linux-do/auth.js
@@ -14,30 +14,39 @@ async function verifyLinuxDoIdentity(page) {
   await page.wait(2);
   const probe = await page.evaluate(`(async () => {
     try {
-      const u = document.querySelector('meta[name="current-user-username"]')?.getAttribute('content') || '';
-      if (!u) return { kind: 'auth', detail: 'Linux.do meta[current-user-username] missing — anonymous' };
-      const r = await fetch('/u/' + encodeURIComponent(u) + '.json', {
+      const r = await fetch('/session/current.json', {
         credentials: 'include',
         headers: { Accept: 'application/json' },
       });
       if (r.status === 401 || r.status === 403) {
-        return { kind: 'auth', detail: 'Linux.do /u/<self>.json HTTP ' + r.status };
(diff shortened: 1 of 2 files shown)


jackwener (approved):
Exact-head review GO on `8a5db8955d595e57c43ffd80f51a1924a4c8d5d4`.

The existing failure is caused by treating a removed presentation-layer meta tag as the authentication contract. This change moves identity verification to Discourse's session contract (`/session/current.json`), retains the `_t` preflight, distinguishes auth/HTTP/exception outcomes with typed errors, and validates both required identity fields before returning success. The implementation stays within the existing site-auth abstraction and does not add a second auth path.

Independent local checks passed: all Linux.do adapter tests 22/22, typecheck, `validate linux-do` (10 commands, 0 errors/warnings), both quality gates, and diff-check. Fresh merge-tree against current `origin/main` is clean.

The PR currently has no hosted check runs, so this approval is a code verdict, not authorization to merge without the repository's required hosted checks.

PR #2430, opened 2026-08-29
feat(twitter): allow post to attach a video

luckydududu:
## Description

`twitter post --images` rejects every video before it ever reaches the browser:

```
$ opencli twitter post "..." --images clip.mp4
Unsupported image format ".mp4". Supported: jpg, png, gif, webp
```

That gate is narrower than the implementation behind it. X's composer file input
(`[data-testid="fileInput"]`) accepts video, and the existing DataTransfer upload path already
builds a `File` from raw bytes — so the allowlist plus an image-shaped MIME guess were the only
things standing between the command and a working video post.

This PR lifts the gate and makes the surrounding code honest about carrying video:

- accept `mp4` / `m4v` / `mov`, and resolve the MIME from a lookup table instead of the previous
  image-only ternary (a video used to arrive tagged `image/jpeg`)
- enforce X's actual media rules: up to 4 images **or** exactly 1 video, never mixed. The composer
  silently drops the extras rather than erroring, so validating up front is the only way the user
  finds out
- give video uploads a 180s budget instead of 30s. X transcodes server-side before the Post button
  re-enables; even a small clip can outlast the image budget on a slow link. Image-only posts keep
  the 30s budget
- rename the image-only helpers and messages to `media-*` now that they carry both

Related issue: —

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

Checks run: `npx tsc --noEmit`, `npx vitest run --project adapter clis/twitter/post.test.js`
(33 passed, 5 of them new), `npm run check:typed-error-lint`, `npm run check:silent-column-drop`
(both gates unchanged at baseline), `npm run build` (regenerates `cli-manifest.json`).

New tests cover: video mime + 180s poll budget, images keeping the 30s budget, mixing a video with
images, and more than one video.

## Screenshots / Output

Verified end-to-end against a real Chrome over remote CDP (`OPENCLI_CDP_ENDPOINT`), posting an
842KB / 7s h264+aac clip in one shot:

```
$ opencli twitter post "…" --images clip.mp4 -f json
[
  {
    "status": "success",
    "message": "Tweet posted successfully.",
    "text": "…",
    "id": "…",
    "url": "https://x.com/…/status/…"
  }
]
```

Reading the posted status back confirms the attachment is a real video rather than a dropped
upload:

```
$ opencli twitter thread <url> -f json
    "has_media": true,
    "media_urls": ["https://video.twimg.com/amplify_video/…/vid/avc1/320x568/….mp4?tag=14"],
    "media_posters": ["https://pbs.twimg.com/amplify_video_thumb/…/img/….jpg"]
```

The upload went through the base64 + `DataTransfer` fallback (`page.setFileInput` is not
implemented on that page type), which is the path this PR fixes the MIME for.

### DIFF
--- a/cli-manifest.json
+++ b/cli-manifest.json
@@ -41574,7 +41574,7 @@
         "name": "images",
         "type": "string",
         "required": false,
-        "help": "Image paths, comma-separated, max 4 (jpg/png/gif/webp)"
+        "help": "Media paths, comma-separated: up to 4 images (jpg/png/gif/webp) or 1 video (mp4/mov)"
       }
     ],
     "columns": [
--- a/clis/twitter/post.js
+++ b/clis/twitter/post.js
@@ -7,25 +7,39 @@ import { isRecoverableFileInputError } from './utils.js';
 
 const MAX_IMAGES = 4;
 const UPLOAD_POLL_MS = 500;
-const UPLOAD_TIMEOUT_MS = 30_000;
+const IMAGE_UPLOAD_TIMEOUT_MS = 30_000;
+// X transcodes video server-side before the Post button re-enables, which takes
+// far longer than an image upload even for a small clip.
+const VIDEO_UPLOAD_TIMEOUT_MS = 180_000;
 const COMPOSER_POLL_MS = 250;
 const COMPOSER_TIMEOUT_MS = 10_000;
 const SUBMIT_POLL_MS = 500;
 const SUBMIT_TIMEOUT_MS = 15_000;
 const COMPOSE_URL = 'https://x.com/compose/post';
 const FILE_INPUT_SELECTOR = 'input[type="file"][data-testid="fileInput"]';
-const SUPPORTED_EXTENSIONS = new Set(['.jpg', '.jpeg', '.png', '.gif', '.webp']);
+const IMAGE_EXTENSIONS = new Set(['.jpg', '.jpeg', '.png', '.gif', '.webp']);
+const VIDEO_EXTENSIONS = new Set(['.mp4', '.m4v', '.mov']);
+const MIME_BY_EXTENSION = {
+    '.jpg': 'image/jpeg',
+    '.jpeg': 'image/jpeg',
+    '.png': 'image/png',
+    '.gif': 'image/gif',
+    '.webp': 'image/webp',
+    '.mp4': 'video/mp4',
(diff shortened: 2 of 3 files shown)


jackwener (changes requested):
Requesting changes on exact head `f0f4d39256fa213d02fe0d3e9a094f09660dfc9a`.

The media-shape validation is directionally correct, but the advertised 180-second video wait cannot actually run for 180 seconds through either production browser backend:

- `waitForMediaUpload()` performs the entire poll inside one awaited `page.evaluate()` (`clis/twitter/post.js:189-210`, called at `:356`).
- Direct CDP rejects any `Runtime.evaluate` after 30 seconds (`src/browser/cdp.ts:41,143-155,232-238`). This is the exact backend used by the PR's live evidence.
- BrowserBridge gives an ordinary command 120 seconds (`src/browser/daemon-client.ts:103,120-125`), while `twitter/post` declares no `timeout` argument, so it cannot raise that budget. The extension undercuts that deadline by 5 seconds.

Therefore a video that needs 31s under direct CDP or roughly 116s under BrowserBridge fails before the new 180s timeout branch can return its typed `TimeoutError`. The new test only asserts that the generated string contains `180s`; it mocks `page.evaluate` and never exercises either transport deadline.

Please move upload polling to short Node-side iterations (quick evaluate + wait), or add and use an explicit timeout-aware evaluate contract. Add coverage that composes the command with the real backend deadline behavior, not just a string assertion.

There is also a simpler production upload path already available: `CDPBasePage.uploadFiles()` uses `setFileInput` on BrowserBridge and `DOM.setFileInputFiles` on direct CDP (`src/browser/base-page.ts:774-843`). The current fallback reads an arbitrarily large video fully into Node, base64-expands it, embeds it in JavaScript, decodes another full copy in the page, and was validated only with an 842KB clip. Please use the shared file-upload primitive rather than scaling the image-era DataTransfer fallback to videos.

Independent checks that do pass: focused 33/33, typecheck, `validate twitter` (46 commands, 0 errors/warnings), typed-error and silent-column gates, and diff-check. The blocker is the untested Core/adapter interaction above; hosted checks are also absent on this head.

PR #2504, opened 2026-09-15
fix(doubao): confirm ask submissions and support page mode selection

Tzhen-X:
## Description

`doubao ask` could leave text in the composer without confirming submission, or return homepage/sidebar text through its transcript fallback. This change pins the leased page before native input, requires a new matching user message, and returns only a new completed reply from the message container. It preserves drafts, waits for an unobstructed stable send button, and never automatically resends an uncertain submission. Timeouts now return a typed error instead of a successful `System` row.

It also adds `--mode current|fast|expert` to `ask`. Explicit modes are selected through the visible menu and checked before sending; expert mode is recognized by its badge rather than a hardcoded model version. The default keeps the current selection. `ask` defaults to a foreground window and a 120-second response wait; explicit window and timeout options remain available. Successful output keeps the existing `Role` / `Text` columns.

Related issue: #2400. The reported tab/input problem motivated part of this change; I have not reproduced the original reporter's macOS environment, so this does not claim to resolve every cause described there. Scope is limited to `ask`; the other Doubao commands are unchanged.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 🌐 New site adapter
- [x] 📝 Documentation
- [ ] ♻️ Refactor
- [ ] 🔧 CI / build / tooling

## Checklist

- [ ] Added doc page under `docs/adapters/` (if new adapter) — N/A, existing page updated
- [ ] Updated `docs/adapters/index.md` table (if new adapter) — N/A
- [ ] Updated sidebar in `docs/.vitepress/config.mts` (if new adapter) — N/A
- [ ] Updated `README.md` / `README.zh-CN.md` when command discoverability changed — N/A, existing command; option documented on adapter page

### Test environment

- Windows, Node.js 24.15.0
- Upstream source CLI 1.8.8 at `8271afc67e8504bda94c147f446ee29775d08274`
- Chrome 152, Browser Bridge 1.0.24, existing OpenCLI 1.8.7 daemon
- Logged-in Doubao web account; local user adapter overrides isolated from source CLI tests

### Validation results

```text
npx vitest run --project adapter clis/doubao --maxWorkers 2
Test Files: 5 passed
Tests: 62 passed

npm run build                         PASS
npm run typecheck                     PASS
node dist/src/main.js validate        PASS: 1366 commands, 0 errors, 8 warnings
npm run check:silent-column-drop      PASS: 0 new violations
npm run check:typed-error-lint         PASS: 0 new violations
npm run docs:build                    PASS
git diff --check                      PASS
```

`npm test -- --maxWorkers 4` does **not** fully pass on this Windows environment. An unmodified checkout has 25 failures / 7278 passes / 14 skips; this branch has the same 25 failing test names / 7308 passes / 14 skips. The failures concern symlink permissions and platform-specific path expectations in unrelated tests. No tests were disabled or changed to hide them. Linux/macOS execution remains unverified locally.

Real Chrome checks used the source CLI and a logged-in account: `ask --mode fast` and `ask --mode expert` each submitted a public museum website lookup, returned a completed answer with its URL, and showed the requested mode in the final page snapshot. These checks used Chrome 152 and Browser Bridge 1.0.24 with the existing 1.8.7 daemon. Automated fixtures cover hidden/ambiguous composers, drafts, verification, a busy page, changed modes/text, moving or covered send buttons, unavailable modes, missing confirmation, old replies, streaming replies, and source extraction. Raw browser traces are excluded because they can contain account data.

Implementation and tests were prepared with AI assistance, followed by local diff review and the checks above. This is not a claim of independent human review.


Tzhen-X:
Additional context: This fix was tested against the Doubao web interface as observed on September 15, 2026, including the work-task entry points, chat composer, and mode selection menu.

